PR #2, opened 2021-01-06
Added y measurements for applyop! and_applyop_branches_measurement

michelle4654:


src/experimental/NoisyCircuits.jl
```diff
@@ -136,9 +136,11 @@ function applyop!(s::Stabilizer, m::BellMeasurement)
     res = 0x00
     for (pauli, index) in zip(m.pauli,affectedqubits(m))
         if pauli==X # TODO this is not an elegant way to choose between X and Z coincidence measurements
-            op = single_x(n,index) # TODO this is pretty terribly inefficient... use some sparse check
-        else
-            op = single_z(n,index)
+            op = op.phase*single_x(n,index) # TODO this is pretty terribly inefficient... use some sparse check
+        elseif pauli==Z
+            op = op.phase*single_z(n,index)
+        elseif
```
Krastanov, line 142: This should probably be an `else`.

You can run `] test QuantumClifford` after you make a change to run all of the correctness tests to ensure there are no obvious bugs.

michelle4654: SORRY!!! thanks -- trying to address problems now. I forgot I need to re-open Jupyter notebooks too, if I change anything

Krastanov: Really do not worry about it, it will take a couple of tries to figure out a convenient workflow

Krastanov: Another thing to note: I do not think `op.phase*single_x` would work because I have not implemented a smart enough `*` operator. Maybe you can just overwrite `op.phase .= pauli.phase`?

PR #5, opened 2021-04-08
dense measurement, decision gates and conditional gates for perturbative expansion trajectory

pravissw:


Krastanov:
I think there is a bit of a mixup here. This pull request is against an old temporary branch that was used to try to implement some bit-packing tricks. It should be against the master branch.

pravissw:
Yes, I did want to get this clarified; that's why I left it as a draft. I have now marked it as ready for review.

PR #6, opened 2021-04-15
Create CliffordOps.jl

gsommers:


CliffordOps.jl
```diff
@@ -0,0 +1,182 @@
...
+    end
+    return hadamard, perm
+end
+
+#= replacement of QuantumClifford's random_clifford function
+from Algorithm 2 of Bravyi and Maslov, with code idioms following
+the Python implementation in Qiskit =#
+function rand_clifford(n)
+
+    @assert n < 200 # otherwise matrix operations could fail
```
Krastanov, line 35: @gsommers , any idea why this is happening?

CliffordOps.jl
```diff
@@ -0,0 +1,182 @@
...
+sample (h, S) from the distribution P_n(h, S) =#
+function quantum_mallows(n)
+    arr = [1:n;]
+    hadamard = falses(n)
+    perm = zeros(Int64, n)
+    for idx in [1:n;]
+        m = length(arr)
+        # sample h_i from given prob distribution
+        r = rand()
+        weight = Int64(2 * m - ceil(log2(r*(4^m-1) + 1)))
```
Krastanov, line 22: There is probably a better way to do `ceil(log2(...))` that does not require to pass through floating point numbers (should I also assume this is related to the `@assert n<200`?). Could you express in words what this is doing?

CliffordOps.jl
gsommers: Are you referring to the assert statement? I didn't actually run into this issue myself,  but this was the upper threshold used in the sample Python code by Bravyi and Maslov. I think it has to do with trying to invert large matrices, although 200 x 200 doesn't seem too large to me...

CliffordOps.jl
gsommers: My understanding is that this is a clever way to sample from the probability vector $p(h_i, k) = \frac{2^{m-1+h_i + (m-k)(-1)^{1+h_i}}{4^m-1}$. The naive way that I originally thought of doing it was:
`using StatsBase`
`prob_vec = vec([2^(m-1+h+(m-k)*(-1)^(h+1))/(4^m-1)
                for h in [0,1], k in [1:m;]])`
`weights = Weights(prob_vec)`
`items = vec([(h, k) for h in [Bool(0), Bool(1)], k in [1:m;]])`
`(hadamard[idx], k) = sample(items, weights)`
but I suspect this is less efficient.

Krastanov: Ah, noted, thanks! There is probably some freakishly optimized sampler for this somewhere in system.

Also, something to worry about: `4^200`  will overflow because it is a 64bit integer. `4.0^200` probably will lose precision. It probably needs `BigInt(4)^2` go get the correct result. I will play a bit with it.

Krastanov:
Thanks for looking into this. I will clean it up and merge it probably sometime this weekend. It will probably replace both the random stabilizer and random clifford functions in the library.

CliffordOps.jl
jlapeyre: > There is probably a better way to do ceil(log2(...))

A crude, but faster way is `ispow2(n) ? ILog2.ilog2(n) : ILog2.ilog2(n) + 1`

PR #30, opened 2022-01-15
Expectation value of Pauli strings

ChenZhao44:


Krastanov:
Thanks! Here are some suggestions (I feel relatively strongly about the first one, any others are mostly just a musing):
- [] Use the same order of arguments as QuantumOptics.jl (theirs is not better, it was just first, and I would prefer to not confuse people using both packages) https://docs.qojulia.org/api/#QuantumOpticsBase.expect
- [] Is the linear algebra file the best one to put it in? My knee jerk reaction was to put it in the same one as `project!`
- [] In the tests, it might be easier/cleaner to just do `p_anc = one(PauliOperator, 2n)` and then `p_anc[...] = ...`

Also, could you help with this:
- [] If you are comfortable with doing it, could you add an entry for yourself in the CONTRIBUTORS file
- [] Could you bump the version so I can immediately make a release
- [] Could you add a 0.3.1 entry in NEWS with your addition (a single sentence or less should be enough)

Krastanov:
Oh, one more thing: Could you mention in the docstring of this new function that it allocates a temporary copy of the stabilizer.

src/linalg.jl
```diff
@@ -148,4 +148,19 @@ end
...
+    expect(st::AbstractStabilizer, p::PauliOperator)
+
+Compute the expectation value of a Pauli operator `p` on a stabilizer state `st`.
+"""
+function expect(s::AbstractStabilizer, p::PauliOperator)
+    nqubits(p) == nqubits(s) || error("The number of qubits does not match")
+    _, _, result = project!(copy(s), p)
+    result === nothing && return 0
+    result === 0x00 && return 1
+    result === 0x01 && return im
```
Krastanov, line 163: The 0x01 and 0x03 checks are probably unnecessary as `project!` assumes `s` is a self-consistent stabilizer state, in which case imaginary eigenvalues should be impossible. It probably does not keep the checks there in principle

test/test_expect.jl
```diff
@@ -0,0 +1,40 @@
...
+@testset "Expectation of Pauli strings on stabilizer states" begin
+    st = bell()
+    apply!(st, sX(2))
+    @test expect(st, P"XX") == alter_expect(st, P"XX") == 1
+    @test expect(st, P"ZZ") == alter_expect(st, P"ZZ") == -1
+    @test expect(st, P"XI") == alter_expect(st, P"XI") == 0
+    @test expect(st, P"ZI") == alter_expect(st, P"ZI") == 0
+    @test expect(st, P"II") == alter_expect(st, P"II") == 1
+    @test expect(st, P"-XX") == alter_expect(st, P"-XX") == -1        
+    @test expect(st, P"-iXX") == alter_expect(st, P"-iXX") == -im
```
Krastanov, line 39: Oh, I see why you have the 0x1 and 0x3 checks. Disregard my previous comment.

ChenZhao44:
Thanks for your suggestions. I also strongly agree with you on the first comment because Yao uses similar API.
Could you help me to bump a new version after other things are fixed and this PR is merged? I am not sure that I have the access.

src/linalg.jl
ChenZhao44: We can assert the phase of Pauli operation should not be 0x1 or 0x3. But I did not see that this will be checked in `project!`. So I just defined `expect(p, st)` as `tr(p*st)` here which is more general.

Krastanov:
The failure is from a known bug that I have not gotten around to yet https://github.com/Krastanov/QuantumClifford.jl/issues/28

PR #47, opened 2022-04-25
Use ILog2 with support for RoundUp

jlapeyre:
Use `ilog2(x, RoundUp)` rather than checking and rounding by hand.
This depends on ILog2 v0.2.3, which was released in Nov. 2021

This moves complexity to one point, and it's in the library. It is also
in some cases more performant. I did not find any cases where the PR
decreases performance.

For `quantum_mallows_bigint`, allocate storage for the `BigInt` in
`__init__`. This is more performant. It does add complexity.

Do the same for quantum_mallows_float. Doing ilog2 is slightly faster than
what was there previously.

For example `@btime` shows the following for
```julia
import Random; const rng = Random.default_rng();
@btime QuantumClifford.quantum_mallows_float(rng, 5);
```
before PR: 236ns
after  PR: 199ns

## Remarks

Julia, including the package ecosystem, has almost no support for automatically optimizing arithmetic with `BigInt`s.
(There is https://github.com/tkluck/InPlace.jl, which could potentially be used here. But it doesn't do everything.) 
There are two problems.
* Storage for `BigInt` is dynamically allocated and it is expensive. There are some inplace functions that you can use to reuse intermediate values to avoid allocation. But, it takes some attention to use these.
* You can't save allocated `BigInt`s in the compiled caches (.ji files). You compile in `const` `Ref`s and then do the allocation in the `__init__` function. This can improve efficiency. But, it does add a bit of complexity.

I left the remnants of a few ways to handle the `BigInt` code. In the current PR, the following is done. You can check that `^(::BigInt, ::UInt64)` does an MPZ call without doing any allocation, except for the return value..
```julia
k = rand(rng, 2:big_four_[]^m)
l = ilog2(k, RoundUp)
```

This is another possibility. IIRC, I did not see any consistent performance difference.
```julia
ulim = big_four_[]^m  # BigInt(4)^m
Base.GMP.MPZ.add!(ulim, big_minus_one_[])
rand(rng, 1:ulim)
```
Another possibility is to forgo allocating storage for BigInt in `__init__` and just lift the allocation of `BigInt(4)` out of the loop. This would perhaps not be as efficient, but it would be simpler.

Another thing. The global identifier `big_four_`  should probably be something like `BIG_FOUR` instead. Several Julia style guides recommend this.


Krastanov:
Thank you, this is incredibly helpful insight and suggestions! I will take some time to read through the state of BigInt in Julia, but I should be able to merge this and perform the suggested changes by the end of the week.

Sorry for the state of testing right now. I am in the middle of deprecating a bunch of not-well-thought-out interfaces, so stuff related to Noisy Circuits and Monte Carlo simulations raise a lot of errors. Feel free to neglect them -- they will be cleaned up in the upcoming 0.5

Krastanov:
John, may I include your name and contact in the CONTRIBUTORS file?

Krastanov:
I got some problems with large random tableaux:

```
julia> @benchmark random_destabilizer(200;phases=false)
ERROR: InexactError: Int64(2.0348336732447735e120)
```

(Sorry, this would have been caught by the tests if it was not for the current removal of a few features that are breaking some tests).

And yes, feel free to remove the cruft commented out alternative ways to process the bigints.

jlapeyre:
> John, may I include your name and contact in the CONTRIBUTORS file?
Sure, go ahead!

jlapeyre:
> I got some problems with large random tableaux

I'll take a look...

Krastanov:
It seems the problem appears only around n>30, which is actually the cut off at which I switch to floats. The reason I switch to floats is because at around n>30, ints stop being able to express the numbers I need to work with. Of course, I could just switch to BigInts at that point, but floats are faster than BigInts and are correct up until n~200. So I do not switch to BigInts until n~200.

Thus, to make this work, I am reverting the changes you made to the floats branch (30<n<200) and keeping the changes you made to the other branches. Let me know if this sounds right?

jlapeyre:
Sure, that sounds right. I'll look at it right now.

jlapeyre:
The problem is that `ilog2(x::Float64)` is currently broken if `x > typemax(Int)`.

I am thinking about how to fix it. If I just do `floor(Int, log2(x))`, this works, but is 2.5 times slower than the current `ilog2`, if `x` is in the acceptable range.

Krastanov:
This is great! It will permit me to simplify a bunch of the branching I am currently doing.

PR #66, opened 2022-07-11
 Error correction code library [JSoC]  

mgg39:
This request pulls current progress on the creation of an error correction library for the Quantum Clifford project.


Krastanov:
There seems to be syntactic error somewhere and the automatic tests can not run and the package generally can not be imported. E.g. check the log from the test runner https://github.com/Krastanov/QuantumClifford.jl/runs/7285815294?check_suite_focus=true

src/ecc/shorcode.jl
```diff
@@ -0,0 +1,15 @@
+struct Shor9 <: AbstractECC end
+
+
+parity_checks(c::Shor9) = Sh9"ZZ_______
```
Krastanov, line 4: It seems the syntactic error is here. No macro `Sh9` exists. Did you mean to use the `S` macro that creates a tableau out of the given string?

mgg39: yes, that's exactly what was happening! Thanks for the guidance!

mgg39:
> I am pretty sure these are syntax errors and this would not even import on startup. The colon character is for defining ranges (e.g. 1:5 is 1,2,3,4,5). You probably want commas and semicolons.

yes I am currently working through these

src/ecc/csscodes.jl
```diff
@@ -0,0 +1,299 @@
...
+r = size(A,1)
+block_len = 2^r - 1
+message_len = 2^r - r - 1
+data = A
+println("data",data)
+
+using LinearAlgebra: I
+parity = I(r) #WARNING: both LinearAlgebra and QuantumClifford export "I"; uses of it in module ECC must be qualified
+println("parity",parity)
+
```
Krastanov, line 178: You have a lot of code inside of a library file that seems to be something that should be in a script or in a test file. I would suggest moving all of this away from the library file and putting it in some "scratch" file with which you can experiment and modify, without modifying the library. That would drastically simplify your debugging experience. Right now you need to reimport the entire library to observe a change in this code.

Try to keep library code to only definitions. Experimentation, testing, etc, should be in script files, temporary "scratch" files, files that you can change without changing the entire library.

Krastanov: After you are happy with how something is working in these "scratch" pads, you can move it to a well packaged function inside of the library.

Krastanov: Think of it as the difference between (A) working on a theorem, where you will be scratching out a lot of dead end ideas, experimenting with tangentially-related calculations, crumpling and throwing out whole pages of paper, etc and (B) presenting the finished theorem, where you want a well packaged presentation, not a ton of unrelated dead-end calculations.

mgg39: makes sense! will do that

Krastanov:
This is low priority, but at some point look at the files you have included in this PR: https://github.com/Krastanov/QuantumClifford.jl/pull/66/files

There are a bunch of unrelated things that have been included. E.g. the .vscode file is something related to your editor and we should exclude it (e.g. with `git rm`). And you have modified the notebook files and we would like to keep the originals (you can fix that with `git checkout master <file_name_to_reset>`).

mgg39:
removed the vscode folder

Krastanov:
Closed in favor of #78

PR #78, opened 2022-09-14
ecc_GSoC

mgg39:


mgg39:
Hi! I am trying to push the new naive syndrome circuit test into the original commit but as you can see I have had an issue with the branch which has lead to a rebase and multiple undesired commits being pushed. I cannot seem to be able to get rid of this issue as the following error keeps appearing:
`ERROR: LoadError: SystemError: opening file "/home/maria/.julia/packages/QuantumClifford/fqijb/src/error_correction/Errorcodes.jl": No such file or directory`
and doesn't allow me to either delete the new commits or fully test the current naive syndrome test externally.
That file hasn't existed since July. I am wondering wherever the most straight forward solution might just be to make the changes and push from (yet) another separate branch.

Krastanov:
The merge commits do not seem to have changed anything, so they are probably not the issue. You can read the diffs by clicking on the small hash string on the right of each commit label in the list above.

Have you checked that your installation is in a well defined separate Project environment and that you have the necessary libraries installed. E.g. what does the `] status` command say?

From the message above it looks like there are at least two problems:
- it is looking for a file called `Errorcodes.jl` that does not seem to exist in this pull request
- it is looking for that file in your public preinstalled packages (i.e. in the public released uneditable version of QuantumClifford), not in a local folder dedicated to an editable development version of QuantumClifford in a git repository. Maybe you have written `add QuantumClifford` instead of `dev your_path_to_the_local_quantumclifford_folder`?

mgg39:
Status does confirm that I am running my local version of QuantumClifford.jl, so it is definetly looking for a file called Errorcodes.jl. The file does not exist on my local version. I tried to remove it via `rm`, but that did not work as the file does not exist. Which is why I am unsure on how to proceed to push or check code, as that error is pretty much not letting me to do so in this branch

> The merge commits do not seem to have changed anything, so they are probably not the issue. You can read the diffs by clicking on the small hash string on the right of each commit label in the list above.
> 
> Have you checked that your installation is in a well defined separate Project environment and that you have the necessary libraries installed. E.g. what does the `] status` command say?
> 
> From the message above it looks like there are at least two problems:
> 
> * it is looking for a file called `Errorcodes.jl` that does not seem to exist in this pull request
> * it is looking for that file in your public preinstalled packages (i.e. in the public released uneditable version of QuantumClifford), not in a local folder dedicated to an editable development version of QuantumClifford in a git repository. Maybe you have written `add QuantumClifford` instead of `dev your_path_to_the_local_quantumclifford_folder`?

Krastanov:
Something does not add up though. This branch does not contain any files named `Errorcodes.jl` and more importantly it does not contain any commands that try to include `Errorcodes.jl` so the code in this branch is not capable of raising the error you see. The code you are running locally must not be the one from this branch.

Could you share the entire stack trace with the error? Could you also share the results of `] status`?

mgg39:
The error message:

`Executing task: julia --color=yes --project=/home/maria/JSoC /home/maria/JSoC/Scratch files/naive.jl 

ERROR: LoadError: SystemError: opening file "/home/maria/.julia/packages/QuantumClifford/fqijb/src/error_correction/Errorcodes.jl": No such file or directory
Stacktrace:
  [1] systemerror(p::String, errno::Int32; extrainfo::Nothing)
    @ Base ./error.jl:176
  [2] #systemerror#80
    @ ./error.jl:175 [inlined]
  [3] systemerror
    @ ./error.jl:175 [inlined]
  [4] open(fname::String; lock::Bool, read::Nothing, write::Nothing, create::Nothing, truncate::Nothing, append::Nothing)
    @ Base ./iostream.jl:293
  [5] open
    @ ./iostream.jl:275 [inlined]
  [6] open(f::Base.var"#391#392"{String}, args::String; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Base ./io.jl:380
  [7] open
    @ ./io.jl:379 [inlined]
  [8] read
    @ ./io.jl:460 [inlined]
  [9] _include(mapexpr::Function, mod::Module, _path::String)
    @ Base ./loading.jl:1405
 [10] include(mod::Module, _path::String)
    @ Base ./Base.jl:428
 [11] include(x::String)
    @ QuantumClifford ~/.julia/packages/QuantumClifford/fqijb/src/QuantumClifford.jl:4
 [12] top-level scope
    @ ~/.julia/packages/QuantumClifford/fqijb/src/QuantumClifford.jl:1279
 [13] include
    @ ./Base.jl:428 [inlined]
 [14] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt64}}, source::String)
    @ Base ./loading.jl:1475
 [15] top-level scope
    @ stdin:1
in expression starting at /home/maria/.julia/packages/QuantumClifford/fqijb/src/QuantumClifford.jl:1
in expression starting at stdin:1
ERROR: LoadError: Failed to precompile QuantumClifford [0525e862-1e90-11e9-3e4d-1b39d7109de1] to /home/maria/.julia/compiled/v1.9/QuantumClifford/jl_0OeF7j.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] compilecache(pkg::Base.PkgId, path::String, internal_stderr::IO, internal_stdout::IO, ignore_loaded_modules::Bool)
   @ Base ./loading.jl:1626
 [3] compilecache
   @ ./loading.jl:1570 [inlined]
 [4] _require(pkg::Base.PkgId)
   @ Base ./loading.jl:1271
 [5] _require_prelocked(uuidkey::Base.PkgId)
   @ Base ./loading.jl:1144
 [6] macro expansion
   @ ./loading.jl:1124 [inlined]
 [7] macro expansion
   @ ./lock.jl:267 [inlined]
 [8] require(into::Module, mod::Symbol)
   @ Base ./loading.jl:1088
in expression starting at /home/maria/JSoC/Scratch files/naive.jl:1

 *  The terminal process "julia '--color=yes', '--project=/home/maria/JSoC', '/home/maria/JSoC/Scratch files/naive.jl'" terminated with exit code: 1. 
 *  Terminal will be reused by tasks, press any key to close it. `

mgg39:
status message:

`(QuantumClifford) pkg> status
Project QuantumClifford v0.5.1
Status `~/JSoC/QuantumClifford.jl/Project.toml`
  [861a8166] Combinatorics v1.0.2
  [ffbed154] DocStringExtensions v0.9.2
  [86223c79] Graphs v1.7.4
  [3e5b6fbb] HostCPUFeatures v0.1.13
  [2cd5bd5f] ILog2 v0.2.4
⌅ [2edaba10] Nemo v0.31.1
  [f517fe37] Polyester v0.6.17
  [ae029012] Requires v1.3.0
  [fdea26ae] SIMD v3.4.2
  [2913bbd2] StatsBase v0.33.21
  [37e2e46d] LinearAlgebra
  [9a3f8284] Random
Info Packages marked with ⌅ have new versions available but cannot be upgraded. To see why use `status --outdated``

Krastanov:
you need triple back-ticks if you want to paste multi-line code in Markdown. I.e. you need ```

mgg39:
I was able to force push the naive syndrome circuit test update (changing the b variable against which the gate measurements are tested) , but I still need to resolve the issue in order to test it outside the library to make sure I am running the test I am expecting to run

mgg39:
> you need triple back-ticks if you want to paste multi-line code in Markdown. I.e. you need ```

thank you!

Krastanov:
It seems you are showing/using two different project files.

The first one is `~/JSoC` (see the `--project` flag), the second one seems to be `~/JSoC/QuantumClifford.jl` (it says so in the status report). The first one is probably not actually referring to the code you are looking at.

Could you show the result of `] status` from the first one?

Make sure that all project environments you are working with (at least two given the messages above) all refer to the same git repository. I assume the repository is in `~/JSoC/QuantumClifford.jl`, so make sure that you do `] dev ~/JSoC/QuantumClifford.jl` in each project environment, if you want to use that repository. I am currently assuming that somewhere on your system you have an outdated copy of your repository and that at least one of your environment is pointing to that outdated copy.

Krastanov:
Also, generally, I would suggest not working from inside of the project that defines the library itself. In other words, never use the `~/JSoC/QuantumClifford.jl` project as a project from which to run code. Just use it as the defining location for the library, and refer to it from your other projects. If `] status` is showing that as your current active project, you are probably going to cause yourself trouble -- it is totally feasible do use it that way, but it is unnecessarily complicated. Thus, go back to the outer "scratch" project, the one in which you are getting the error messages, and check the result of `] status` there. I suspect you will see that the QuantumClifford installed there is not the one you are currently working on, but some other unrelated folder.

mgg39:
the ```~/JSoC``` folder contains the ```QuantumClifford.jl``` folder and any scratch files I work with.
It's status updates is: 
```
(JSoC) pkg> status
Status `~/JSoC/Project.toml`
  [0525e862] QuantumClifford v0.5.6 `QuantumClifford.jl#codes_framework`
```
which could be the weak link here. I'll re-point it to the current folder and see if that fixes the issue. I'll do the same for that status on my scratch folder. Thanks!

Krastanov:
```QuantumClifford v0.5.6 `QuantumClifford.jl#codes_framework` ``` indeed looks suspicious. I would suggest doing `] rm QuantumClifford` and then `] dev ./QuantumClifford.jl`.

The first command removes the installed library in that project by its name.

The second command adds a new library by its folder path (which can be slightly confusing because the folder path contains a string that happens to also be the name of the library).

We are using `dev` instead of `add` so that the project environment knows that the library will be changing a lot and should always check for changes before importing it. `add` would have just made a static copy.

If we did not use the folder path, rather we just used the name `QuantumClifford`, the library would have been downloaded from the internet -- it would have been one of the published versions, not your branch.

The `#codes_framework` part of the status report means that one very particular branch of your git repository was added, and everything was set to the branch, without tracking future changes. Your code here seems to actually be on the branch `#cleaner_ecc`, which I assume is different.

I personally compulsively do `] status` and `git status` all the time. Double checking these commands before doing anything else is an extremely good habit, especially before you start actual code debugging.

mgg39:
It seems that regardless of the folder, if I '''rm'' & then '''add'' '''./Quantumclifford.l''', I always add the '''QuantumClifford v0.5.6 `QuantumClifford.jl#codes_framework`''' version. It is indeed related to the branch '''dev''' is pulling from. After looking into the issue in existing Julia issues [https://discourse.julialang.org/t/pkg3-julia-1-0-how-to-dev-a-specific-branch-of-unregistered-package/15927](url), I looked at changing the branch with git after '''dev''' through: '''git switch cleaner_ecc''' but that didn't seem to work

Krastanov:
But that also means your git repository was at branch `dev` not at branch `cleaner_ecc` so any changes you made while working on it would be in branch `dev`, not in the branch which is being suggested for merging in this pull request. I guess that is another thing to pay attention to: a git repository can have many different branches, so make sure that if you are running a branch locally it is also the same branch you are talking about in online discussions. The only branch that we can see here online is `cleaner_ecc` because it was the only one that was pushed publicly and turned into a pull request.

Krastanov:
closing this as cleanup is happening in https://github.com/Krastanov/QuantumClifford.jl/pull/89

PR #88, opened 2023-03-19
clipping mixed states

gsommers:
Updated `canonicalize_clip!`, `entanglement_entropy(Val(:clip))` and `entanglement_entropy(Val(:rref))` to handle mixed states.


Krastanov:
Thanks, Grace!

I added some tests, added some automation for the `pure` keyword (`pure` is equivalent to `rank==nqubits`), and updated a bit the documentation.

The additions on top of your commit are here: https://github.com/Krastanov/QuantumClifford.jl/commit/0eba0dac6a2736551ab3450d1712f5eef7e10140

**Could you check that the added documentation about the possibility to give a wrong result is correctly described (for algorithm :clip)?** Could you give an example tableaux for which it fails, so that I can add it in the documentation?

Please let me know if any of these changes are clashing with something you are doing.

This should be out shortly in a new library version.

PR #89, opened 2023-03-21
New ECC

dmtrung14:


CONTRIBUTORS
```diff
@@ -7,4 +7,5 @@ Grace Sommers <gsommers@princeton.edu>
 Shu Ge <geshu@mit.edu>
 Chen Zhao <zhaochen17@mails.ucas.ac.cn>
 John Lapeyre <john.lapeyre@gmail.com>
-Yuxuan Yan <yyan.tsinghua@gmail.com>
\ No newline at end of file
+Yuxuan Yan <yyan.tsinghua@gmail.com>
+Maria Gragera Garces <m.gragera.garces@gmail.com>
```
Krastanov, line 11: you should add your name to CONTRIBUTORS as well (unless you have privacy concerns)

src/ecc/ECC.jl
```diff
@@ -0,0 +1,126 @@
...
+    function encoding_circuit end 
+
+    """Parity check tableau of a code."""
+    function parity_checks(c::AbstractECC)
+        
+    end
+
+    """The number of physical qubits in a code."""
+    function code_n(c::AbstractECC)
+        return length(c[0]) #do we even need to return in this function?
```
Krastanov, line 21: `c` is unknown here, there is some bug

src/ecc/ECC.jl
```diff
@@ -0,0 +1,126 @@
...
+    end
+    
+    """The number of stabilizer checks in a code."""
+    function code_s(c::AbstractECC) 
+        s = length(parity_checks(c))
```
Krastanov, line 26: same issue with `c`

src/ecc/ECC.jl
```diff
@@ -0,0 +1,126 @@
...
+    
+    """The number of logical qubits in a code."""
+    function code_k(c::AbstractECC)
+        k = code_n(c) - code_s(c)
+        return k
+    end
+    
+    """The rate of a code."""
+    function rate(c::AbstractECC)
+        rate = code_k(c)/code_s(c)
```
Krastanov, line 38: maybe we should use Rational here (`//` instead of `/`)

src/ecc/ECC.jl
```diff
@@ -0,0 +1,126 @@
...
+        end
+
+        convert(Vector{AbstractSymbolicOperator},naive_sc)
+
+        return naive_sc
+    end
+
+    """The distance of a code."""
+    function distance(c::AbstractECC)
+
```
Krastanov, line 68: same issue with virtual methods as mentioned above

src/ecc/bitflipcode.jl
```diff
@@ -0,0 +1,21 @@
...
+
+"""Parity check tableau of a code."""
+
+parity_checks(c::Bitflip3) = S"___
+                               _ZZ
+                               Z_Z
+                               ZZ_"
+
+#Enconding circuit ----------------------------------
+function encoding_circuit(c::Bitflip3)
```
Krastanov, line 17: a general purpose encoding circuit function would be better. This can be moved to a test file for comparisons (not high priority, probably not done in this PR)

src/ecc/steanecode.jl
```diff
@@ -0,0 +1,84 @@
...
+                              _XZZX
+                              X_XZZ
+                              ZX_XZ"
+
+code_n(c::Steane5) = 5
+
+#Encoding circuit ----------------------------------
+#https://www.researchgate.net/figure/Encoding-circuit-for-the-five-qubit-code-a-Circuit-to-encode-the-logical-minus-state_fig1_337273308
+
+function encoding_circuit(c::Steane5) #black, white
```
Krastanov, line 15: a general purpose encoding circuit function would be better. This can be moved to a test file for comparisons (not high priority, probably not done in this PR)

src/ecc/ECC.jl
dmtrung14: I did not notice that Julia defines a//b as the fraction a/b instead of the integer part of the division. However when I typed in the REPL: 
`julia> 1//2 == 0.5`
I obtained `true` anyways. So is there any particular advantage of using `//` in this case? Maybe the float will not get truncated with larger divisions is one reason?

Krastanov: Floating point is not associative:

```
julia> (0.1 + 0.2) - 0.3
5.551115123125783e-17

julia> 0.1 + (0.2 - 0.3)
2.7755575615628914e-17

julia> 1//10 + 2//10 - 3//10
0//1
```
Rationals avoid this type of numerical mistakes.

test/test_ecc.jl
```diff
@@ -0,0 +1,135 @@
...
+            @test encoding_circuit_physical == encoding_circuit_logical
+        end
+
+    end
+end
+
+
+function test_ns(c::AbstractECC)
+
+    @testset "Naive syndrome circuits" begin
```
Krastanov, line 91: here you have some random state that corresponds to a logical state of a code. You test the various different syndrome measurement circuits. Only one such circuit is currently implemented (naive_syndrome_circuit)

- Make some state consistent with the code, e.g. `s = MixedDestabilizer(parity_checks...)` and make a few copies of it, e.g., s1, s2, ...
- For s1 make the measurements in the "linear algebra" way, without having any circuit simulations. The syndrome would be something like `syndrome1 = [project!(s1, check)[...] for check in each_row_of_the_code_tableaux]`
- For s2 make the measurements in the "circuit simulation" way, e.g. by creating `Register(s2, falses(syndrome_size))` and using `apply!` or `mctrajectory!` on it and then use `bitsview` to see the content for the classical register
- Compare the two ways to measure the syndrome

Krastanov:
@dmtrung14 , I made quite a few changes. Please look one by one at the commits I added at the end of this list and make sure you understand why they are necessary. Feel free to ask questions about them. Next step would be to address the TODOs I have left in (in a new pull request).

Krastanov:
superseded by https://github.com/QuantumSavory/QuantumClifford.jl/pull/112

PR #90, opened 2023-03-31
DRY out quantum_mallows function

Shayan-P:
resolves #49

I added a helper function `sample_geometric_2` to sample a random number with respect to the desired distribution.

I removed the three versions of `quantum_mallows_int`, `quantum_mallows_float`, and `quantum_mallows_bigint` since they were almost identical (except for the random number generation for which I used the `sample_geometric_2` function).

PR #91, opened 2023-03-31
closes #49

### LINKED ISSUE #49: simplify random operator generation
Currently the code branches a lot on the size of the operator being generated (using ints for n<30, float for n<500, bigint ootherwise). ILog2 takes care of some of that branching. See how much you can simplify the code by using it. See https://github.com/Krastanov/QuantumClifford.jl/pull/47#issuecomment-1113556209=

adrianariton:
## Issue
The three functions for `quantum_mallows` (`_int`, `_bigint` and `_float`) were needed because `ilog2` had some bugs.
Now that it is fixed, they are no longer required.

## Solution
Merged the three functions into the existing `quantum_mallows` and added an if statement that is needed because of the `rand` function which cannot be written in a similar range type for Float as for Int and BigInt. Also writing it as `rand() * number` for all ranges of n slows it down drastically.

## Improvements
- Cleaner code
- A bit of time unplanned-for time improvement

### DIFF
--- a/src/randoms.jl
+++ b/src/randoms.jl
@@ -175,66 +175,26 @@ end
 
 """Sample (h, S) from the distribution P_n(h, S) from Bravyi and Maslov Algorithm 1."""
 function quantum_mallows(rng, n) # each one is benchmakred in benchmarks/quantum_mallows.jl
-    if n<30 # TODO Do in a prettier way without repetition.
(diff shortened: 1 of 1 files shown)


Krastanov:
Thank you for your contribution! Very similar pull request was submitted earlier today (#90) and I merged them already, but I deeply appreciate your contribution as well!

PR #101, opened 2023-04-11
Initial prototype for Pauli frame error simulation

amicciche:
This initial prototype performs simple Pauli frame error simulation, using for now only out-of-the-box Julia. Listed at the top of the file are some TODOs. The most important of which is to use the Tableau data structure, which is already provided by QuantumClifford.jl, to hold the frame . However, as of writing this, whether the Tableau functions the same as the Pauli frame structure when modified by a gate has not been tested. Regardless, new measurement operators will be needed since Pauli frames measure very differently (and they need reference measurements). 

Aside from that, the most important feature to add is setting the probability of an error occurring. Currently I,X,Y,Z are all equally likely.

There are also a few ways this could be speed up - replacing for loops with array operations, and using Julia symbols instead of strings for representing operations.


Krastanov:
The first things to do would be to:

- switch to using `Tableau` for the set of Pauli frames. E.g. each row in the tableau would correspond to a separate frame
- switch to using objects like sCNOT, sX, sHadamard, etc

These are relatively easy as they can just mostly reuse the `apply!` function with minor tweaks.

Then we need to deal with measurements, which would actually need some new code.

Could you try to prototype what this would look like (without using the measurements yet)?

amicciche:
Github question: I pushed new changes to the master branch on repo I forked. I assume you can see them and I don't need to open a pull request again?

I now use the tableau to hold the Pauli frames, but I used tHadamard and tCNOT. I cannot seem to figure out how to use sCNOT, or more importantly because there's no t* version, sX with apply!().  

I also went ahead and coded functionality for the measurements. The only thing left to do with them is to reset the the measured qubit's state. So for now, my code assumes that all measurements happen at the end, and then the circuit is over. 

I also added the parameter that determines how likely an error occurs.

Example output screenshot:
image

Krastanov:
> Github question: I pushed new changes to the master branch on my fork of the repo. I assume you can see them and I don't need to open a pull request again?

yes, the new commits are visible (see above this comment), as long as they have been committed to the same branch as the one referenced in this pull request.

Krastanov:
> sCNOT

tCNOT is a dense tableau representing the CNOT gate. These dense tableaux are particularly convenient when having large n qubit operations. For small "named" single- and two-qubit gates, it is frequently better to use the special sparse (symbolic) gates. E.g. `sCNOT` which is a structure that simply contains two numbers, the indices of the qubits it needs to be applied to.

So `sCNOT(1,2)` is the gate CNOT applied to qubits 1 and 2. They also work with the `apply!`  function

src/experimental/PauliFrame.jl
```diff
@@ -0,0 +1,137 @@
...
+            
+            # TODO
+            # Reset target bit's frame - the Z component is randomly 0 or 1
+            # For now, we assume nothing else will happen after a qubit is measured
+            
+        #  Hadamard 
+        elseif op[1]== :"H"
+            bit_t = op[2]
+            apply!(frame, tHadamard,  bit_t)
+        end
```
Krastanov, line 95: if you use the sparse symbolic objects, this can probably be as simple as 

```
for op in circuit
    apply!(frame, op)
end
```

Not yet super clear to me how the error insertion works.

Krastanov: also, do not do `:"name"`. That is simply equivalent to `"name"`, you are just creating a string and the `:` is redundant. If you want to create a symbol, use `:name`

src/experimental/PauliFrame.jl
```diff
@@ -0,0 +1,137 @@
...
+        
+    end
+    
+    if showFrame
```
Krastanov, line 99: I would suggest removing these flags from the final version of the function. That style is not composable (every library would have differently named flags for debugging with ad-hoc rules for what flags to be propagated).

src/experimental/PauliFrame.jl
```diff
@@ -0,0 +1,137 @@
...
+    # Randomly apply Z error with prob 1/2 on init
+    # This generates a random vector of values from 0 to the maximum binary number for the number of qubits.
+    # For example, with 2 qubits: this will be 0 through 3 or 00 01 10 11. The vector length is numframes.
+    frame.tab.xzs[2,:] = rand(0:2^(qubits)-1,numframes,1)
+    
+    if showFrame
+        println("\nInitial Pauli frame\n", frame)
+    end
+    
+    # It's nice to confirm the inputs, no?
```
Krastanov, line 41: yes, it is nice to confirm at the current development stage, but this should probably be removed from the production version as people do not want stdout to be filled with un-introspectable messages.

src/experimental/PauliFrame.jl
```diff
@@ -0,0 +1,137 @@
...
+    # whether the result of the corresponding ith reference measurement flipped
+    #               M1        M2  
+    #   F1[   M1,F1  M2,F1]
+    #  F1[  M1,F2   M2,F2]
+    # 
+    return reshape(sim_m, (numframes, num_m)), frame
+end
+
+
+if runTests
```
Krastanov, line 113: similar to other comments above, this does not need to be addressed right now, as this is just a draft. But these need to be moved to the tests folder at some point.

Krastanov:
Let's start focusing on making this not a one-off example, but something that fits with the rest of the code.

One of the first things to do would be to define a structure to represent a Pauli frame.

Not quite sure what is the best option yet, but maybe something like this:
```
struct PauliFrame
    reference::Register
    frame::Tableaux # each row is one frame
    measurements::Matrix{Bool} # in which the measurement results would be stored
end
```

We can then define an

```
function apply!(f::PauliFrame, op) 
    s=Stabilizer(f.frame)
    apply!(s, op; phases=Val(false))
    f
end
```

This leaves us with having to define `apply!` methods for noise and for measurements.

Basically, the goal is to significantly simplify your code organization and to structure things to use more idiomatic multi-methods.

Let me know what parts you would like to discuss in more detail.

amicciche:
I reworked the structure of the code to make it more API friendly, and also now any 1 or 2 qubit gate from QuantumClifford works. The two biggest deviations from what you maybe expected are:
1.) I implemented a struct and functions like you asked, but also added a function to run an entire circuit. To me, it seems like most use cases of this will be to run an entire circuit at once, but a user could also elect to just use the struct and methods directly.
2.) The measurement function, takes the reference value as an input, and returns new measurements in the form of a vector of length num_frames. Because of this, the struct doesn't store the results or the reference measurements. However, the function I wrote to process a whole circuit does take the reference values and returns a matrix of measurements.

amicciche:
Oh, I also added a file to the test directory. I got it to work on my machine, but I'm pretty sure I'm not "connecting" the test file to my src file in the proper way.

My tests were written in a simple way, and I think my next direction is to look into how to write proper tests that conform with the style of the others test file. (i.e., using things like @test)

test/test_pauliframe.jl
```diff
@@ -0,0 +1,38 @@
+
```
Krastanov, line 1: tests should use `using ...`, as if they are running a normal library, not `include`. And they should not be printing things, rather using the @test macro. You need to add this file to `runtests.jl`

src/experimental/PauliFrame.jl
```diff
@@ -0,0 +1,127 @@
...
+# For most of the operations, I manipulated the Stabilizer.tab.xzs matrix from the QuantumClifford.jl library.
+# It's a 2 by f matrtix, where f is the number of frames
+# If the first index is 1, then it refers to an X component, if 2 then Z. For Y, use both i.e. [1:2]
+# The value of [(X or Z), frame_number] is equal to a binary number that represents whether the provided index for X or Z is "on"
+#   Example: say we want to represent ZZ_Z_ . First convert to a binary string, treating the leftmost qubit as our 0 place
+#            ZZ_Z_ -> 01011 = 11 (in decimal). So to set this on frame f, do frames.tab.xzs[2,f] = 11
+# Using this, most of the manipulations of the frames were programmed, especially the injection of Pauli error channel errors.
+# # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # # #
+
+module PauliFrames
```
Krastanov, line 18: Not sure whether this needs to be a separate module. I would suggest keeping it in the same namespace as the rest of the library. Look at the bottom of `QuantumClifford.jl` where it has a ton of `include` statements for various components of the library

src/experimental/PauliFrame.jl
```diff
@@ -0,0 +1,127 @@
...
+    PauliFrames.initZ(frame)
+
+    # This is the returnable
+    sim_m = []
+
+    # Process the circuit
+    for op in circuit
+        # op wasn't measure or pauli error channel
+        if (op[1] != :MZ) && (op[1] != :PE)
+            PauliFrames.apply!(frame, op[1], op[2])
```
Krastanov, line 109: You seem to have created two different `apply!` functions -- now there is one in QuantumClifford and one in PauliFrames. Instead we want one single `apply!` function that has multiple methods.

src/experimental/PauliFrame.jl
```diff
@@ -0,0 +1,127 @@
...
+    f.frame.tab.xzs[2,:] = rand(0:2^(f.qubits)-1,f.numframes,1)
+end
+
+# applies a gate like tCNOT
+function apply!(f::PauliFrame, op::QuantumClifford.CliffordOperator, args) 
+    QuantumClifford.apply!(f.frame, op, args)
+end
+
+# applies a gates like sX, sCNOT
+function apply!(f::PauliFrame, op::DataType, args::Vector{Int})
```
Krastanov, line 43: Having `DataType` as a type constraint is quite strange. Any reason you have picked it?

`args` might very well be a tuple instead of a list.

The user would probably never be using `apply!(s, sCNOT, [1,2])`, rather `apply!(s, sCNOT(1,2))`.

amicciche: The reason for using DataType was that typeof(sCNOT) returned DataType, while typeof(tCNOT) returned CliffordOperator. I wanted to overload the apply!() function to work with either sCNOT or tCNOT. While I agree it does seem weird, my informal testing showed that it did correctly call the proper apply!() function based on the supplied argument.

If the user is only ever going to use s* gates and never t* gates, then I should delete that function that is constrained by DataType.

Krastanov: `sCNOT` is a type, not an instance of a type. All types are themselves instances of the meta type `DataType`. Thus restricting to `DataType` means you are expecting inputs like `Int64` or `Tuple{...}` to be valid inputs

 `tCNOT` is not a type, rather just one instance of the type `CliffordOperator`.

Krastanov: Yes, for now it would make sense to assume only the symbolic `s*` operators, not the dense tableaux operators `t*`

Krastanov: related to this, it would make sense to start representing your circuits more in the fashion of what the library already does. E.g. `circuit = [sCNOT(1,2), sX(3), MX(1),...]`

src/experimental/PauliFrame.jl
amicciche: I was a little confused about scope in a Julia package and how everything fit together. My goal was to use function overloading (maybe this doesn't work in Julia the way I think it does), to have apply!() behind the scenes be multiple methods, while appearing as a single function to the user.

I apologize, to me currently, function and method are interchangeable. What did you mean by one single function that has multiple methods?

src/experimental/PauliFrame.jl
amicciche: I see! That makes a lot of sense. I'll get on that!

src/experimental/PauliFrame.jl
amicciche: This was sort of my duck tape to test my code, without knowing how to properly connect it. I'll fix this with my next commit.

src/experimental/PauliFrame.jl
Krastanov: no need to apologize! The multimethod paradigm, while extremely powerful, is not well known outside functional programming and lisp circles.

In Julia:

A function is just a name in some namespace. Here you have two functions. `QuantumClifford.apply!` and `PauliFrames.apply!`. Each one of them has multiple methods (each time you "overload" it, it gets a new method).

You can even run `methods(function_name)` in order to see how many methods are implemented for a given function.

amicciche: Ohhh, so if I understand correctly, I just need to move my apply!() method(s) from the PauliFrames namespace to the QuantumClifford namespace and that would satisfy your initial comment?

Krastanov: Yup, pretty much. Removing the `PauliFrames` namespace altogether is probably easiest (and having everything in the QuantumClifford namespace). That is straightforward to do, check the include statements at the end of `QuantumClifford.jl`

amicciche: I tried to do this, but was having an issue with invoking the modified QuantumClifford from my repo and not the standard QuantumClifford installed on my machine when I tried "using QuantumClifford". Trying to fix this, I googled around and perhaps naively added

Pkg.activate(".") 

to the top of my test file. This did work and tests seem successful. But now, whenever I try to print a tableau, despite trying to activate other environments and even after trying in Jupyter notebook, and restarting my computer, instead of printing for example "X_", it prints "..". Have you come across this before? Do you have any idea what might be causing the print on a Stabilizer to behave this way? I'll attach a screenshot
スクリーンショット 2023-04-22 11 13 19


Also, to answer my main issue, do you have any suggestions on how to compile my modified QuantumClifford repo? So that I can I can do "using QuantumClifford; QuantumClifford.PauliFrameFunctionIWrote();"

adrianariton: Using this hopefully does the trick: `show(IOContext(stdout::IO, :limit => true), s)`

amicciche: Replacing print(s) with that does work, but I'm wondering if you have any idea about what may have happened, or if this is a sign of something deeper being wrong  with the installation on my machine?

Krastanov: a couple of points:

- Are you comfortable with the notion of environments in Julia (or Python for that matter)? Whenever you open a julia REPL, get in the habit of going to pkg mode (by pressing `]`) and typing `status` in order to see in what environment you are working.
- How have you installed QuantumClifford? It should be **not** added as a normal library (i.e. **not** `add QuantumClifford`), rather the repository should be cloned locally and you should do `dev ./relative/path/to/quantumclifford/repository`
- Recenty a new better printing mode was implemented for the library so that it does not print out gigantic tableaux. It seems it still needs some polishing. It detects the size of the terminal and if the terminal is too small it prints out ... It seems in your case it is detecting the size wrongly and it seems it does not have a reasonable minimal size cutoff.

amicciche: - I'm very much familiar with the notion in Python, but in practice, it's one of those things that I never really fully got down. I'll start working on that gap in my knowledge.
- Oh, thank you very very much! Just doing "dev ." allowed it to work as I intended!
- For printing, I extended my terminal to fullscreen and got the same behavior.

Krastanov: concerning the `dev .`

It is a good idea to keep the library and the environment in which you are working as two distinct entities. Currently it sounds you are treating them as one and the same. Check in the documentation for what a Project.toml and Manifest.toml files are and make sure you have a Project.toml file for your local environment and that file is NOT the Project.toml inside of the library. A typical work directory structure might look like:

```
- MyWorkEnvironment
|- Project.toml # that lists what libraries you frequently use in your work -- this should be the active environment
|- some_scratch_files.jl # files in which you keep various interactive notes and experiments, generally temporary stuff
|- SomeDevLibrary.jl # the repository for some library that you have `dev`-ed
    |- Project.toml # a non-activated project file that lists metadata and dependencies for the library
    |- the rest of the library files
```

src/experimental/PauliFrame.jl
```diff
@@ -88,22 +84,19 @@ end
...
+        if (typeof(op) != Tuple{Symbol, Vector{Int64}})
+            apply!(frame, op)      
+
         # op was Pauli error channel
         elseif op[1]==:PE
             pauliError!(frame, p, op[2][1])
-            
+
         # op was measurement in Z basis
         elseif op[1]==:MZ
```
Krastanov, line 101: you can use `sMZ` and friends from https://krastanov.github.io/QuantumClifford.jl/dev/allops/

You can just define a new `apply!(::PauliFrame, ::sMZ, ...)`.

src/experimental/PauliFrame.jl
```diff
@@ -88,22 +84,19 @@ end
...
     for op in circuit
         # op wasn't measure or pauli error channel
-        if (op[1] != :MZ) && (op[1] != :PE)
-            apply!(frame, op[1], op[2])
-        
+        if (typeof(op) != Tuple{Symbol, Vector{Int64}})
+            apply!(frame, op)      
+
         # op was Pauli error channel
         elseif op[1]==:PE
```
Krastanov, line 97: we would need to figure out a nice type abstraction for these. There is NoiseOp that might be a reasonable way to do it https://krastanov.github.io/QuantumClifford.jl/dev/API/#QuantumClifford.NoiseOp

amicciche:
Sorry, I forgot a couple things in my last commit message.  In addition to adding apply!(PauliFrame, sMZ()) and apply!(PauliFrame, PauliError()), I also added the reference measurements as well as final measurements as data fields for the PauliFrame struct. I also added an additional circuit handler called circuitSim, which perhaps is what you mentioned earlier?

test/test_pauliframe.jl
amicciche: I just pushed a reworking of the test file to use the \@test macro. My tests seem to work, but some of the other tests failed, namely test_stabs.jl. I believe its failure is completely unrelated to code I wrote, but perhaps might be due to my environment or way that I ran the runtests.jl . Are you aware of this failure?

Krastanov:
Changes look good. I will wait on the CI to complete before discussing further steps. I also added you to the github organization so that we do not need to wait on me to manually enable checks.

test/runtests.jl
```diff
@@ -25,6 +25,7 @@ end
 
 println("Starting tests with $(Threads.nthreads()) threads out of `Sys.CPU_THREADS = $(Sys.CPU_THREADS)`...")
 
+@doset "pauliframe"
```
Krastanov, line 28: Lets put this towards the bottom, as it is testing very high-level functionality. It is more easy to debug if the basic tests have already passed so you do not know you need to worry about them.

src/experimental/PauliFrame.jl
```diff
@@ -0,0 +1,196 @@
...
+
+# Examples
+```julia-repl
+julia> ghz_circuit = [sHadamard(1), sCNOT(1,2), sCNOT(1,3), sMZ(1,1), sMZ(2,2), sMZ(3,3)]
+julia> ref = [0,0,0]
+julia> frame = QuantumClifford.PauliFrame(10^6, 3, ref); QuantumClifford.initZ(frame)
+julia> f = QuantumClifford.circuitSim(frame, ghz_circuit); m = f.measurements; frame = f.frame 
+```
+"""
+function circuitSim(state::PauliFrame, circuit)
```
Krastanov, line 191: This looks very similar to `mctrajectory!`. Let's just make a new method for `mctrajectory!` instead of defining a whole new function.

src/experimental/PauliFrame.jl
```diff
@@ -0,0 +1,196 @@
...
+Calling this after initialization is essential for simulating any non-deterministic circuit.
+
+# Examples
+```julia-repl
+julia> frame = PauliFrame(numframes, qubits, ref_measurements)
+julia> initZ(frame)
+```
+"""
+function initZ(f::PauliFrame)
+    f.frame.tab.xzs[2,:] = rand(0:2^(f.qubits)-1,f.numframes,1)
```
Krastanov, line 77: 1. I suspect this works only for <64 qubits
2. This would be relatively slow due to the need to allocate a large temporary array. Let's use an explicit loop instead
3. I would have thought that just using `UInt64` instead of `0:2^(f.qubits)-1` would be good enough (and faster). Generating one random uint is simpler than making sure we cover only the first f.qubits bits.
4. For 3, instead of hardcoding UInt64, we should use `eltype` to make sure we are using the correct type

src/experimental/PauliFrame.jl
```diff
@@ -0,0 +1,196 @@
...
+julia> frame = PauliFrame(numframes, numqubits, [0,1]])
+
+julia> apply!(frame, sMZ(5,1)) 
+julia> apply!(frame, sMZ(1,2)) 
+
+The first ['sMZ']@ref measured qubit 5 and treated 0 as its reference measurement.
+The second ['sMZ']@ref measured qubit 1 and treated 1 as its reference measurement.
+```
+"""
+struct PauliFrame
```
Krastanov, line 43: to ensure type stability, let's add a type parameter:
```
struct PauliFrame{T} where T
    ...
    frame::T
    ...
end
```

src/experimental/PauliFrame.jl
Krastanov: If this function is just a slight modification of `pauliFrameCircuitHandler`, then it probably should be the same function with two different methods.

Thus let's think of how we can turn the `pauliFrameCircuitHandler` and `circuitSim` functions into methods of the `mctrajectory!` function

Krastanov: or of `mctrajectories`

test/runtests.jl
amicciche: I did this because some of the existing tests always fail, and I wanted to see my result before an error was thrown. 
Specifically, stabs and bitpack always fail for me.

src/experimental/PauliFrame.jl
amicciche: 1.) Yes, I found out the hard way today that this doesn't work for more than 63 qubits. I'm working on that now.
2.) Avoiding the loop was my python brain thinking - I'll fix that with 1.)

test/runtests.jl
Krastanov: oh, that is worrysome... they do not seem to fail on the master branch or in the CI of this pull request. Do they fail only locally for you?

amicciche: I cloned the master repo, activated a new Julia environment, ran the testfile and still got an error. If you look at the top row of the stabilizers its comparing the sign is off:
スクリーンショット 2023-05-02 21 11 38

amicciche: That screenshot is from my fork, but I got the same off by a sign from the master repo:
スクリーンショット 2023-05-02 21 22 00

amicciche: Here's all of the output. On this run, I couldn't spot anything obvious like a sign off in the first line, but it still failed
[teststab_fail.txt](https://github.com/QuantumSavory/QuantumClifford.jl/files/11378800/teststab_fail.txt)

Krastanov: welp, this is worrisome. I executed that test a few hundred times just now and it is not giving any errors.
Are there any simpler tests that fail on your machine that we can use to track down the issue?

amicciche: Oh no... In the runtest.jl suite, only this and bitpack fail (this might be something different though):
[bitpack_fail.txt](https://github.com/QuantumSavory/QuantumClifford.jl/files/11378882/bitpack_fail.txt)

src/experimental/PauliFrame.jl
```diff
@@ -73,9 +73,17 @@ julia> frame = PauliFrame(numframes, qubits, ref_measurements)
...
-function initZ(f::PauliFrame)
-    f.frame.tab.xzs[2,:] = rand(0:2^(f.qubits)-1,f.numframes,1)
-    return f
+function initZ(frame::PauliFrame)
+    z_index = (2+frame.qubits÷65, 2*(1+frame.qubits÷65))
+    for f in 1:frame.numframes 
+        frame.frame.tab.xzs[z_index[1]:z_index[2],f] = rand(UInt64, 1 + frame.qubits÷65, 1)
+        # This line removes garbage values from the the bit spill over. For example if we had 68 qubits,
+        # then we want a random 64bit number in the first cell and a random  4 bit number in the second. Not two 64 bit numbers. 
+        # TODO make this line more readable if possible
```
Krastanov, line 82: maybe we can just skip that line altogether? After all, these bits are never really used for anything

src/experimental/PauliFrame.jl
```diff
@@ -73,9 +73,17 @@ julia> frame = PauliFrame(numframes, qubits, ref_measurements)
 julia> initZ(frame)
 ```
 """
-function initZ(f::PauliFrame)
-    f.frame.tab.xzs[2,:] = rand(0:2^(f.qubits)-1,f.numframes,1)
-    return f
+function initZ(frame::PauliFrame)
+    z_index = (2+frame.qubits÷65, 2*(1+frame.qubits÷65))
+    for f in 1:frame.numframes 
+        frame.frame.tab.xzs[z_index[1]:z_index[2],f] = rand(UInt64, 1 + frame.qubits÷65, 1)
```
Krastanov, line 79: this still allocates intermediary arrays (e.g., as can be seen with `@btime`).

The fastest and simplest solution (unless we start thing about SIMD RNGs) is along the lines of

```
@inbounds @simd for i in CartesianIndices(@view(xzs[1:end÷2,:]))
           xzs[i] = rand(UInt64)
end
```

The `CartesianIndices` is a fast way to look at the array as if it is flat. The view is in order to work on only the first half of the array (maybe I wrote something wrong there). The simd and inbounds macros are to tell the compiler that loop iterations are independent and can be done in any order and that it does not need to check the bounds of an array.

amicciche: I added inbounds and simd to my method, but using CartesianIndices made the performance much worse. The method actually creates random X's  but nevertheless demonstrates that it is much slower. Here are the two functions I compared:
image

I then created a PauliFrame with 10^3 frames and 10^4 qubits. I created a clean frame before I tested each function. Here are the results:
image
image

src/experimental/PauliFrame.jl
amicciche: Sorry, I don't seem to understand what this is suppose to achieve? All the data fields of PauliFrame have type restrictions, and currently:

-     numframes::Int
-     qubits::Int
-     ref::Vector{Bool}

are the values that need to be provided to the constructor.

src/experimental/PauliFrame.jl
amicciche: I changed `circuitSim` to `mctrajectory!`, but decided to keep `pauliFrameCircuitHandler` because its slightly different but not so specific parameter list might cause a problem in the future and because it creates and initializes a PauliFrame object. 

My intention was that  `pauliFrameCircuitHandler` is a convenience function, but `mctrajectory!` is _the_ function to apply an entire circuit in one go.

I could get rid of `pauliFrameCircuitHandler`, leave it alone, rename it, or change it to be a wrapper for `mctrajectory!`, just creating the PauliFrame object and then calling `mctrajectory!`.

amicciche:
I just added some benchmarks. It's in its own file just for now. Moving the code into `benchmarks.jl` should take almost no time, but I wanted some confirmation before I start editing other files. Also because I defined a function to create the circuit. 
Here were the results:
スクリーンショット 2023-05-04 23 12 36
On the top is 10^3 Hadamards and 10^3 CNOTs on a 1001 qubit circuit with 1 measurement handled by `mctrajectory` on a register. (This was the only data type I could get to work on a whole circuit including measurements.)

On the bottom is 10^6 Hadamards and 10^6 CNOTs on a 10^6+1 qubit circuit handled by `pauliFrameCircuitHandler`. Furthermore, it held 100 frames, in effect simulating this situation 100 times.

amicciche:
I just added some benchmarks. It's in its own file just for now. Moving the code into `benchmarks.jl` should take almost no time, but I wanted some confirmation before I start editing other files. Also because I defined a function to create the circuit. 
Here were the results:

スクリーンショット 2023-05-04 23 41 11


On the top is 10^3 Hadamards and 10^3 CNOTs on a 1001 qubit circuit with 1 measurement handled by `mctrajectory` on a register. (This was the only data type I could get to work on a whole circuit including measurements.)Below that is 10^4 Hadamards and 10^4 CNOTs on a 10^4+1 circuit handled the same way.

Second from the bottom is 10^3 Hadamards and 10^3 CNOTs on a 10^3+1 qubit circuit handled by `pauliFrameCircuitHandler`. Furthermore, it held 1000 frames, in effect simulating this situation 1000 times. On the very bottom is 10^6 Hadamards, CNOTS, and qubits handled again by `pauliFrameCircuitHandler`, this time with 100 frames.

src/experimental/PauliFrame.jl
Krastanov: `Stabilizer` is not a concrete type, as itself it has a type parameter. To have a concrete type it needs `Stabilizer{T}`.

You can see this if you type `S"X" |> typeof`.

If the compiler is given a concrete type it can be more restrictive with the memory layout it uses, have fewer boxed objects, and provide better machine code.

It is a quirk of julia related to how type parameters work.

Here are a few more examples:

```
julia> isconcretetype(Array)
false

julia> isconcretetype(Vector)
false

julia> isconcretetype(Vector{Int})
true
```

Krastanov:
Could you add a documentation page about this (in the `docs` folder) and move most of what is in comments to properly formatted documentation. Generally, try to use docstrings or documentation, not comments, when possible (at top of files or above function definition). That way users can more easily see what is happening. The very comments inside of functions are pretty much the only thing that should remain as comments.

Do a final round of polish along these lines (removing leftover comments, questions, etc) and I will give it a final round of review, do some quick edits, and merge it.

Krastanov:
I am reworking this tonight, aiming to have it polished and merged this week. I would like you to look through the commits I am making as there are a lot of "software engineering" issues to be fixed. While the algorithm you implemented works great, there is still a lot to learn in terms of how to organize and document code that is meant to be a high-quality public research tool. Some of the things are minor and stylistic, but even they are important to get used to.

Krastanov:
you can read each commit separately by clicking on it in this timeline

Krastanov:
I will write commit messages with explanations for some of the commits -- read them and if there is something unclear discuss it in a github comment on the given commit

Krastanov:
`pauliFrameCircuitHandler` is a weird name for a function in julia. It is important to follow the typical conventions of the language so that someone else reading your code does not get confused:

- in julia CamelCase is for types and snake_case is for functions
- in python that is true as well
- in javascript camelCase with a lower-case for the first letter is common for functions

Krastanov:
The documentation and examples rely on `circuitSim` which does not exist

Krastanov:
concerning initZ!, you were creating temporary vectors of random numbers inside of a loop: that would definitely cause significant slowdown. Allocating temporary data is one of the greatest sources of slowdowns.

Krastanov:
avoid writing things like:

```
a = some_function(); some_other_thing()
```

Makes things harder to read (it is unidiomatic) and probably points to a place where it helps to write a cleaner API for better UX

Krastanov:
I see you have been making all these changes directly on your master branch: in the future, always make a separate branch on which to develop a feature. Otherwise it becomes very difficult to sync up with other changes to the original repository (e.g., after this is merged, you might need to force reset your master branch, because it would have deviated significantly from the original master branch).

Krastanov:
Check out these two (very similar) essays on how to write documentation:

- https://documentation.divio.com/
- https://diataxis.fr/

They would be extremely useful also in the future when writing papers or when teaching.

Krastanov:
The indexing both in the `initZ!` (already fixed) and in `apply!(::PauliFrame, ::PauliError)` is convoluted or maybe wrong. Check the corrected versions when they are done. There should not be any special case for sizes bigger-or-smaller than 64 qubits.

Krastanov:
use `enumerate` not

```
i = 1
for element in array:
    ...
    i+=1
end
```

Krastanov:
As a convention, fields of structs are considered private. Users should not be encouraged to do things like `object.field`, rather there should be multimethods defined that access fields. That way if we need to change something in a future version, we will not break user code.

Similarly, fields should not be documented in the parts of the documentation meant for users (but certainly, it is great to document them in the parts of the docs that are meant for other developers)

Krastanov:
executable code in docstrings should have value as an example, showing something interesting. For instance, consider the docstring for `initZ!` which contains the following:

```
julia> frame = PauliFrame(numframes, qubits, ref_measurements)
julia> initZ!(frame)
```

This is not a great example. It does not show anything new -- users already know the function can be called, so they do not learn from the example.

On a separate notes, examples should almost always be in the doctest format, so that the package tester executes them and automatically detects if they are outdate.

Krastanov:
Documentation should be carefully formatted, and if possible it should follow the style of the rest of the library. In a lot of places you use things like 

```
# Example

...
```

On its own this is not bad, but:

1. the rest of the library does not have that style of sections in the docstrings
2. You are using a single `#`, i.e. a h1 html tag, i.e. the largest possible header, usually meant for titles of pages. Such large typeface should appear only at the top of pages

Krastanov:
The text in the docstrings seems rather arbitrarily wrapped. Aim to either:

- have a new line at the end of each sentence (i.e. new line per semantic unit)
- have a new line so that the text is neatly wrapped at 80ish characters (common convention in Python)

Generally, make the code and docstrings look pretty to the eye.

This can also be automated in your editor (anything menial can be automated)

Krastanov:
be careful with markdown syntax

Occasionally you have things like `['sMZ']@ref` - this contains the wrong type of quotes and the reference link is not generated because the link syntax is broken (the `@ref` should be in parentheses)

Krastanov:
My comment above about the `# Examples` being rendered too large is wrong. What you had written was properly formatted -- it seems Documenter.jl takes care of formatting.

Krastanov:
A lot of the documentation was outdated, referring to things that existed only in your early version. Stuff like that should be cleaned up as the code evolves.

Issue #45, opened 2022-03-25
Implement more two-qubit symbolic gates


Krastanov:
Like `sCNOT` and `sSWAP` in https://github.com/Krastanov/QuantumClifford.jl/blob/master/src/symbolic_cliffords.jl

anonymousr007:
Hi @Krastanov, here is what you mean about two-qubit symbolic gates?

Krastanov:
Specific "named" two-qubit gates can be special-cased so that they are much faster than the generic tableau gate. E.g. compare the performance of applying a CNOT as a tableau (`apply!(stab, tCNOT, [1,2])`) and as a special cased operator (`apply!(stab, sCNOT(1,2))`). The later is much faster but it has to be implemented separately. Same as `sSWAP` and other ones in the aforementioned file. Implementing other special cased two-qubit gates would be very helpful.

adrianariton:
Hello! I made a small list of 2 Qubit operators, and I was wondering if there'd be any more to add to it. I will then start implementing each.

This is what I gathered:
`XNOR` `DCNOT` `iSWAP`.

I assume that the goal is to implement Clifford gates for now, so SQRTSWAP and ones alike are not the target, but if I'm wrong correct me:)!

Krastanov:
There are 24 single-qubit gates and 11000 two-qubit gates. Pretty much any single-qubit gate can be added and many multi-qubit gates can be added as long as they have a reasonable name. The stim library is a good place to compare to.

Krastanov:
e.g. this list in stim https://github.com/quantumlib/Stim/blob/main/src/stim/stabilizers/tableau_transposed_raii.cc

adrianariton:
Thank you so much for the input. I've started working on them and found quite a large list of 2 qubit operators! I am in the process of implementing all of them, and will finish this week.

Fe-r-oz:
Dear Stefan,

Thank you very much for your valuable feedback on PRs #336 and #368.  I would like to confirm whether we can now consider this issue resolved 😅? I have incorporated all the remaining two-symbolic gates in alignment/comparison with [Stim’s two-qubit gate documentation](https://github.com/quantumlib/Stim/blob/main/doc/gates.md). Thank you for considering my request.

PR #104, opened 2023-04-19
Added X and Z Controlled symbolic gates

adrianariton:
# Regarding issue #45

I will soon be adding more gates, until there are enough to merge. 

This week I will add `Y` controlled gates as well as `SQRT` gates and their inverses.

### DIFF
--- a/controlled_testing.jl
+++ b/controlled_testing.jl
@@ -1,0 +1,46 @@
+include("../QuantumClifford/src/QuantumClifford.jl")
+using Test
+using .QuantumClifford: random_stabilizer, random_destabilizer, @P_str,  @S_str, tCNOT, sCNOT, apply!, sDCNOT, random_pauli, sSWAP, sXNOR, sX, X, sZCX, sZCZ, sHadamard, sInvPhase, tHadamard, sZCY, sPhase, tPhase, sXCY, sYCZ, sYCX, sYCY, sXCX, sXCY, sXCZ
+
+function transform_Zbasis(qubit)
+    transformations = Dict(:X => [sHadamard(qubit),], :Y => [sInvPhase(qubit),sHadamard(qubit)], :Z => [sHadamard(qubit), sHadamard(qubit)])
+    inverse_transformations = Dict(:X => [sHadamard(qubit),], :Y => [sHadamard(qubit), sPhase(qubit)], :Z => [sHadamard(qubit), sHadamard(qubit)])
+    return transformations, inverse_transformations
+end
+
+function transform_Xbasis(qubit)
+    transformations = Dict(:Z => [sHadamard(qubit),], :Y => [sInvPhase(qubit),], :X => [sHadamard(qubit), sHadamard(qubit)])
+    inverse_transformations = Dict(:Z => [sHadamard(qubit)], :Y => [sPhase(qubit),], :X => [sHadamard(qubit), sHadamard(qubit)])
+    return transformations, inverse_transformations
+end
+
+function get_gates(control, target)
+    transform1, inverse1 = transform_Zbasis(1) # get gates which swith control from Z basis to specified
+    transform2, inverse2 = transform_Xbasis(2) # get gates which swith control from X basis to specified
+    return cat(transform1[control], transform2[target], dims=(1)), cat(inverse2[target] , inverse1[control], dims=(1))
+end
+
+function walk(gates, state)
+    cstate = copy(state)
+    for gate in gates
+        cstate = apply!(cstate, gate)
+    end
+    return cstate
+end
+
+@testset "Controlled gates" begin
+    # test ten time for each
+    for i in 1:10
+        for control in (:X, :Y, :Z)
+            for target in (:X, :Y, :Z)
+                random_state = random_destabilizer(2)
+                implemented_gate = eval(Symbol(:s,control,:C,target))(1,2)
+
+                tr, inv = get_gates(control, target)
+                # print(tr, inv)
+                test_gates = cat(cat(tr, sCNOT(1,2), dims=(1)), inv, dims=(1))
+                @test apply!(copy(random_state), implemented_gate) == walk(test_gates, copy(random_state))
+            end
+        end
+    end
+end;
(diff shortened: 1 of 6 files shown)


Krastanov:
Thank you! These look great! My main worry right now is figuring out a way to test them for correctness. Is there some relationship between the name and an equivalence to what the gate is supposed to look like when decomposed into:
image

adrianariton:
Yes, there is! I will provide each decomposition today as comments above the macros and also as a comment here!

Krastanov:
no need to add extra comments to the file, but maybe we can discuss it in this thread so we can figure out a good way to do tests?

adrianariton:
Ok! Here are the decompositions of each (so far) implemented gate! I will soon add more!


`XCX` - H(1), CNOT (1,2), H(1)
`XCY` - H(1), S3(2), CNOT(1,2), H(1), S(2)
`XCZ` - CNOT(2,1)

`YCX` - S3(1), H(2), CNOT(2,1), S(1), H(2)
`YCY` - in progress
`YCZ` - S3(1), CNOT(2,1), S(1)

`ZCX` - equivalent for CNOT(1,2)
`ZCY` - S3(2), CNOT(1,2), S(2)
`ZCZ` - H(2), CNOT(1,2), H(2)

`XNOR` - X(1), CNOT(1,2), X(1)
`DCNOT` - CNOT(1,2), CNOT(2,1)

Krastanov:
Could you elaborate on what the logic of the `{X,Y,Z}C{X,Y,Z}` decomposition is? I am uncomfortable relying on just a dictionary, rather it would be nice if we can write tests from first principles. Something like:

```
for control in (:X, :Y, :Z)
    for target in (:X, :Y;, :Z)
        random_state = random_destabilizer(2)
        implemented_gate = eval(Symbol(control,:C,target))(1,2)
        test_gates = (???, sCNOT(1,2), ???) # these question marks need to be filled in with something that depends on `control` and `target`
        @test implemented_gate*random_state == test_gates[3]*(test_gates[2]*(test_gates[1]*random_state))
    end
end
```

Krastanov:
From your notes it seems you have already implicitly done that: there is conjugation with H to change to X basis and conjugation with S to go to Y basis. Having some neat short tests for that would be very helpful.

adrianariton:
Yes, you're right the tests would be easier this way. So the main idea of the notation is that the gate ACB flips the second qubit's components wrt base B depending on the state of the first qubit's. (Acting like `CNOT`, but from a different base to a different base). **The notation for CNOT would be `ZCX`, and the following are true**:
- To 'send' the i'th qubit from base `Z` to `X` we commute by `H(i)`
- To 'send' the i'th qubit from base `Z` to `Y` we commute by `Sdg(i)*H(i)` (or `S3(i)*H(i)` as I notated it in the previous message)
- As a result one can 'send' a qubit from base `X` to `Y` by commuting with `Sdg(i)`

### This can be performed as such:

Let's say we want to find YCY:
- We know `ZCX` == `CNOT`
- We need to send the first qubit from computational Z basis to Y basis: so we commute by `S3(1)*H(1)`
- We need to send the second from X to Y so we commute by `S3(2)`

We arrive at `YCY` := `S3(2) * S3(1) * H(1) * CNOT(1,2) * H(1) * S(1) * S(2)`

### Another example would be finding `XCX`:
- We just need to get the first qubit from Z Basis to X to X basis, so we commute by `H(1)`
We get: `XCX` := `H(1) CNOT(1,2) H(1)`, since `ZCX` == `CNOT`.

### One more interesting example is `YCX`:
- We commute with `S3(1) H(1)` to go from Z to Y Basis on qubit 1

We get `S3(1) H(1) CNOT(1,2) H(1) S3(1)` which is equivalent to the result from the prev message: `S3(1), H(2), CNOT(2,1), S(1), H(2)`, because `H(1) H(2) CNOT(1,2) H(2) H(1)` is `CNOT(2,1)`

adrianariton:
I will soon be committing with the `YCY` gate and try to add the tests for the `{}C{}` gates, and also begin the `SQRT` gates!

adrianariton:
Hello!
- I added a new testing file as a sketch on how the testing will work.
- And I also have a question: Do I need to add this to the test symcliff file, or to create a new one only for controlled gates?

Krastanov:
This looks great! You can just add the tests as a new test file.

adrianariton:
I added the testing file (symcontrolled), and also hopefully correctly deleted that include that I needed for local testing [you can see it in the difference between the two last commits]. I also remembered that I need to export the gates in QuantumClifford.jl. I will do this right now!

Krastanov:
This looks great! I will look a bit more into it later over the weekend.

adrianariton:
Seeing that some tests did not work made me realise I forgot to remove DCNOT and XNOR, as they do not belong to the {X,Y,Z}C{X,Y,Z} set! Now the tests should work fine!

Krastanov:
Hi, @adrianariton ! I made a few small changes to fit the style of the library a bit closer. I also added a test for the change of basis operations themselves. There is one piece that I am not sure I understand. Consider for instance the following change of basis:

```
mctrajectory!(S"X", transform_Xbasis(1)[1][:Y])
```

This returns -Y, while I would have expected it would return +Y. Could you share if you have consulted any other resource on these gate implementation? I imagine there might be different conventions, I just want to have a reference to which conventions we have picked.

Krastanov:
If I have borked up the tests about basis changes, feel free to modify them as well.

adrianariton:
Regarding the -Y thing. I think my notation was a little misleading! As you can see in the tests I swapped `forward` and `backward` (lines 53,55). That is because transform_Zbasis, does not take the Z basis to a given basis, but does the exact opposite: it takes that basis to the computational basis! 

I now realise that the names don't really confer this, but this is the idea:

To measure a qubit in a given basis, one needs to take that basis to the Computational Z basis, hence the inversion of forward/backward.

If it would be necessary, I shall change the names to the 2 functions to something more suggestive!

adrianariton:
Also, the tests should now work!

PR #109, opened 2023-04-30
Added apply_right! function for Hadamard gate

FalafelGood:
Hi there, 

This is my first PR for a code repo that's not my own! Super keen to be contributing!

Anyways, here's an apply_right! function for the Hadamard gate which [Krastanov](https://github.com/Krastanov) recommended I start off with. Here's how it works:

As Craig Gidney describes in his [Stim paper](https://arxiv.org/abs/2103.02202): 
> "inplace prepending (note: equivalent to right multiplication of operators) $B$ into $A$ is done by computing the result of conjugating each of $B$'s generators' outputs by $A$. For each generator $g_q$ in $B$ we compute $g'_q = A^{-1}B^{-1}g_q B A$

Let $H_i$, $X_j$, $Z_k$ be a Hadamard, X and Z gate acting on qubits $i, j, k$ respectively. Taking conjugations of the stabilizer generators $\\{X_i, Z_i \\}$ for all qubits, we find that:

$$ H_i Z_i H_i  = X_i $$

$$ H_i X_i H_i  = Z_i $$

$$ H_i X_j H_i = X_{j \neq i} $$ 

$$ H_i Z_j H_i = Z_{j \neq i} $$

Therefore, the effect of a right Hadamard product on the $k$ th qubit is to swap the tableau columns of the $k$ th $Z$ and $k$ th $X$ stabilizer column.

I'm not sure what the practice is for submitting unit tests is, or whether they're strictly necessary for small functions like this. Regardless, here's a quick sanity check. Let me know if I should add this (or something more robust) to the unit tests.

```
tCNOT * (tHadamard \otimes tId1)
```
Which yields
```
X₁ ⟼ + XX
X₂ ⟼ + _X
Z₁ ⟼ + Z_
Z₂ ⟼ + ZZ
```

Testing against my function
```
myop = tCNOT
H1 = sHadamard(1)
apply_right!(myop, H1)
```

Yields
```
X₁ ⟼ + XX
X₂ ⟼ + _X
Z₁ ⟼ + Z_
Z₂ ⟼ + ZZ
```

Let me know if there's anything else I can clarify.

AMDG

### DIFF
--- a/src/symbolic_cliffords.jl
+++ b/src/symbolic_cliffords.jl
@@ -333,6 +333,16 @@ function apply_single_y!(stab::AbstractStabilizer, i)
     stab
 end
 
+"""Apply a right Hadamard on a CliffordOperator"""
+function apply_right!(op::CliffordOperator, H::sHadamard)
+    s = tab(op).xzs
+    n = nqubits(op)
+    @inbounds @simd for k = 1:size(s,1)
+        s[k, H.q], s[k, H.q + n] = s[k, H.q + n], s[k, H.q] # Swap column X(i) with Z(i)
+    end
+    op
+end
+
 ##############################
 # Measurements
 ##############################


Krastanov:
Hi, Hudson! Thank you for starting this!

Let's start it out in a separate file as when it is finished it will grow into a pretty large set of functionalities. Maybe we can call it `symbolic_cliffords_rightapply.jl` and include it after `symbolic_cliffords.jl`.

A new set of tests would definitely be needed. It does not really matter whether the function is small or larger, rather the fact that there is a bunch of hard-coded behavior here and a minor typo would make it completely wrong. We want some form of "consistency check" -- not simply copying what we have already implemented as a test (that would not be useful), but rather comparing it against something completely independent (e.g. the already implemented left multiply).

Maybe something like:

```
for a in all_sym_gates
    for b in all_sym_gates
        @test a left_multiply b == b right_multiply a # this needs to be expanded into actual code
    end
end
```

That would also be a new test file.

Lastly, this is a fine way to write the Hadamard apply_right, but at some point the goal would be to have an apply_right for all symbolic cliffords. If we write them all in the current style that would be a very significant code repetition (and it would not be taking advantage of possible multithreading, etc). The solution already employed for the left apply is here: https://github.com/QuantumSavory/QuantumClifford.jl/blob/8273233c951aad678fd0bab2a94e2b9ed6782c7d/src/symbolic_cliffords.jl#L49-L81

Do not worry about that last part though, lets first flesh out the tests and add one or two other operators to see how things are shaping.

FalafelGood:
Cool, thanks for the advice! I'll go ahead and get those things started now.

> Lastly, this is a fine way to write the Hadamard apply_right, but at some point the goal would be to have an apply_right for all symbolic cliffords. If we write them all in the current style that would be a very significant code repetition (and it would not be taking advantage of possible multithreading, etc). The solution already employed for the left apply is here:
> 
> https://github.com/QuantumSavory/QuantumClifford.jl/blob/8273233c951aad678fd0bab2a94e2b9ed6782c7d/src/symbolic_cliffords.jl#L49-L81

Great, that was my gut feeling. I won't worry about this now but I will ask one question: 

Naturally, we want to hardcode common operations to maximize performance. This means that each common op requires a custom method, which I assumed meant unavoidable boilerplate code. Yet this doesn't seem to be the case for left multiply! How is it possible for an operation like _apply! is able to capture the nuanced optimization of {Hadamard, Phase, X, etc.} yet still be contained in a single function?

Krastanov:
The `_apply!` function is dispatched only on `AbstractSingleQubitOperators` because of its type signature: 

https://github.com/QuantumSavory/QuantumClifford.jl/blob/3604bc159b27835bc234509069e1480662ff7039/src/symbolic_cliffords.jl#L49

This happens thanks to the multimethod facilities of julia (see https://www.youtube.com/watch?v=kc9HwsxE1OY)

Inside of that `_apply!` method there is a dispatch to `qubit_kernel`:

https://github.com/QuantumSavory/QuantumClifford.jl/blob/3604bc159b27835bc234509069e1480662ff7039/src/symbolic_cliffords.jl#L55

That is the **only** function that needs methods defined for each concrete type.

But even that gets to be simplified and boilerplate gets removed thanks to the use of macros. The kernel gets defined in lines like 

https://github.com/QuantumSavory/QuantumClifford.jl/blob/3604bc159b27835bc234509069e1480662ff7039/src/symbolic_cliffords.jl#L55

The `@qubitop1` macro defines all the necessary boilerplate:

https://github.com/QuantumSavory/QuantumClifford.jl/blob/3604bc159b27835bc234509069e1480662ff7039/src/symbolic_cliffords.jl#L63-L74

You can yourself see that macro in action by using the `@macroexpand1` introspection tool:

```
julia> using QuantumClifford

julia> @macroexpand1 QuantumClifford.@qubitop1 Hadamard (z_expression, x_expression, phase_expression)
```
which outputs
```julia
quote
    struct sHadamard <: QuantumClifford.AbstractSingleQubitOperator
        q::QuantumClifford.Int
    end
    @doc "A \"symbolic\" single-qubit Hadamard. See also: [`SingleQubitOperator`](@ref), [`AbstractSymbolicOperator`](@ref)" sHadamard
    @inline qubit_kernel(::sHadamard, x, z) = begin
        (z_expression, x_expression, phase_expression)
    end
end
```

Krastanov:
Could you try to define your `apply_right!` to internally use `qubitkernel_right` and then define that kernel for the Hadamard gate? Do not worry about the macros for now. Let's have two or three gates implemented in that fashion (a single `apply_right!` and multiple different `qubitkernel_right` for each) and then we can think about how to remove even more boilerplate.

Krastanov:
Hi, @FalafelGood ! Thanks for contributing this! It has been quiet for a while, so I will close it for now, but feel free to re-open it if you end up having the time to work on it. The help is certainly valued!

Issue #64, opened 2022-07-05
Conversion of `Stabilizer` to canonical `CliffordOperator`


Krastanov:
`CliffordOperator(s::Stabilizer) = CliffordOperator(tab(MixedDestabilizer(s)))` would be wrong because the row operations involved in creating a MixedDestabilizer changes the order of the qubits. Create a version of `canonicalize_gott!` that knows how to undo row operations and use it for the creation of `CliffordOperator`. Check that there are no lingering bugs in `logdot`.

adrianariton:
Hello again:)! I wish to start working on this issue!
It is a little bit unclear to me how MixedDestabilizers' row operations permute the bits! Could you explain it to me please?
Best regards, Adrian!

Krastanov:
Hi, Adrian,

Here is an example:

The ground state of a two qubit system is |00⟩ written as a tableau as `S"ZI IZ"`

The following stabilizer state, if interpreted as "the operation that maps the ground state to the given state" should be interpreted as the SWAP gate:
```
julia> S"IZ ZI"
+ _Z
+ Z_
```

However, we can not define an operation just by giving how it maps Zs. There is some arbitrary choice to be made about phases, but the complete operation should look like

```
X₁ ⟼ + _X
X₂ ⟼ + X_
Z₁ ⟼ + _Z
Z₂ ⟼ + Z_
```

That means we need to find the destabilizer operators if we are given only the stabilizer part of the tableau. There is a standard way to do that (the `MixedDestabilizer` constructor does it by using `canonicalize_gott!`) , but the canonicalization operation involved requires a ton of row operations (swaps and multiplications) that have to be tracked and reversed. The tracking and reversing is not happening currently as you can see from the example below: 

```
julia> MixedDestabilizer(S"IZ ZI")
𝒟ℯ𝓈𝓉𝒶𝒷
+ X_
+ _X
𝒮𝓉𝒶𝒷
+ Z_
+ _Z
```

PR #110, opened 2023-05-01
Conversion of Stabilizer to canonical CliffordOperator (resolving #64)

adrianariton:
The main thing I thought would be a first try to make the implementation work is just to store the operations as they are performed (swapping and pivoting) and undo them. There is deffinetly a better way of doing this by building the inverse matrix to undo the gaussian elimination iteratively and I will try to implement it later this week if this is the way of doing it.

My only issue right now is that I think modifying the `canonicalize_gott!` return argument messes it's other calls (from what I see it is used multiple times, in `QuantumClifford.jl` and `graphs.jl`). 

Also, I see that by the current return of `canonicalize_gott!`, only the x and z permutations get returned for them to be inverted later, but not also the `mul_left` operations. To also invert the `mul_left` operations, my first thought was to add this vector of operations.

I also rewrote the `undoperm` in MixedDestabilizer(Stabilizer), and on a few test I randomly selected and inputed manually it seems to work!

- ~~**Edit**: Will fix: Used `s` instead of `r + s`. [QuantumClifford.jl, lines 749-779 undoperm]~~ `DONE`
- **Edit 2**: Phases get a bit scrambled, probably from commutating row-operations. I will fix this soon by changing mul_left (lines
768-769) to it's inverse! `DONE` `and also commented it, because I don't know if it is necesarry to reverse multiplication ops`

### DIFF
--- a/src/QuantumClifford.jl
+++ b/src/QuantumClifford.jl
@@ -713,7 +713,7 @@ end
 # Added a lot of type assertions to help Julia infer types
 function MixedDestabilizer(stab::Stabilizer{T}; undoperm=true) where {T}
     rows,n = size(stab)
-    stab, r, s, permx, permz = canonicalize_gott!(copy(stab))
+    stab, r, s, permx, permz, ops = canonicalize_gott!(copy(stab))
     t = zero(T, n*2, n)
     vstab = @view tab(stab)[1:r+s] # this view is necessary for cases of tableaux with redundant rows
     t[n+1:n+r+s] = vstab # The Stabilizer part of the tableau
@@ -747,7 +747,22 @@ function MixedDestabilizer(stab::Stabilizer{T}; undoperm=true) where {T}
         t[n+r+s+1:end] = sZ # The other logical set in the tableau
     end
     if undoperm
-        t = t[:,invperm(permx[permz])]::T
+        for op in reverse(ops)
+            # print("Performing: ", op)
+            if op[1] == 1
+                rowswap!(t, op[2], op[3]; phases=Val(true))
+                rowswap!(t, op[2] + s, op[3] + s; phases=Val(true))
+
+            elseif op[1] == 2
+                mul_left!(t, op[2], op[3]; phases=Val(true))
+                mul_left!(t, op[2] + s, op[3] + s; phases=Val(true))
+
+            elseif op[1] == 3
+                t = t[:,invperm(permx)]::T
+            elseif op[1] == 4
+                t = t[:,invperm(permz)]::T
+            end
+        end
     end
     MixedDestabilizer(t, r+s)::MixedDestabilizer{T}
 end
--- a/src/QuantumClifford.jl
+++ b/src/QuantumClifford.jl
@@ -703,6 +703,9 @@ When the constructor is called on an incomplete [`Stabilizer`](@ref) it
 automatically calculates the destabilizers and logical operators, following
 chapter 4 of [gottesman1997stabilizer](@cite).
 
+If `undorowops` is specified, the constructor will track and undo the row operations
+performed for gaussian elimination.
+
 See also: [`stabilizerview`](@ref), [`destabilizerview`](@ref), [`logicalxview`](@ref), [`logicalzview`](@ref)
 """
 mutable struct MixedDestabilizer{T<:Tableau} <: AbstractStabilizer
@@ -711,7 +714,7 @@ mutable struct MixedDestabilizer{T<:Tableau} <: AbstractStabilizer
 end
 
 # Added a lot of type assertions to help Julia infer types
-function MixedDestabilizer(stab::Stabilizer{T}; undoperm=true, trackoperations=false) where {T}
+function MixedDestabilizer(stab::Stabilizer{T}; undoperm=true, undorowops=false) where {T}
     rows,n = size(stab)
     stab, r, s, permx, permz, xswaps, zswaps, xmul, zmul = canonicalize_gott!(copy(stab); recordops=true)
     
@@ -750,7 +753,7 @@ function MixedDestabilizer(stab::Stabilizer{T}; undoperm=true, trackoperations=f
     if undoperm
         t = t[:,invperm(permz)]::T
 
-        if trackoperations
+        if undorowops
             i = length(zmul)
             for zsw in reverse(zswaps)
 
@@ -773,7 +776,7 @@ function MixedDestabilizer(stab::Stabilizer{T}; undoperm=true, trackoperations=f
         
 
         t = t[:,invperm(permx)]::T
-        if trackoperations
+        if undorowops
             i = length(xmul)
             for xsw in reverse(xswaps)
                 while i > 0 && xmul[i][2] == xsw[2]
--- a/src/QuantumClifford.jl
+++ b/src/QuantumClifford.jl
@@ -747,15 +747,38 @@ function MixedDestabilizer(stab::Stabilizer{T}; undoperm=true) where {T}
         t[n+r+s+1:end] = sZ # The other logical set in the tableau
     end
     if undoperm
+        # print("s, r = ", s, r)
         for op in reverse(ops)
-            # print("Performing: ", op)
+            # print("\n\nPerforming: ", op, '\n')
             if op[1] == 1
+                # show(IOContext(stdout::IO, :limit => true), t)   
+                # print('\n')
+                # print("=>\n")
+
                 rowswap!(t, op[2], op[3]; phases=Val(true))
-                rowswap!(t, op[2] + s, op[3] + s; phases=Val(true))
+                rowswap!(t, op[2] + r + s, op[3] + r + s; phases=Val(true))
+                # show(IOContext(stdout::IO, :limit => true), t)   
+                # print('\n')
 
             elseif op[1] == 2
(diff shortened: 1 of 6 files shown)


adrianariton:
I noticed that most tests fail when also undoing the `mul_left` operation, maybe has something to do with the fact that it brings the lines to the original form, which `canonicalize!` does not (it makes them linearly independent). So I commented that part for now.

src/canonicalization.jl
```diff
@@ -218,42 +218,77 @@ See also: [`canonicalize!`](@ref), [`canonicalize_rref!`](@ref)
...
     permute!(stabilizer,zperm)
-    stabilizer, r, s, xperm, zperm
+    push!(ops, (4, 0, 0))
+    # print("\n-------------\n")
+    # print('\n', ops, '\n')
+    # 1-swap, 2-mul, 3-permx, 4-permz
+    # print(xperm, '\n', zperm, '\n')
+    # There must be a cleaner implementation by iteratively creating the
+    # inverse of the gaussian elimination matrix
+    stabilizer, r, s, xperm, zperm, ops
```
Krastanov, line 293: We can not afford to modify the return type of this function as this is a public function on which people depend.

src/canonicalization.jl
```diff
@@ -218,42 +218,77 @@ See also: [`canonicalize!`](@ref), [`canonicalize_rref!`](@ref)
 function canonicalize_gott!(stabilizer::Stabilizer; phases::Bool=true)
     @valbooldispatch _canonicalize_gott!(stabilizer; phases=Val(phases)) phases
 end
+
 function _canonicalize_gott!(stabilizer::Stabilizer; phases::Val{B}=Val(true)) where {B}
     xzs = tab(stabilizer).xzs
     rows, columns = size(stabilizer)
+    ops = []
+    retxperm = collect(1:rows)
+    retzperm = collect(1:rows)
```
Krastanov, line 227: retxperm and retzperm do not seem to be used for anything

src/QuantumClifford.jl
```diff
@@ -747,7 +747,45 @@ function MixedDestabilizer(stab::Stabilizer{T}; undoperm=true) where {T}
         t[n+r+s+1:end] = sZ # The other logical set in the tableau
     end
     if undoperm
-        t = t[:,invperm(permx[permz])]::T
+        # print("s, r = ", s, r)
+        for op in reverse(ops)
+            # print("\n\nPerforming: ", op, '\n')
+            if op[1] == 1
+                # show(IOContext(stdout::IO, :limit => true), t)
```
Krastanov, line 754: If something as complicated as this is needed for printing (as opposed to `@show` or `println`) there is an issue that should be reported in the bug tracker. Printing should be easy.

src/canonicalization.jl
Krastanov: Maybe an appropriate solution would be to add a new keyword argument `recordops::Val{Bool}=Val(false)` and condition the extra information on it.

src/canonicalization.jl
```diff
@@ -218,42 +218,77 @@ See also: [`canonicalize!`](@ref), [`canonicalize_rref!`](@ref)
...
         end
     end
+
     zperm, s = gott_standard_form_indices((@view xzs[end÷2+1:end,:]),rows,columns,skip=r)
     permute!(stabilizer,zperm)
-    stabilizer, r, s, xperm, zperm
+    push!(ops, (4, 0, 0))
+    # print("\n-------------\n")
+    # print('\n', ops, '\n')
+    # 1-swap, 2-mul, 3-permx, 4-permz
```
Krastanov, line 289: I am not a big fan of having this implicit integer encoding. Given that these operations are all in sequence, I think I would prefer if we use 4 separate arrays.

src/QuantumClifford.jl
adrianariton: I wanted to use this to be able to check bigger outputs with diff, to see where they differ!

src/canonicalization.jl
adrianariton: You're right, I will modify this immediately. Also from what I notice on tests, I don't think it is necesary to also track the multiply_left operations (even though the Stabiliser part of the MixedDestab looks like the original stab and that is nice, canonicalise! seems to want the lines linearly-independent).

adrianariton: I returned and undid all operations: row swap, aswell as mul_left. Should I be alarmed if the tests from test_stabs do not work, or do I also need to redo them given the new modified output for MixedDestabilizer?

adrianariton:
I added a new commit, in which I cleaned the code and also added the `trackoperations` option to MixedDestabilizer, defaulted on false, that tracks and undoes row operations: 

```Julia
# QuantumClifford.jl , line 714
function MixedDestabilizer(stab::Stabilizer{T}; undoperm=true, trackoperations=false) where {T}
...
```
This is then used in the conversion of Stab to CliffordOp:
```Julia
# dense_cliffords.jl, line 65
function CliffordOperator(stab::Stabilizer)
    md = MixedDestabilizer(copy(stab); trackoperations=true)
    return CliffordOperator(tab(md))
end
```

Krastanov:
A new file for the tests would be best. I am trying to follow the SciML style of tests where test files are small and selfcontained.

My bad on the email, I will respond shortly.

adrianariton:
Ok! I will start working on the test file later this weekend.

adrianariton:
Hello! I added a bunch of tests, checking the conversion went down correctly!
I do think there are some more tests I will be needing to add, could you help me with a range of operations I should check between the stab and it's conversion?

src/canonicalization.jl
```diff
@@ -215,45 +215,62 @@ Based on [gottesman1997stabilizer](@cite).
...
             end
             i += 1
         end
     end
+
     zperm, s = gott_standard_form_indices((@view xzs[end÷2+1:end,:]),rows,columns,skip=r)
     permute!(stabilizer,zperm)
-    stabilizer, r, s, xperm, zperm
+
+    if R
```
Krastanov, line 271: You are allocating a lot of expensive arrays even if `R` is false. It would be better if they do not get populated at all. Each `push!` can be prefaced by `R && push!(...)`. That way they get executed only if R is true.

also, there is a bunch new whitespace/newlines added. Could you clean it up?

test/test_stabtocliff.jl
```diff
@@ -0,0 +1,36 @@
...
+using QuantumClifford: stab_looks_good, destab_looks_good, mixed_stab_looks_good, mixed_destab_looks_good
+using Test
+
+
+@testset "Stabilizer to Clifford Conversion" begin
+    for i in 1:10
+        stab = random_stabilizer(4)
+        cl = CliffordOperator(stab)
+        cltab = tab(cl)
+        @test Stabilizer(cltab[end÷2+1:end]) == (stab);
```
Krastanov, line 12: could you clean up the extra parentheses and semicolons?

test/test_stabtocliff.jl
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+
+    for i in 1:10
+
+        gnd4 = S"ZIII
```
Krastanov, line 18: you can create this with `one(Stabilizer, ...)`

src/QuantumClifford.jl
```diff
@@ -711,9 +711,10 @@ mutable struct MixedDestabilizer{T<:Tableau} <: AbstractStabilizer
 end
 
 # Added a lot of type assertions to help Julia infer types
-function MixedDestabilizer(stab::Stabilizer{T}; undoperm=true) where {T}
+function MixedDestabilizer(stab::Stabilizer{T}; undoperm=true, trackoperations=false) where {T}
```
Krastanov, line 714: May we call it `undorowops` instead of `trackoperations`? Add an explanation of what it is to the docstring.

src/canonicalization.jl
```diff
@@ -215,45 +215,62 @@ Based on [gottesman1997stabilizer](@cite).
 
 See also: [`canonicalize!`](@ref), [`canonicalize_rref!`](@ref)
 """
-function canonicalize_gott!(stabilizer::Stabilizer; phases::Bool=true)
-    @valbooldispatch _canonicalize_gott!(stabilizer; phases=Val(phases)) phases
+function canonicalize_gott!(stabilizer::Stabilizer; phases::Bool=true, recordops::Bool=false)
```
Krastanov, line 218: the `recordops` needs to be explained in the docstring

test/test_stabtocliff.jl
```diff
@@ -0,0 +1,36 @@
...
+                 IZII
+                 IIZI
+                 IIIZ"
+
+        stab = random_stabilizer(4)
+        cl = CliffordOperator(stab)
+        cltab = tab(cl)
+        @test apply!(gnd4, cl) == (stab);
```
Krastanov, line 26: I am very surprised that this works without the need for canonicalization. Something might be wrong.

adrianariton: Shouldn't the Clifford operator bring the ground state to the desired one? Why do we need an extra canonicalize?

Krastanov:
A few conflicts exist now, so it would take some work to get this in a mergeable condition again. I will close it for now.

Thank you for tackling this, it is greatly appreciated! If you feel like taking on it again, feel free to ping me.

PR #127, opened 2023-06-10
initial example of PauliFrames with ECC

Krastanov:


Krastanov:
@ChenZhao44 - this is the example I had promised. There are still a few things to fix in the library for it to run well (both the PauliFrames and the compactification need some performance fixes)

ChenZhao44:
Thank you, Stefan. I am trying to run this example and conduct a benchmark.

Krastanov:
It seems that there are spurious allocations due to some type instability in Polyester thread local storage. I would suggest benchmarking only single threaded until this is fixed.

Krastanov:
@ChenZhao44 , the allocations from the threadlocal storage do not seem to be causing noticeable slowdown, so it should be fine to use it with multiple threads too. When you have some simple benchmarks, please do share them, I am eager to see a comparison and see what needs fixing.

PR #138, opened 2023-06-27
implement fault_matrix

Krastanov:


Krastanov:
@amicciche , this is what we had discussed in terms of decoding directly from a Pauli frame simulation. We probably need a few more helper methods, e.g. for creating the circuit for measuring the logical operations. Could you look through this and let me know if you have any questions about how to proceed with using it?

PR #144, opened 2023-07-06
naive encoding circuit to the ecc module

dmtrung14:


test/test_ecc.jl
```diff
@@ -1,6 +1,7 @@
 using Test
 using QuantumClifford
 using QuantumClifford.ECC: AbstractECC, Steane7, Shor9, Bitflip3, naive_syndrome_circuit, code_n, parity_checks, encoding_circuit, code_s, code_k, rate, distance,logx_ops, logz_ops
+include("../src/ecc/ECC.jl")
```
Krastanov, line 4: if this is necessary, something is quite broken with your project setup

src/ecc/ECC.jl
```diff
@@ -88,7 +88,122 @@ function logz_ops(c::AbstractECC)
...
+              return true
+        else
+              push!(dictionary, temp)
+        end
+    end
+    return false
+end
+
+""" Canonicalize the logicals x operators of a code by @gottesman1997 arxiv:quant-ph/9705052 """ # TODO: Implement the same code for the Z operators
+function canonicalize_logicals(c::AbstractECC)
```
Krastanov, line 122: this should probably act on `MixedDestabilizer`

src/ecc/ECC.jl
```diff
@@ -88,7 +88,122 @@ function logz_ops(c::AbstractECC)
     logicalzview(MixedDest)
 end
 
-# TODO implement isdegenerate
+""" The bimatrix representation of a code"""
+function tableau_representation(c::AbstractECC)
```
Krastanov, line 92: there is a `stab_to_gf2` function that already does this

src/ecc/ECC.jl
Krastanov: and it should canonicalize the tableau object, not convert it to a binary matrix (that would be more efficient)

Krastanov:
This seems like a very good start, and the fundamental math seem to be on the right track.

There are some software engineering things we need to clean up though.

Krastanov:
There are some git related issues too.

- you have been developing on your master branch instead of making a separate branch -- that can make keeping this work in sync with the main repo quite difficult and we should probably fix it
- there are merge conflicts between your changes and how the library has evolved since you have started making the changes. This is usually fixed with a merge commit or a rebase, but both of these are difficult if you have worked on top of the master branch.

PR #152, opened 2023-07-12
Support for GPU

Shayan-P:
added the QuantumCliffordGPU as an extension to QuantumClifford


src/gpu_adapters.jl
```diff
@@ -0,0 +1,11 @@
+GPU_DEPENDENCIES = ["CUDA"]
```
Krastanov, line 1: is this being used for anything?

test/test_gpu.jl
```diff
@@ -0,0 +1,118 @@
...
+
+function apply_single_or_double_qubit_and_compare(n, s, s_gpu)
+    op = if rand(Int) % 2 == 0
+        qbitidx = (((rand(Int) % n) + n) %n) + 1
+        random_clifford1(qbitidx)
+    else
+        # todo what other gates can we use? how to randomly generate two qubit operation?
+        # can we use random_clifford(2) with this function?
+        qbitidx1 = (((rand(Int) % n) + n) %n) + 1
+        qbitidx2 = (((rand(Int) % n) + n) %n) + 1
```
Krastanov, line 24: occasionally you end up with `qbitidx1==qubitidx2` which is not a valid gate, so some tests fail. This check was added after you started this branch

src/symbolic_cliffords.jl
```diff
@@ -37,6 +37,26 @@ end
 Base.@propagate_inbounds setxbit(s::TableauType{Tzv, Tme}, r::Int, c::Int, x::Tme, shift::Int) where {Tzv, Tme} = setxbit(s, r, c, x<<shift)
 Base.@propagate_inbounds setzbit(s::TableauType{Tzv, Tme}, r::Int, c::Int, z::Tme, shift::Int) where {Tzv, Tme} = setzbit(s, r, c, z<<shift)
 
+
```
Krastanov, line 40: the new getbit and setbit can probably be used inside of the older already existing Tableau methods, to avoid some repetition

Krastanov:
@Shayan-P , I added three comments on things to check for the next PR. I still merged this one, because I would like to make it easier for folks to test it out

src/symbolic_cliffords.jl
Shayan-P: I will do that in the new pr. but I'm not sure if it would affect performance due to the function call overhead

Krastanov:
Concerning the function call overhead, I would be really surprised if this does not get inlined. I would go as far as saying that it is a bug in the julia compiler heuristics if it does not get inlined.

PR #154, opened 2023-07-14
Naive Implementation of the Encoding Circuit

dmtrung14:
Compared to the code we saw today I broke the `naive_encoding_circuit` function into 3 parts: `rank` calculates the rank, `standard_tab_gott` calculates the standardized tableau in the format of PhysRevA.56.76, and the real naive_encoding_circuit. 

I put all of them in `ecc.jl`

I tested each function on the codes I knew and found no apparent mistakes. But they error out when I tried to swap `encoding_circuit` for `naive_encoding_circuit` in the `naive_syndrome_test`

Also I added noise to the `naive_syndrome_test`


src/ecc/papercode.jl
```diff
@@ -0,0 +1,9 @@
+struct Paper8 <: AbstractECC end
```
Krastanov, line 1: May we rename this to `Cleve8` (from the last name of the author). `Paper8` is a bit confusing :D 

I will test all of this later today and give comments

Krastanov:
I will make a few changes to this branch and push them here. I will comment on why the changes are made in the commit messages. Most of them are just stylistic questions. Please check through the messages so we are on the same page for why I was making the changes.

Do not forget to pull the now modified branch before you start making changes of your own.

src/ecc/ECC.jl
```diff
@@ -88,10 +88,105 @@ function logz_ops(c::AbstractECC)
...
+    for column in 1:2*n 
+        temp = tableau[:, column]
+        if temp in dictionary
+              return true
+        else
+              push!(dictionary, temp)
+        end
+    end
+    return false
+end
```
Krastanov, line 106: Could you document where is this from? Where did you find this algorithm?

src/ecc/ECC.jl
```diff
@@ -88,10 +88,105 @@ function logz_ops(c::AbstractECC)
...
+"""The standardized logical tableau of a code by [PhysRevA.56.76](@cite)"""
+function standard_tab_gott(c::AbstractECC)
+    n, s, k, r = code_n(c), code_s(c), code_k(c), rank(c)
+    # The standard form is 
+    # I A1 A2 | B C1 C2
+    # 0  0 0  | D  I  E  
+    # and we augment the following third line (for logical qubits)
+    # 0 E^T I | 0  0  0
+    # Then we apply the gates line by line bottom up in accordance with the formalisms here: arXiv:quant-ph/9607030
+    standard_tab = stab_to_gf2(stabilizerview(MixedDestabilizer(parity_checks(c), undoperm=false)))
```
Krastanov, line 130: at some point you have to undo the permutations of columns, otherwise your qubits are now reindexed from what the user was expecting, and great amount of confusion can ensue

src/ecc/ECC.jl
```diff
@@ -88,10 +88,105 @@ function logz_ops(c::AbstractECC)
...
+        for j in 1:n
+            if j > r && j <= n - k
+                augment[j] = standard_tab[j, 2*n-k+i] # the corresponding column of E in E^T 
+            elseif j == n-k+i 
+                augment[j] = 1
+            end
+        end
+        standard_tab = vcat(standard_tab, augment)
+    end
+    # Flipping the table so it has the same format as the papercode
```
Krastanov, line 143: Just use `transpose`, and if you need to reverse axes you can use `reverse`. It would be much shorter and more legible.

Krastanov:
Could you place any references in this file: https://github.com/QuantumSavory/QuantumClifford.jl/blob/master/docs/src/references.bib

You can then refer to them with a `[id](@cite)` string.

src/ecc/ECC.jl
Krastanov: also, check this function, it seems to be doing most of what you need. `canonicalize_gott` specifically seems to be doing all of the necessary steps, including creating the E^T part https://github.com/QuantumSavory/QuantumClifford.jl/blob/5ab1f0fe552f944c2afed7c6e00db8ac57b4a2a1/src/QuantumClifford.jl#L514

Krastanov: nope, I am wrong, it does not do everything you need, but might be a useful inspiration for how to write your code in a more simple way

Krastanov:
@dmtrung14 , I pushed two commits with small changes. Could you look at the TODOs I have left in them.

Krastanov:
Trung, have you tested the new encoding circuit function you have created? Just running it with default parameters is raising errors? New functionality should be tested - it helps both with review and with correctness.

dmtrung14:
> Trung, have you tested the new encoding circuit function you have created? Just running it with default parameters is raising errors? New functionality should be tested - it helps both with review and with correctness.

Yes, I did run the functions to test and nothing seems wrong. I did not write new tests for it though.

dmtrung14:
> > Trung, have you tested the new encoding circuit function you have created? Just running it with default parameters is raising errors? New functionality should be tested - it helps both with review and with correctness.
> 
> Yes, I did run the functions to test and nothing seems wrong. I did not write new tests for it though. I will do it now

Oh I think I figured out what is the problem: When I undoperm before I create the circuits, then the check `standard_tab[i, i] == 1` will return wrong when the columns are permuted. So we have the error. I fixed it.

Krastanov:
I submitted a commit that has a bit of cleanup and a few tests. Currently, with or without undoing permutations I seem to have the wrong results. E.g. with `encoding_circuit(Cleve8())` I get this:

image

Which seems pretty different from the one in the paper.

If you are making modifications, could you add them on top of the cleanup commit I submitted (a `git pull` should be enough on your end).

Let me know if you figure out what the problem is.

dmtrung14:
> I submitted a commit that has a bit of cleanup and a few tests. Currently, with or without undoing permutations I seem to have the wrong results. E.g. with `encoding_circuit(Cleve8())` I get this:
> 
> image
> 
> Which seems pretty different from the one in the paper.
> 
> If you are making modifications, could you add them on top of the cleanup commit I submitted (a `git pull` should be enough on your end).
> 
> Let me know if you figure out what the problem is.

Why do I keep getting this message when I tried to pull (and consequently my local repo is not updating)? Also, is there a built in method that allows me to create a circuit like that? It might help me to debug faster. Currently I am doing everything by hand.

PS C:\Users\progr\OneDrive\Quant SWE\Quantum Research\Spring 2023\julia_dev\QuantumClifford.jl> git pull upstream master
From https://github.com/QuantumSavory/QuantumClifford.jl
 * branch            master     -> FETCH_HEAD
Already up to date.

Krastanov:
the changes were made to your origin/minor branch (the branch which is being provided in this pull request), not the master branch

Krastanov:
the last two commits you made that I see here seem to match what I was doing, so maybe things are fine?

dmtrung14:
> the last two commits you made that I see here seem to match what I was doing, so maybe things are fine?

I think I figured out what is wrong with the circuit. I think we must have ignored what the `canonicalize_gott!` or `MixedDestabilizer` does on the second half of the bimatrix. 

It is currently returning this for `stabilizerview(c)`, 

```
 1  0  0  0  1  1  1  0 | 1  0  1  1  0  0  1  0
 0  1  0  0  1  1  0  1 | 1  1  0  1  0  1  0  0
 0  0  1  0  1  0  1  1 | 1  0  1  0  0  1  0  1
 0  0  0  1  0  1  1  1 | 1  1  0  0  0  0  1  1
 0  0  0  0  0  0  0  0 | 1  1  1  1  1  1  1  1
```

while it should return this: 

image


As you can see, the 3rd and the 4th rows have their bits flipped. I don't have any reasonable explanation for it.

Krastanov:
@amicciche , I think this should implement all we need. Thank you for the presentation last week, it made it much easier to implement. I also added the references you found to the documentation.

@dmtrung14 , thank you so much for the initial implementation during the summer! Anthony gave it another look and pointed out that there seems to be some issues in the paper. It turns out that both the thesis and the paper by Cleve contain some severe but subtle mistakes. This should address them all.

PR #171, opened 2023-08-26
Add noise gate

Shayan-P:
implements noise `applynoise!` function for `UnbiasedUncorrelatedNoise` on GPU


Krastanov:
ibig is an int because it is the typical "pointer" type, i.e. index into memory. `ismallm` is a bit mask (notice the m suffix) so it has to be of the same type as whatever it is masking.

ext/QuantumCliffordGPUExt/apply_noise.jl
```diff
@@ -0,0 +1,40 @@
...
+    ibig::Int, # todo why is this Int and others Tme?
+    ismallm::Tme, # why is rows Unsigned sometimes and sometimes Int?
+    rows::Unsigned) where {Tme <: Unsigned} 
+
+    f = (blockIdx().x - 1) * blockDim().x + threadIdx().x;
+    if f > rows
+        return nothing
+    end
+
+    r = rand() # todo add type to it? is it uniform?
```
Krastanov, line 30: Does cuda magically make rand be something that runs on the GPU?

Shayan-P: I'm not sure if the implementation of rand that is being used here is dispatched by CUDA.jl. However, I once checked the distribution of the random numbers that were generated, and that looked fine.
Let me ask this on Slack.

PR #172, opened 2023-08-26
Optimize inner gpu type

Shayan-P:
benchmark_gpu_innertype

Based on the plot above, changing the underlying type for GPU gives significant improvement.

This PR parametrized the inner type of `to_gpu` and `to_cpu` and sets the default type to `UInt32`


Shayan-P:
I removed the extra report files.

Issue #57, opened 2022-06-02
Inconsistency when calculating entanglement, possibly because of wrong graphs from stabilizers


royess:
I implemented two ways of calculating bipartite entanglement entropy base on this repo. One is by clipping algorithm. (Ref: Phys. Rev. B 100, 134306 (2019).) The other is by calculating the rank of bipartite adjacency matrix obtained by `Graphs.Graph(s::AbstractStabilizer)`. (Ref: arXiv:quant-ph/0602096.) And I also used Qiskit to cross check my results, using "statevector_simulator" backend.

In my 800 test cases, clipping algorithm agrees with Qiskit simulation. But there are 30 cases that entropy values from graphs differ by 1 or 2. So I think this may be a result of wrong graphs with more or less edges between two subsystems.

To help reproduce the issue:
- Codes to calculate entanglement entropy and run tests: https://github.com/royess/QuantumClifford.jl/tree/dev-clipping
- Codes to generate Qiskit data: https://github.com/royess/test-stab-entanglement/blob/main/get_data.py
- My test data in HDF5: [stab_ent.zip](https://github.com/Krastanov/QuantumClifford.jl/files/8822538/stab_ent.zip)

I know the issue involves many implementation details. I am willing to do more checks if needed.

royess:
I find a simple example:
```julia
julia> g = SimpleGraphFromIterator((Edge(1,2),Edge(1,3)))
{3, 2} undirected simple Int64 graph

julia> g
{3, 2} undirected simple Int64 graph

julia> Stabilizer(g)
+ XZZ
+ ZX_
+ Z_X
```

And in fact, the graph state should be:
$$
| g \rangle = \frac{1}{2\sqrt{2}} (| 000 \rangle + | 001 \rangle + | 010 \rangle + | 011 \rangle +| 100 \rangle - | 101 \rangle - | 110 \rangle + | 111 \rangle ).
$$

And it is not stabilized by `+ ZX_` or `+ Z_X`.

Krastanov:
Thanks for sharing this fork, the work there seems very interesting! Could you elaborate a bit on how one can reproduce the bug? It seems you are suggesting it is based on something related to the generation of a graph out of a stabilizer. That is implemented here https://github.com/Krastanov/QuantumClifford.jl/blob/master/src/graphs.jl#L4= and it is indeed very poorly tested and as of yet quite unused, so bugs might be lurking there.

royess:
[...]

PR #184, opened 2023-09-04
initial implementation of naive_encoding_circuit

amicciche:
Everything seems to be working for the encoding circuits I'm generating except for the logical X pauli frame measurement. My hunch is that it has to do with preparing the initial state in the X basis, but my reasoning leads me to think that there should be Hadamard gates  on the first k qubits, where k is the number of qubits to be encoded. The current implementation of the logical X measurement applies a Hadamard to gate 1, which for Shor9 and Steane7 is correct (for them, k=1). Is my understanding correct? Do you have some other idea what might be the problem?

Anyway here are the good results: 
Logical vs Physical error plots, generated using Cleve97 style encoding circuits, as well as the reindexing stabilizer tableau:

image
image

Encoding circuit generated for Steane7:
image

Encoding circuit generated for Shor9:
image

Encoding circuit generated for Cleve8 (the circuit in the paper)
image

My $X*$ and $Z*$ ended up a little different from the paper, but I believe that it was simply due to some differences in the exact Gaussian elimination operations performed. Here are my matrices:
```
julia> Xstar
8×8 Matrix{Bool}:
 1  0  0  0  1  1  0  1
 0  1  0  0  1  1  1  0
 0  0  1  0  1  0  1  1
 1  1  1  0  0  1  1  1
 0  0  0  0  1  0  0  0
 0  0  0  0  0  1  0  0
 0  0  0  0  0  0  1  0
 0  0  0  0  0  0  0  1

julia> Zstar
8×8 Matrix{Bool}:
 0  0  0  1  1  1  1  1
 0  0  0  1  1  0  1  0
 0  0  0  1  0  1  0  1
 0  0  0  1  1  1  0  0
 0  0  0  1  0  0  0  0
 0  0  0  1  1  0  0  1
 0  0  0  1  0  0  1  1
 0  0  0  1  0  1  1  0
```

Tomorrow I will work on cleaning things up, doing more testing, hopefully figuring out the logical X measurement, and also implementing code for $B'$ part of the paper. Steane7, Shor9, and Cleve8 all have $r_2=0$, so $X^{(1)}==X^{(2)}$ and $Z^{(1)}==Z^{(2)}$. For that reason, I left that as a TODO, since I need to find or come up with a code which has a nonzero value for $r_2$ so that I can properly test and troubleshoot.

### DIFF
--- a/src/ecc/ECC.jl
+++ b/src/ecc/ECC.jl
@@ -2,6 +2,7 @@ module ECC
 
 using QuantumClifford
 using QuantumClifford: AbstractOperation
+using LinearAlgebra
 import QuantumClifford: Stabilizer, MixedDestabilizer
 
 abstract type AbstractECC end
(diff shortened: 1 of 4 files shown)


amicciche:
Speaking of there not being a built in code that has the property needed to fully test the canonicalization in the 1997 Cleve and Gottesman paper, should we open an issue to combine all the src/ecc/[insert_name]code.jl files into one big src/ecc/codezoo.jl file?
And then add a considerable number of additional codes?

src/ecc/ECC.jl
```diff
@@ -398,4 +399,146 @@ include("./bitflipcode.jl")
...
+                             XIYZXIYZ
+                             XZIYIYXZ"
+                             
+function canonicalize_cleve97(checks::Stabilizer)
+    d, n = size(checks)
+    X0 = (checks |> stab_to_gf2)[:,1:n]'
+    Z0 = (checks |> stab_to_gf2)[:,n+1:2n]'
+
+    # Now let's work on getting X1 and Z1
+    b = rank(X0)
```
Krastanov, line 419: This is probably not the rank you want. This is rank over the real field, not over the boolean field. E.g. this matrix

```
1 0 1
1 1 0
0 1 1
```

has rank 3 over the reals but rank 2 over boolean numbers (the first row is the sum of the second and third row).

See here for a workaround https://github.com/QuantumSavory/QuantumClifford.jl/issues/57#issuecomment-1146270352

Krastanov:
I think I would prefer to keep the various codes in separate files that get `include`d, just for organization purposes. But yes, adding more codes would be quite valuable and I am trying to get a few undergrads to help with that.

Currently I have a couple of vague worries about the approach taken (but I can be convinced otherwise):
- the tableaux are converted to much more expensive boolean array form and then back, which is a bit wasteful -- other canonicalization schemes just work directly on the tableaux
- the canonicalize_gott scheme is extremely close to the one used here (together with the extra processing done in the constructor of `MixedDestablizer`) -- it would be quite valuable to figure out how much can be reused.

I will try to provide a more detailed review later in the week.

A few minor things:
- use `error("error message")` instead of `println("ERROR")`
- use `@debug` instead of `println("debug message")` and set your log level to debug

src/ecc/ECC.jl
amicciche: Whoops, you're almost definitely right! Good catch!

amicciche:
Just realized that I made a mistake - so of course the logical X measurement isn't working - the encoding circuit is correct for Cleve8, but wrong for Steane7 and likely Shor9 as well. Cleve8 had $X^{1}=X^{2}$ because $r_1=1$, NOT because $r_0=0$, although the former implies the latter. This is not true for Steane7 and Shor9, so I need to implement the B' part for them to work. 

I'll try to get a commit with the proper B' submitted sometime today, and at that time I'll let you know if logical X measurement still isn't working.

amicciche:
My recent commit fixed the problem with the logical X error in pauli frame simulation!

Now I need to just clean it up (following the above comments) and maybe also find a way to test the generated encoding circuits on codes that encode more than one qubit. 

Here are some results:
**Steane7**: generated circuit and performance of that circuit using the logical pauli frame decoder evaluator. 
image
image

**Shor9**: generated circuit and performance of that circuit using the logical pauli frame decoder evaluator. 
image
image

**Cleve8**: generated circuit. My current setup can't evaluate this code - I believe because $k>1$. 
image
Here are the $X*$, $Z*$ matrices for Cleve8
```
julia> Xstar
8×8 Matrix{Bool}:
 1  0  0  0  1  1  1  0
 0  1  0  0  1  0  1  1
 0  0  1  0  0  1  1  1
 1  1  1  0  1  1  0  1
 0  0  0  0  1  0  0  0
 0  0  0  0  0  1  0  0
 0  0  0  0  0  0  1  0
 0  0  0  0  0  0  0  1

julia> Zstar
8×8 Matrix{Bool}:
 0  0  0  1  1  0  1  0
 0  0  0  1  0  1  0  1
 0  0  0  1  1  1  0  0
 0  0  0  1  1  1  1  1
 0  0  0  1  0  0  0  0
 0  0  0  1  1  0  0  1
 0  0  0  1  0  0  1  1
 0  0  0  1  0  1  1  0
```

Krastanov:
finished in #154

PR #195, opened 2023-10-31
Error correction methods for creating Bicycle and Unicycle codes + belief decoder code evaluation

Benzillaist:
Added multiple methods:

code_generation:
- Methods for creating Bicycle and Unicycle codes
- Methods for assembling CSS codes
- Utility methods for working with codes

code_evaluation:
- Methods for evaluating codes with a belief propagation decoder
- Methods for plotting codes given error rates

ECC.jl was modified to include the required imports for the two added files.


src/ecc/ECC.jl
```diff
@@ -1,18 +1,24 @@
...
+using Statistics: std
+using LinearAlgebra: rank
 
 abstract type AbstractECC end
 
 export Shor9, Steane7, Cleve8, Perfect5, Bitflip3,
     parity_checks, naive_syndrome_circuit, shor_syndrome_circuit, naive_encoding_circuit,
     code_n, code_s, code_k, rate, distance,
-    isdegenerate, faults_matrix
+    isdegenerate, faults_matrix, CSS_Code, Bicycle_Code, Unicycle_Code, Circ2BicycleH0,
```
Krastanov, line 17: some naming conventions: `CSS_Code` should be `CSSCode`, same for Bicycle and Unicycle; use `To` not `2`

src/ecc/ECC.jl
```diff
@@ -1,18 +1,24 @@
...
 abstract type AbstractECC end
 
 export Shor9, Steane7, Cleve8, Perfect5, Bitflip3,
     parity_checks, naive_syndrome_circuit, shor_syndrome_circuit, naive_encoding_circuit,
     code_n, code_s, code_k, rate, distance,
-    isdegenerate, faults_matrix
+    isdegenerate, faults_matrix, CSS_Code, Bicycle_Code, Unicycle_Code, Circ2BicycleH0, 
+    Circ2UnicycleH0, AssembleCSS, BicycleSetGen, BicycleSetGenRand, GetCodeTableau, 
+    GetXTableau, GetZTableau, parity_checks, ReduceBicycle, ReduceUnicycle, 
+    create_lookup_table, evaluate_code_decoder_w_ecirc_pf, plot_code_performance, pf_encoding_plot
```
Krastanov, line 20: Exporting all these new things implies they are part of the public stable API, and we can not modify them without making a breaking release; I would prefer to be much more conservative in what we export. Which of these are things that the end user definitely needs? The "maybe needs" and "does not need" should not be exported for now.

src/ecc/code_generation.jl
```diff
@@ -0,0 +1,298 @@
...
+# using Nemo: residue_ring, matrix
+# using LinearAlgebra: rank
+
+"""Struct for arbitrary CSS error correcting codes.
+
+This struct holds:
+    - tab: Boolean matrix with the X part taking up the left side and the Z part taking up the right side
+    - stab: Stabilizer of the parity check matrix
+    - n: Block length
+    - d: Code distance"""
```
Krastanov, line 14: there does not seem to be a `d` parameter in the struct. Also distance is usually unknown for most codes

src/ecc/code_generation.jl
```diff
@@ -0,0 +1,298 @@
...
+# using Statistics:std
+# using Nemo: residue_ring, matrix
+# using LinearAlgebra: rank
+
+"""Struct for arbitrary CSS error correcting codes.
+
+This struct holds:
+    - tab: Boolean matrix with the X part taking up the left side and the Z part taking up the right side
+    - stab: Stabilizer of the parity check matrix
+    - n: Block length
```
Krastanov, line 13: n can be deduced from tab or stab, so no need to keep it here. There is a `code_n` you can define a method for

src/ecc/code_generation.jl
```diff
@@ -0,0 +1,298 @@
...
+# import QuantumClifford.ECC: parity_checks
+# using Statistics:std
+# using Nemo: residue_ring, matrix
+# using LinearAlgebra: rank
+
+"""Struct for arbitrary CSS error correcting codes.
+
+This struct holds:
+    - tab: Boolean matrix with the X part taking up the left side and the Z part taking up the right side
+    - stab: Stabilizer of the parity check matrix
```
Krastanov, line 12: stab can be deduced from tab, so let's just keep tab for now, no need to duplicate it

src/ecc/code_generation.jl
```diff
@@ -0,0 +1,298 @@
...
+    n::Int
+end
+
+"""Takes an untrimmed bicycle matrix and removes the row which keeps the spread of the column weights minimal.
+
+Required before the bicycle code can be used.
+
+Typical usage:
+ReduceBicycle(Circ2BicycleH0(array_indices, (block length / 2) ) )"""
+function ReduceBicycle(H0::Matrix{Bool})
```
Krastanov, line 27: naming conventions: keep functions in `snake_case`, not `CamelCase` (unless they are constructors)

src/ecc/code_generation.jl
```diff
@@ -0,0 +1,298 @@
...
+    end
+    comp_matrix[1:n,1:n] = circ_matrix
+    comp_matrix[1:n,n+1] = one_col
+    return comp_matrix
+end
+
+function AssembleCSS end
+
+"""Creates a CSS code using the two provided matrices where H contains the X checks and G contains the Z checks."""
+function AssembleCSS(H::Matrix{Bool}, G::Matrix{Bool})::CSS
```
Krastanov, line 125: This should be just a CCS constructor, no need to introduce a new function. I.e. rename `AssembleCSS` to `CSS`

src/ecc/code_generation.jl
```diff
@@ -0,0 +1,298 @@
...
+            end
+        end
+    end
+    return circ_arr
+end
+
+"""Takes in a boolean Matrix and returns the parity check tableau as a string of characters.
+
+Note: Only works when the block length for the X and Z checks are the same!"""
+function GetCodeTableau(ecc::Matrix{Bool})
```
Krastanov, line 242: do not create a string just to parse it afterwards, it is much simpler to directly use a `Tableau` or `Stabilizer` constructor (there is one that takes two boolean matrices)

src/ecc/code_generation.jl
```diff
@@ -0,0 +1,298 @@
...
+        end
+        ps = string(ps,"\n")
+    end
+    return ps
+end
+
+"""Takes in a matrix and returns just the X checks portion while keeping the full height of the matrix.
+
+Note: Only works when the block length for the X and Z checks are the same!"""
+function GetXTableau(ecc::Matrix{Bool})
```
Krastanov, line 266: You can just do `ecc[:, 1:end÷2]` which would also have the benefit of being sufficiently easy to write/read that you do not need a dedicated function for it.

src/ecc/code_generation.jl
```diff
@@ -0,0 +1,298 @@
...
+
+"""Takes in a matrix and returns just the Z checks portion while keeping the full height of the matrix.
+
+Note: Only works when the block length for the X and Z checks are the same!"""
+function GetZTableau(ecc::CSS)
+   return GetZTableau(ecc.tab)
+end
+
+"""Returns the matrix form of the X and Z checks."""
+tableau(c::CSS) = c.tab
```
Krastanov, line 292: this word (`tableau`) usually refers to a matrix made of Pauli operators, not a matrix made of booleans

src/ecc/ECC.jl
Krastanov: also, I think you are exporting a few names that are not actually defined here (e.g. `pf_encoding_plot`)

Krastanov:
This is pretty exciting! I left a bunch of comments in, but they are mostly on coding conventions, not on the substance of the PR.

I would suggest putting the CCS things in a separate file called `css.jl` and then the bicycle and unicycle codes in a file `uni_bi_cycle_codes.jl` or something like that. The name `code_generation.jl` is much too general, while the content is just for a few specific codes.

We should think of adding some tests, but we can discuss this later.

src/ecc/css.jl
```diff
@@ -0,0 +1,31 @@
+"""Struct for arbitrary CSS error correcting codes.
+
+This struct holds:
+    - tab: Boolean matrix with the X part taking up the left side and the Z part taking up the right side"""
```
Krastanov, line 4: The inside components of a struct should be considered private (and thus not documented in user-facing documentation). Only get/set-like interfaces should be considered public because (a) they are more universal, (b) internal structure can now change without breaking user code.

src/ecc/css.jl
```diff
@@ -0,0 +1,31 @@
...
+struct CSS <: ECC
+    tab::Matrix{Bool}
+end
+
+function CSS end
+
+"""Creates a CSS code using the two provided matrices where H contains the X checks and G contains the Z checks."""
+function CSS(H::Matrix{Bool}, G::Matrix{Bool})::CSS
```
Krastanov, line 12: avoid adding too much type constraints in methods. Maybe someone would define an object that has the `Matrix{Bool}` interface but is of different type. That object could in principle still work here. Type signatures in methods are there to specify dispatch rules, nothing more -- the compiler can then infer types that this works with and compile a fast version for each type.

src/ecc/css.jl
```diff
@@ -0,0 +1,31 @@
+"""Struct for arbitrary CSS error correcting codes.
+
+This struct holds:
+    - tab: Boolean matrix with the X part taking up the left side and the Z part taking up the right side"""
+struct CSS <: ECC
+    tab::Matrix{Bool}
+end
+
+function CSS end
```
Krastanov, line 9: not necessary. `CSS` already exists as a default constructor, and you can just define extra methods for it (as you do below)

src/ecc/css.jl
```diff
@@ -0,0 +1,31 @@
...
+    # comp_matrix = Matrix{Bool}(undef, Hy + Gy, Hx + Gx)
+    comp_matrix[1:Hy, 1:Hx] = H
+    comp_matrix[Hy+1:end, Hx+1:end] = G
+    pcm_stab = Stabilizer(fill(0x0, Hy+Gy), GetXTableau(comp_matrix), GetZTableau(comp_matrix))
+    return CSS(comp_matrix)
+    # return comp_matrix
+end
+
+"""Returns the matrix form of the X and Z checks."""
+get_xzs(c::CSS) = c.tab
```
Krastanov, line 25: it does not seem this is used anywhere, so I am deleting it

src/ecc/css.jl
```diff
@@ -0,0 +1,31 @@
...
+
+"""Creates a CSS code using the two provided matrices where H contains the X checks and G contains the Z checks."""
+function CSS(H::Matrix{Bool}, G::Matrix{Bool})::CSS
+    Hy, Hx = size(H)
+    Gy, Gx = size(G)
+    comp_matrix = fill(false, (Hy + Gy, Hx + Gx))
+    # comp_matrix = Matrix{Bool}(undef, Hy + Gy, Hx + Gx)
+    comp_matrix[1:Hy, 1:Hx] = H
+    comp_matrix[Hy+1:end, Hx+1:end] = G
+    pcm_stab = Stabilizer(fill(0x0, Hy+Gy), GetXTableau(comp_matrix), GetZTableau(comp_matrix))
```
Krastanov, line 19: it does not seem getxtableau and so on are defined

Krastanov:
I did some minor cleanup along the lines of the comments I just posted. Make sure you pull locally before you continue modifying.

Also, it is helpful to always create a separate branch when making modifications. Currently all of this seems to be on your master branch which makes it difficult for you to experiment with multiple things at the same time.

I think there are still some things not functional in simple_sparse_codes.jl

When you address a comment, please also click "resolve" on it, as it makes it easier for me to review.

I will mark this as a draft. When you are ready for the next round of reviews, just mark it back to not-a-draft.

Krastanov:
Ben, I just now noticed that you had actually marked this as ready for review for quite a while. I apologize for my oversight! Could you click "resolve conversation" on the comments you have addressed in the follow-up edits? That would make it much easier for me to finish the review.

Benzillaist:
Hi Stefan, I apologize for not informing you that I pushed an update. I believe that my most recent commits should resolve any problems. I also tested my code to ensure it works, but please let me know if I've made any glaring errors!

src/ecc/css.jl
```diff
@@ -0,0 +1,22 @@
...
+    Gy, Gx = size(G)
+    comp_matrix = falses(Hy + Gy, Hx + Gx)
+    comp_matrix[1:Hy, 1:Hx] = H
+    comp_matrix[Hy+1:end, Hx+1:end] = G
+    return CSS(comp_matrix)
+end
+
+"""Returns the stabilizer making up the parity check tableau."""
+function parity_checks(c::CSS)
+    Stabilizer(fill(0x0, size(c.tab, 2)), c.tab[:,1:end÷2], c.tab[:,end÷2+1:end])
```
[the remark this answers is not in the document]
Krastanov: This will probably just take the code that is currently in your constructor above

src/ecc/simple_sparse_codes.jl
```diff
@@ -0,0 +1,231 @@
...
+            end
+        end
+    end
+    return m
+end
+
+"""Takes a list of indices and creates the base of the unicycle matrix.
+
+For example:
+Circ2UnicycleH0([1, 2, 4], 7)
```
Krastanov, line 110:
```suggestion
`Circ2UnicycleH0([1, 2, 4], 7)`
```

src/ecc/ECC.jl
```diff
@@ -1,18 +1,23 @@
...
 using LinearAlgebra
-using QuantumClifford
-using QuantumClifford: AbstractOperation, AbstractStabilizer
+using QuantumClifford, SparseArrays, LDPCDecoders
+using QuantumClifford: AbstractOperation, AbstractStabilizer, Stabilizer
 import QuantumClifford: Stabilizer, MixedDestabilizer
 using DocStringExtensions
 using Combinatorics: combinations
+using Statistics: std
+using LinearAlgebra: rank
```
Krastanov, line 10:
```suggestion
```

Already imported above

src/ecc/ECC.jl
```diff
@@ -289,6 +294,8 @@ function isdegenerate(H::Stabilizer, d::Int=1)
 end
 
 include("circuits.jl")
+include("css.jl")
+include("simple_sparse_codes.jl")
```
Krastanov, line 298: move these two to the `codes` folder

src/ecc/simple_sparse_codes.jl
```diff
@@ -0,0 +1,231 @@
...
+    comp_matrix[1:n,n+1:2*n] = transpose(circ_matrix)
+    return comp_matrix
+end
+
+"""Takes an untrimmed unicycle matrix and removes linearly dependent rows.
+
+Required before the unicycle code can be used.
+
+Typical usage:
+ReduceUnicycle(Circ2UnicycleH0(array_indices, block length) )"""
```
Krastanov, line 88:
```suggestion
`ReduceUnicycle(Circ2UnicycleH0(array_indices, block length) )`"""
```

Krastanov (changes requested):
I added a few more comments. This will also need tests before we can merge it. In particular, tests for whether the generated codes are valid.

One possible type of tests is to just check exactly for the parity check matrix of known small codes.

Another type of tests to add is tests for random large codes verifying they produce valid tableaux.

Krastanov:
I also noticed that you are working on your master branch. That makes work generally more difficult, as it is cumbersome to keep things in sync with the main repository.

Krastanov:
@amicciche , skim through the CSS definition that Ben is making here. Just making sure you guys are not duplicating work that is done by the other.

amicciche:
> @amicciche , skim through the CSS definition that Ben is making here. Just making sure you guys are not duplicating work that is done by the other.

My definition of a CSS struct was the bare minimum to get the idea of the functionality implemented quickly. In other words, it was sort of a functional placeholder. Here it is in its entirety:
```julia
struct CSS; tableau; cx; cz; end
```
Ben's version seems to be a bit more detailed and fleshed out, and could easily just replace my version. I do have some code that was not included in my pull request that coverts the matrices into a stabilizer tableau, and perhaps could be part of the CSS struct?

src/ecc/css.jl
```diff
@@ -0,0 +1,22 @@
+"""An arbitrary CSS error correcting code defined by its X and Z checks."""
+struct CSS <: AbstractECC
+    tab
```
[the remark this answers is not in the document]
Benzillaist: Will do, I made some slight modifications to what you wrote which is why I didn't immediately commit your changes.

src/ecc/css.jl
```diff
@@ -0,0 +1,22 @@
...
+    return CSS(comp_matrix)
+end
+
+"""Returns the stabilizer making up the parity check tableau."""
+function parity_checks(c::CSS)
+    Stabilizer(fill(0x0, size(c.tab, 2)), c.tab[:,1:end÷2], c.tab[:,end÷2+1:end])
+end
+
+"""Returns the block length of the code."""
+code_n(c::CSS) = size(c.tab,2)÷2
```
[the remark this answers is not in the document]
Benzillaist: Thank you, fixed

src/ecc/css.jl
```diff
@@ -0,0 +1,22 @@
+"""An arbitrary CSS error correcting code defined by its X and Z checks."""
+struct CSS <: AbstractECC
+    tab
+end
+
+"""Creates a CSS code using the two provided matrices where H contains the X checks and G contains the Z checks."""
+function CSS(H, G)
+    Hy, Hx = size(H)
```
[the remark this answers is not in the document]
Benzillaist: I updated this to work with the updated struct

src/ecc/css.jl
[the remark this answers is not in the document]
Benzillaist: Modified to work with updated struct

src/ecc/ECC.jl
Benzillaist: They have been moved

Benzillaist:
Thank you for the review! I have implemented all the changes you requested. 
I can definitely prepare some codes, but I'm unsure how to prove that they are valid. Do you know where I might be able to come across some existing bicycle codes for comparison?

Additionally, I will be using new branches for all future commits. Thank you for that recommendation as well!

Krastanov:
After #212, a lot of the functionality here has already been implemented. I will close this one for now, but please consider opening a new one that adds the new codes you have defined -- they are not in #212 and will be greatly appreciated.

PR #238, opened 2024-02-28
Zoo: [[2^j, 2^-j-2,3]] Gottesman /Quantum Hamming Code

Fe-r-oz:

PR #240, opened 2024-03-17
Gottesman's code

Krastanov:
A couple of cleanups on top of #238 by @Fe-r-oz


Fe-r-oz:
Kindly please see the output of j==4 which is  

```
parity_checks(Gottesman(4))
+ XXXXXXXXXXXXXXXX
+ ZZZZZZZZZZZZZZZZ
+ X_X_X_X_YZYZYZYZ
+ X_X_YZYZ_X_XZYZY
+ X_YZ_XZYX_YZ_XZY
+ XZ_YXZ_YXZ_YXZ_Y
```

This does not match with the Stabilizer of [16 10 3 ] that is presented in his Gottesman thesis, [Table 8.1 Page 91](https://thesis.library.caltech.edu/2900/2/THESIS.pdf). More details can be found in Pages 90 to 95.

Here is what the [[16,10, 3]] Stabilizer is like:

```
+ XXXXXXXXXXXXXXXX
+ ZZZZZZZZZZZZZZZZ
+ _X_X_X_XZYZYZYZY
+ _X_XZYZYX_X_YZYZ
+ _XZYX_YZ_XZYX_YZ
+ _YXZ_YXZ_YXZ_YXZ
```

Krastanov:
Aren't they the same up to a row operation?

image

Krastanov:
There is really only one fundamental feature in these codes: Every single-qubit error needs to be detectable and it has to provide a unique syndrome (that is where the name "quantum Hamming code" comes from). The newly added test file explicitly checks.

Of course I might be missing something, but to convince me you need to at least explain what you mean by "there are fundamental changes in the main generators M1 to MJ".

Fe-r-oz:
> Aren't they the same up to a row operation?

Only first two rows are the same.

The third row to the last row are different which are a large number of rows.

Krastanov:
Check the screenshot I put above. The two stabilizer tableaux you posted correspond to the same state (the same code) because they can be obtained from each other by performing row operations. Take the first row and multiply it in 3rd-to-6th rows and you get the other stabilizer tableau. Just like the matrix representing a system of linear equations does not change its solution after performing row operations.

Certain, when talking about codes it is worthwhile to also consider other features, like row-weights, but those also do not change between the two representations.

Fe-r-oz:
> Of course I might be missing something, but to convince me you need to at least explain what you mean by "there are fundamental changes in the main generators M1 to MJ".

But there is a fundamental design error according to my understanding of his thesis, M1..Mj , first two digits can't be X. 

The stabilizer is very different from what he presented. You can think of it if you write it as a full blown matrix. This is clearly violating the symmetric structure!

Please see his comments below:

Screenshot_select-area_20240318020645

M1 to Mj starts with I...not X. That's the characteristics of this family  Please see pages 90 to 95

Krastanov:
Feroz, I already answered in multiple ways:

- the tableaux are equivalent up to a pretty simple row operation
- the fundamental property of distinguishing all single-qubit errors is present

Here is a third version of the answer:

The difference between the two approaches is simply using `not x` instead of `x` when enumerating stabilizer.

> From my understanding of his thesis, M1 to Mj starts with I...not X. That's the characteristics of this family. There are many other changes in the stabilizer as well Please see pages 90 to 95

Why should they? Presumably the answer is that this makes some enumeration a bit easier. Sure, that is a valid reason to pick a given set of generators. But I have shown you above how to prove that this set of generators is the same.

> It would be better to produce the stabilizer that is in his thesis.

This type of statements need to be defended. Why is it better? Especially given that the thesis and the paper are basically the same algorithm with slightly different conventions (and it seems the thesis does not describe it in particularly explicit detail).

Fe-r-oz:
> This type of statements need to be defended. Why is it better? Especially given that the thesis and the paper are basically the same algorithm with slightly different conventions (and it seems the thesis does not describe it in particularly explicit detail).

It's the opposite, the thesis describe many checks and is explicit for straight 3 pages from 92 to 95 and the paper does not. Why did he then explained those checks if there was no need to? He could have described the same 3 paragraph algorithm in his paper, but he explained in with many checks for even j and odd j for both [[8,3,3]] and [[16,10,3]] cases?

Fe-r-oz:
We can write the full blown H with 6 rows and 64 columns matrix from this stabilizer code just like the stabilizer Matrix.

Similarly, we can write the matrix for this stabilizer, and study whether its correct or not.

Screenshot_select-area_20240318033419

We can compare the structures. We can also see if there are extended hamming matrices present or not. Also, there is a characteristic structure. It will take some time, but I can sent the attachment here.

Krastanov:
> There is more detail then just matter of change of conventions

Feroz, answer with yes or no, do you understand that two parity check matrices are equivalent if they can be converted from one to the other just by row operations?

Second question, yes or no, are the matrices from the thesis and from the paper equivalent?

Krastanov:
I am preparing a small public website that will contain some visualizations related to codes like this. I would like to add the work you have contributed here. May I list you as one of the contributors? Currently I just have your name as Feroz, but most folks will be listed as FirstName LastName, position, institution, e.g. "Joe Smith, undergraduate, UMass Amherst". Let me know how to fill out that information about you if you are comfortable being listed.

Krastanov:
Tests pass, including the new "fundamental feature" test. I am merging this and closing the other related pull request.

Thank you for contributing this! If you did not start this process, this would never have been added! And I learnt quite a bit from the bibliography you shared.

Krastanov:
here are the performance plots:

image

Here they are up to j=6, but in the test suite there is a test that pushes it to j=12 by explicitly enumerating the errors

Fe-r-oz:
> Feroz, answer with yes or no, do you understand that two parity check matrices are equivalent if they can be converted from one to the other just by row operations?

> Second question, yes or no, are the matrices from the thesis and from the paper equivalent?

I needed time to be sure about second so I verified the Row Operations of last three rows with XXXXXXXXXXXXXXXX. @Krastanov 

1) yes 2) yes 
Document 17_1

Fe-r-oz:
> I am preparing a small public website that will contain some visualizations related to codes like this. I would like to add the work you have contributed here. May I list you as one of the contributors?

Alright 👍🏻

Feroz Ahmad, graduate, NUCES.

> Thank you for contributing this! If you did not start this process, this would never have been added! And I learnt quite a bit from the bibliography you shared.

Thank you as well. 👍🏻

PR #212, opened 2024-01-19
provide interfaces to numerous ECC decoders as well as ECC evaluation pipelines

Krastanov:
This requires PyQDecoders to be published first: https://github.com/JuliaRegistries/General/pull/99144

Until then it can be tested with executing `dev https://github.com/QuantumSavory/PyQDecoders.jl` before dev-ing this branch.

PR #241, opened 2024-03-18
 adding iscss for issue #212

Fe-r-oz:
Given the stabilizer has the knowledge, the `isscss` function is implemented.


Krastanov:
And this probably needs some tests. E.g. for each code check what `iscss` says and then check whether their parity check matrix is indeed in CSS form. The check would be expensive, it will have to go through the entire tableau of a specific code instance, but it will be happening only when running the test suite.

Fe-r-oz:
> And this probably needs some tests. E.g. for each code check what `iscss` says and then check whether their parity check matrix is indeed in CSS form. The check would be expensive, it will have to go through the entire tableau of a specific code instance, but it will be happening only when running the test suite.

I thought this was not needed because the stabilizer has the knowledge as you said.  Only, tests are to be remaining.

Please have a look at modifications.

so, now I need to use an algorithm that check whether H matrix is in CSS form? Can the stabilizer knowledge can be used here or it's an expensive method, I need to look for a general algorithm in literature?

Krastanov:
> I thought this was not needed because the stabilizer has the knowledge as you said. Only, tests are to be remaining.

The point of tests is to check whether such assumptions or internal constants are correct. It is not something that runs when the user is using the library, rather it is for the developers to check whether they have made an error (e.g. like setting `iscss(::SomeCode) = true` when it should not be).

Krastanov:
> I need to look for a general algorithm in literature?

It is a simple check, not something that you need to look up in the literature. For each row of a parity check tableau check whether it contains a mix of X and Z. As long as they do not, it is a CSS code.

src/ecc/ECC.jl
```diff
@@ -49,6 +49,11 @@ function parity_checks_z(code::AbstractECC)
     throw(lazy"Codes of type $(typeof(code)) do not have separate X and Z parity checks, either because they are not a CSS code and thus inherently do not have separate checks, or because its separate checks are not yet implemented in this library.")
 end
 
+# default CSS method
+function iscss(::Type{T}) where T<:AbstractECC
```
Krastanov, line 53: it would make sense to have a check for an instance of a type, not just a check for the type. Something like `iscss(c::AbstractECC) = iscss(typeof(c))`

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@
 # News
 
 ## v0.8.21 - 2024-03-17
-
+- adding iscss function
```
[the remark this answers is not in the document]
Krastanov: this is not resolved, do not mark it as resolved until done

Fe-r-oz:
> It is a simple check, not something that you need to look up in the literature. For each row of a parity check tableau check whether it contains a mix of X and Z. As long as they do not, it is a CSS code.

Prepared according to desired functionality.

Fe-r-oz:
I added tests for Gottesman(3) and some other CSS codes. The results are correct in the DownGrade Test!

test/test_iscss.jl
```diff
@@ -0,0 +1,48 @@
+using Test
+using QuantumClifford
+using QuantumClifford.ECC
+using QuantumClifford.ECC: AbstractECC
+
+function is_css_matrix(H)
```
Krastanov, line 6: I think this test will claim that 
```
YI
IY
```
is a CSS matrix.

I believe the following would be much cleaner and correct:

```
has_x |= H[i,j][1]
has_z |= H[i,j][2]
```

test/test_iscss.jl
```diff
@@ -0,0 +1,48 @@
...
+        for j in 1:ncols
+            e = H[i, j]
+            if e == (true, false)
+                has_x = true
+            elseif e == (false, true)
+                has_z = true
+            end
+            if has_x && has_z
+                return false
+            end
```
Krastanov, line 20:
```suggestion
            has_x |= H[i,j][1]
            has_z |= H[i,j][2]
            has_x && has_z && return false
```

test/test_iscss.jl
```diff
@@ -0,0 +1,48 @@
...
+        end
+    end
+    return true
+end
+
+function test_code(code)
+    is_css = iscss(code)
+    H = parity_checks(code)
+    is_css_matrix_result = is_css_matrix(H)
+    println("Is CSS: ", is_css)
```
Krastanov, line 30: do not print in tests, this would significantly pollute the logs, making them difficult to read

test/test_iscss.jl
```diff
@@ -0,0 +1,48 @@
...
+    # Test Steane7
+    @testset "Steane7" begin
+        test_code(Steane7())
+    end
+
+    # Test Gottesman(3)
+    @testset "Gottesman" begin
+        test_code(Gottesman(3))
+    end
+end
```
Krastanov, line 48: rework this in the form of:
```
for code in known_css_codes
    @test iscss(code) == is_css_matrix(parity_checks(code))
end
for code in known_noncss_codes
...
```

Krastanov: Currently your tests are just printing things, not actually testing for properties. It should be the opposite, a human should not need to read the tests, rather the test runner will notify the human if the tests fail.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 # News
 
+## v0.8.22 - 2024-03-19
+
+- adding iscss function
```
Krastanov, line 11: Add a full sentence and proper formatting. E.g. use backticks to mark `iscss` as code and explain what it is for.

test/runtests.jl
```diff
@@ -63,6 +63,7 @@ end
 @doset "ecc_encoding"
 @doset "ecc_syndromes"
 @doset "ecc_gottesman"
+@doset "iscss"
```
Krastanov, line 66: let's call it `ecc_iscss` to keep the convention followed by the other ECC-related files

test/test_iscss.jl
Krastanov: actually, no, the tests should be more thorough than what I wrote above and there is no need to split it in two separate lists. You want this tested for all codes and the equality will be valid no matter whether the code is css or not.

src/ecc/codes/clevecode.jl
```diff
@@ -3,6 +3,10 @@ struct Cleve8 <: AbstractECC end
 
 code_n(c::Cleve8) = 8
 
+function iscss(::Type{Cleve8})
+    return false
+end
+
```
Krastanov, line 9:
```suggestion
```

This is not needed because of the default you have already set for all subtypes of AbstractECC

test/test_ecc_iscss.jl
```diff
@@ -0,0 +1,42 @@
...
+    end
+    return true
+end
+
+function test_code(code)
+    is_css = iscss(code)
+    H = parity_checks(code)
+    is_css_matrix_result = is_css_matrix(H)
+    if is_css
+        @assert is_css_matrix_result == true "If code is CSS, then parity_check_matrix must be in CSS form."
```
Krastanov, line 26: this should be `@test` not `@assert`

test/test_ecc_iscss.jl
```diff
@@ -0,0 +1,42 @@
...
+known_all_codes = [Shor9(), Steane7(), Gottesman(3), Cleve8(), Perfect5(), Toric(8,8), CSS([0 1 1 0; 1 1 0 0], [1 1 1 1]), Bitflip3()]
+
+for code in known_all_codes
+    @test iscss(code) == is_css_matrix(parity_checks(code))
+end
+
+@testset "Test parity check matrix" begin
+    @testset "known codes" begin
+        for code in known_all_codes
+            test_code(code)
```
Krastanov, line 39: you seem to be running the same test twice. You can just keep line 32-34 and delete the rest of the unused stuff.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 # News
 
+## v0.8.22 - 2024-03-19
+
+- Added `iscss` function to determine whether a given code is a CSS (Calderbank-Shor-Steane) code and to verify if its parity check matrix is indeed in CSS form.
```
Krastanov, line 11: This does not seem like a correct description. It is not verifying its parity check matrices. It just says whether the code is known to be a CSS code. The tests are running actual verifications of course, but that is not part of the function.

This matters, because the way it is currently written it sounds like `iscss` is a slow function that needs to read the matrix.

Krastanov (changes requested):
Thank you! This is getting pretty close. I left a few comments.

Krastanov:
I also see you are making a ton of small commits just as you are working through things.

It helps to do the following:

- try to make commits that are somewhat selfcontained and tell a clear story of how the work is evolving
- use the `--amend` flag to just add to the last commit instead of creating a new commit, if they are not logically separated
- aim to push in one big batch, not after each commit, because each time you push, fairly expensive automated testing is being run on your code. No need to test code you know is unfinished. Generally, run the tests locally (with the `]test QuantumClifford` command) before pushing, to make sure the tests pass for you first

src/ecc/codes/gottesman.jl
```diff
@@ -12,6 +12,10 @@ struct Gottesman <: AbstractECC
     end
 end
 
+function iscss(::Type{Gottesman})
+    return false
+end
```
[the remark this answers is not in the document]
Fe-r-oz: You are very right!

test/test_ecc_iscss.jl
Fe-r-oz: You are right!

Krastanov:
do not worry about the buildkite failure -- it will work again after a few new version registrations get propagated through the ecosystem

Fe-r-oz:
> I also see you are making a ton of small commits just as you are working through things.

Thanks for your comments. I will keep this in mind. Actually, this one had merge conflicts, and commit from other branches got mixed with this branch.

test/test_ecc_iscss.jl
```diff
@@ -0,0 +1,35 @@
...
+    end
+    return true
+end
+
+function test_code(code)
+    H = parity_checks(code)
+    @test iscss(code) == is_css_matrix(H)
+    if iscss(code)
+        @test is_css_matrix(H) == true
+    end
```
Krastanov, line 26: this does not make much sense to me. If line 23 is correct, then line 25 will also be correct. The test can be much simpler (we do not even need a separate function for this, it can all be in line 33)

test/test_ecc_iscss.jl
```diff
@@ -0,0 +1,35 @@
...
+
+function is_css_matrix(H)
+    nrows, ncols = size(H)
+    for i in 1:nrows
+        has_x = false
+        has_z = false
+        for j in 1:ncols
+            e = H[i, j]
+            has_x |= H[i,j][1]
+            has_z |= H[i,j][2]
```
Krastanov, line 14: `e` is not used here

Krastanov:
There are two rather minor issues, but they are also issues that are simply due to inattentiveness, not to some big conceptual problem. Please make sure you reread your code closely before submitting for reviews -- It is important to build up the habit to detect this type of minor issues on your own.

test/test_ecc_iscss.jl
Fe-r-oz: this e is from the previous approach!

test/test_ecc_iscss.jl
Fe-r-oz: yes, it's seem line 24 to 25 does any difference since line 23 asks the same thing!

Fe-r-oz: yeah, we can put 22 and 23 at line 33 34, delete line 33. no use in calling the function

Fe-r-oz:
> There are two rather minor issues

Minor issues are dealth with

Krastanov:
I made a minor rephrasing in the changelog. I will wait for the tests and then merge. Thank you, having this done is very helpful!

Krastanov:
Be more careful with title and references. This closes issue 222, not issue 212. A bit too late to change the commit message now, so we can just close it manually.

Fe-r-oz:
> I made a minor rephrasing in the changelog. I will wait for the tests and then merge. Thank you, having this done is very helpful!

Thank you as well!

PR #244, opened 2024-03-20
adding classical Reed-Muller Code

Fe-r-oz:
Prof. Stefan,

Adding Classical **Reed-Muller Code, RM (r, m )**. This code will be the second addition after `RepCode` which is a important subroutine in already implemented codes such as` Toric.`

Reed Muller Code will have a similar objective, as classical reed muller codes can be used to create quantum codes.

While updating the changelog, I found something odd.  It has ****(breaking)** and ****(fix)** appears in the main master branch. Please see to it.

Added **Test Throws** as well.

I tested locally as well.

```
Reed-Muller Code Parameters: ReedMuller(2, 3)
Parity Matrix:
[1 1 1 1 1 1 1 1; 1 1 1 1 0 0 0 0; 1 1 0 0 1 1 0 0; 1 0 1 0 1 0 1 0; 1 1 0 0 0 0 0 0; 1 0 1 0 0 0 0 0; 1 0 0 0 1 0 0 0]

Reed-Muller Code Parameters: ReedMuller(1, 3)
Parity Matrix:
[1 1 1 1 1 1 1 1; 1 1 1 1 0 0 0 0; 1 1 0 0 1 1 0 0; 1 0 1 0 1 0 1 0]

### CODE REVIEW

**Comment by Krastanov** [comment]
This is great! Thank you! I will leave some superficial comments today, just on style or more obvious things. I will try to review the more sophisticated math content later

File: src/ecc/ECC.jl [hunk 1]
```diff
@@ -17,6 +17,7 @@ export parity_checks, parity_checks_x, parity_checks_z, iscss,
     isdegenerate, faults_matrix,
     naive_syndrome_circuit, shor_syndrome_circuit, naive_encoding_circuit,
     RepCode,
+    ReedMuller,
```
Krastanov, line 20: Let's not export it for now, until we see where it will be used.

src/ecc/codes/reedmullercode.jl
```diff
@@ -0,0 +1,31 @@
...
+
+    function ReedMuller(r, m)
+        if r < 0 || m < 1
+            throw(ArgumentError("Invalid parameters: r must be non-negative and m must be positive"))
+        end
+        new(r, m)
+    end
+end
+
+function construct(m, i)
```
Krastanov, line 15: rename this so it is not confusing when people see it in a listing out of context. Maybe something like `_construct_reedmuller`, maybe even with a leading underscore to mark it is some private function

src/ecc/codes/reedmullercode.jl
```diff
@@ -0,0 +1,31 @@
...
+    return repeat([fill(1, 2^(m - i - 1)); fill(0, 2^(m - i - 1))], outer = 2^i)
+end
+
+function vmult(vecs...)
+    return [reduce(*, a, init=1) for a in zip(vecs...)]
+end
+
+function parity_checks(c::ReedMuller)
```
Krastanov, line 23: There does not seem to be any tests or correctness checks for this. How can someone know it is correct? Are there some properties or examples that can be verified? Is there a reference on which you based this code?

test/test_throws.jl
```diff
@@ -65,3 +66,6 @@ for m in [sMX,sMZ,sMY,sMRX,sMRZ,sMRY]
     @test_throws ArgumentError m(0,1)
     @test_throws ArgumentError m(-1,0)
 end
+
+@test_throws ArgumentError ReedMuller(-1, 3)
```
Krastanov, line 70: could you make a new file for these, so that it is easy to keep track what is coming from a submodule. Otherwise this will become difficult to manage. E.g. `test_ecc_throws.jl`

src/ecc/ECC.jl
```diff
@@ -345,4 +346,5 @@ include("codes/shorcode.jl")
 include("codes/clevecode.jl")
 include("codes/toric.jl")
 include("codes/gottesman.jl")
+include("codes/reedmullercode.jl")
```
Krastanov, line 349: to keep things consistent with the rest of the filenames, let's name it `reedmuller` instead of `reedmullercode`. Actually, maybe a whole new folder `classical` would make sense here.

Krastanov (changes requested):
Some minor style comments, nothing particularly significant.

It needs tests.

Could you let me know what you used as a reference when implementing this?

> While updating the changelog, I found something odd. It has **(breaking) and **(fix) appears in the main master branch. Please see to it.

I do not see anything, let me know if I have missed something.

src/ecc/codes/reedmullercode.jl
Fe-r-oz: You are right. I tested all these functions locally, I will add all of their tests, I didn't added their tests as I  wanted to ask you which files to add their tests. 

But it seems I will create a new file for reed muller to add their tests as you pointed out!

Fe-r-oz:
> This is great! Thank you! I will leave some superficial comments today, just on style or more obvious things. I will try to review the more sophisticated math content later


I verified the results of R(2,3) and R(1,3) from wiki: They have the complete[ matrices ](https://en.wikipedia.org/wiki/Reed%E2%80%93Muller_code)

As the the logic and how we generate these codes, here is a proper reference from [Google Scholar](https://scholar.google.com/citations?view_op=view_citation&hl=en&user=BRN7EZMAAAAJ&citation_for_view=BRN7EZMAAAAJ:Tyk-4Ss8FVUC). This has been very helpful as it explained them very simply.

Fe-r-oz:
> It needs tests.

I will create a test file. I was looking were were Rep Code tests were defined, but as you pointed out, I will tests all the functions. 

As for changelog, I see this **breaking** at some lines,.

Screenshot_select-area_20240321015410

Krastanov:
on "breaking": The changelog is there to tell other people about what has changed so they do not need to look at the git history. Last night I released version 0.9.0 (a "breaking" release as defined in SemVer https://semver.org/ ) and so the changelog has these breaking changes clearly marked up so that users know if they need to change something about their use of the library.

on references: Add that paper to the bibtex and reference it in the docstring. Also, due to a current limitation of how bibtex references are managed, add a mention of it in the list at the bottom of references.md

on tests: yeah, it is possible that `test_throws.jl` contains things that would have been better put in a different file. This will happen frequently, situations where someone reviews your code and asks you to do something to a better standard than what is already done in the library. It is the "leave a place better than you found it" principle, and it is the only way to not devolve into a mess, but it can be understandably annoying to the new contributor.

When you add the tests, mark in comments where the "correct" results in the comparisons are coming from.

src/ecc/codes/reedmullercode.jl
Fe-r-oz: This function generates the binanry variables xi which are building blocks of Rm codes. I will name it appropriately,

src/ecc/ECC.jl
Fe-r-oz: For testing purpose, I need to export it so that I can used it in `test_ecc_reedmuller.jl`

Krastanov:
I mentioned this before but just a reminder: do not send a ton of small commits one after the other. That causes a ton of emails and a ton of expensive test runs on unfinished code. Do a push when you have something ready.

Fe-r-oz:
> I mentioned this before but just a reminder: do not send a ton of small commits one after the other. 

Sorry for that. I did send the first commit will most files. Actually, I missed some changes.

docs/src/references.bib
```diff
@@ -262,3 +262,38 @@ @article{gottesman1996class
...
+  author={Abbe, Emmanuel and Shpilka, Amir and Ye, Min},
+  journal={IEEE Transactions on Information Theory},
+  volume={67},
+  number={6},
+  pages={3251--3277},
+  year={2020},
+  publisher={IEEE}
+}
+
+@incollection{DJORDJEVIC2021193,
```
Krastanov, line 286: For the reference label, could you keep them in google scholar format, just for consistency? The content of the reference here is much more thorough than what google scholar gives you, but just for the label, keep their format. E.g. `djordjevic2021information`

docs/src/references.bib
```diff
@@ -262,3 +262,38 @@ @article{gottesman1996class
...
+title = {Chapter 6 - Information Theory and Classical Error Correcting Codes},
```
Krastanov, line 287: remove "Chapter 6" from the title

src/ecc/codes/classical/reedmuller.jl
```diff
@@ -0,0 +1,38 @@
+"""The family of `[[2ᵐ, ∑ⱼ₌₀ʳ ᵐC₍ⱼ₎, 2ᵐ⁻ʳ`]] Reed-Muller codes, as described in [Raaphorst's 2003 paper](@cite raaphorst2003reed).
```
Krastanov, line 1: The backtick seems to be in the wrong location.

The reference here is kinda wrong. You can say you can consult Raaphorst's tutorial, but they simply have the equivalent of lecture notes on the topic. They are not the ones that have introduced these codes.

Krastanov: I would also suggest deleting the `[[...]]` string -- it is not legible and understandable.

Fe-r-oz: Then, we can use the [second reference,](https://arxiv.org/abs/quant-ph/9608026) it's much more pedagogical. 

Yes, this first reference is more like Preskill notes.

test/test_ecc_reedmuller.jl
```diff
@@ -0,0 +1,45 @@
+using Test
+using QuantumClifford
+using QuantumClifford.ECC
+using QuantumClifford.ECC: AbstractECC
+
+@testset "Testing common examples RM(r,m) codes [raaphorst2003reed](@cite) and [abbe2020reed](cite)" begin
```
Krastanov, line 6: this is great!

I am trying to think trough what would be a more universal way to test these. I am think potentially checking the rank of the matrices would make sense. You can try something like:

```
for r in ... for m in ...
    H = parity_checks(ReedMuler...)
    @test size(H) == something that depends on r and m
    mat = Nemo.matrix(Nemo.GF(2), H) # necessary to convert to a matrix type for which `rank` works correctly
    @test LinearAlgebra.rank(mat) == ...
end
```

The weird conversion from `H` to `mat` is because `rank` by default assumes you are asking for the matrix rank over the real numbers, not for the rank over the boolean field.

src/ecc/codes/classical/reedmuller.jl
Krastanov: It is more about phrasing. The reference you have here already is good pedagogical reference. Rather, my point is that your current phrasing makes it sound like these guys invented the RM code.

One possible solution about the phrasing: You can remove this reference from the header and just include it as one of the other "consult also" references. Including the latest one from your last comment also makes sense.

test/test_ecc_reedmuller.jl
Fe-r-oz: Yes!

We can compute the rank using the standard formula for Reed Muller codes that is given in the refernce which is that `rank = sum from i to r  [m choose i]` and compare it with rank calculated via Nemo. It's working. I tested locally!

src/ecc/codes/classical/reedmuller.jl
Fe-r-oz: Point noted. I will do all these suggested changes and do a single commit.

Krastanov:
you do not need to make a single commit, but please make sure you are making a single push, not a ton of frequent pushes

src/ecc/codes/classical/reedmuller.jl
```diff
@@ -0,0 +1,38 @@
+"""The family of Reed-Muller codes, as discovered by Muller in his 1954 paper [muller1954application](@cite) and Reed who proposed the first efficient decoding algorithm [reed1954class](@cite).
+
+You might be interested in consulting [raaphorst2003reed](@cite), [abbe2020reed](@cite), and [djordjevic2021quantum](@cite) as well.
+
+The ECC Zoo has an [entry for this family](https://errorcorrectionzoo.org/c/reed_muller)
```
Krastanov, line 5: wonderfully written docstring

Fe-r-oz: Thanks!

src/ecc/codes/classical/reedmuller.jl
```diff
@@ -0,0 +1,38 @@
...
+
+function vmult(vecs...)
+    return [reduce(*, a, init=1) for a in zip(vecs...)]
+end
+
+function parity_checks(c::ReedMuller)
+    r=c.r
+    m=c.m
+    xi = [variables_xi(m, i) for i in 0:m - 1]
+    row_matrices = [reduce(vmult, [xi[i + 1] for i in S], init = ones(Int, 2^m)) for s in 0:r for S in combinations(0:m - 1, s)]
```
Krastanov, line 34: I imagine the 2^m will overflow quickly. Could you add a check in the constructor that throws an error if m is too big?

Krastanov:
This looks really great, thank you for contributing it! I have one very minor comment, but otherwise this seems ready to merge.

Fe-r-oz:
> This looks really great, thank you for contributing it!

Thank you! I learned a lot as well!

Krastanov:
there was another thing I fixed in the last commit. You were calling `collect`, which can be an extremely expensive operation. Directly calling `length` is cleaner. I will wait for the tests and merge. Thank you!

Krastanov:
Thank you for the addition!

PR #246, opened 2024-03-22
Surface code

royess:
The PR adds surface codes, which are tailored toric codes that require less boundary connectivity,

- Implement surface codes based on existing toric code implementation.
- Implement a specially tailored 9-qubit surface code.
- Improve the docstrings for the toric code and add its reference paper.


royess:
I have fixed an overwrite bug; now, most CI tasks are passed. I noticed some failures, but they seem to be unrelated to this PR. 

@Krastanov, could you help review this PR? Thanks!

src/ecc/codes/toric.jl
```diff
@@ -1,4 +1,27 @@
-"""The Toric code."""
+"""The Toric code [kitaev_fault-tolerant_2003](@cite).
+
+Illustration of a 2x2 toric code, where qubits are located on the edges:
+
+|--1-(Z)-2--|
```
Krastanov, line 5: these diagrams should be in code blocks (backticks), otherwise they will render very weirdly in the html docs.

src/ecc/codes/surface.jl
```diff
@@ -0,0 +1,79 @@
...
+
+struct Surface9 <: AbstractECC end
+
+function iscss(::Type{Surface9})
+    return true
+end
+
+code_n(c::Surface9) = 9
+
+parity_checks(c::Surface9) = S"
```
Krastanov, line 68: I do not think we need a separate `Surface9` code.

src/ecc/codes/surface.jl
```diff
@@ -0,0 +1,79 @@
+"""The surface code [fowler2012surface](@cite), which is a tailored toric code with only the nearest neighboring connections in a 2D plane.
```
Krastanov, line 1: Fowler should probably not be the first reference, they are not the ones that discovered the code. But they can be a reference added elsewhere in the docstring.

If I get this right (I might not), this is the bad option for surface codes. I have not checked the tableau very carefully, but consult these two:

- I see at https://errorcorrectionzoo.org/c/surface that the code is `[[2L²-2L-1, 1, L]]`
- or at https://errorcorrectionzoo.org/c/rotated_surface is `[[L²,1,L]]` which is two times better

It is perfectly valid to implement first this code, we should have it. But we should probably comment on this and warn people that they should probably be using the rotated surface code.

src/ecc/codes/surface.jl
Krastanov: oh, I misunderstood. This is not simply an instance of the surface code you have already implemented. It makes sense to have this one, but probably under a different name and with a docstring that explains what it is. Maybe the name can be `Tomita25`

test/test_ecc_decoder_all_setups.jl
```diff
@@ -76,7 +77,9 @@ import LDPCDecoders
 @testset "matching decoder, good as long as column weight of the code is limited" begin
     codes = [
         Toric(8,8),
-        Toric(9,9)
+        Toric(9,9),
+        Surface(8,8),
+        Surface(9,9)
```
Krastanov, line 82: Adding your `Surface9` code (or however it would be name) to all tests would make sense.

Krastanov (changes requested):
Thanks! This looks great!

src/ecc/codes/surface.jl
royess: A different name would be better, I agree.

If we keep the convention that the number in the name is the qubit number, `Tomita9` probably would be a better name.

src/ecc/codes/surface.jl
royess: > If I get this right (I might not), this is the bad option for surface codes.

I am unsure whether the rotated surface code should be considered a better option. 

The `[[L²,1,L]]` mentioned in the link (https://errorcorrectionzoo.org/c/rotated_surface) refers to a small 9-qubit code I previously implemented as `Surface9`. I don't expect this "advantage" to remain when the number of qubits gets large. 

Basically, the rotated surface code is designed to save qubits. That probably should come with some sacrifices in error tolerance or error threshold. (Also, it brings some subtlety to the transversal H gate, if I remember correctly.)

So, for a solution, I suggest we do as follows:

- Add a mention of the rotated surface code in `Surface`'s docstring.
- Re-name Tomita's 9-qubit code (old name: `Surface9`) as `RotatedSurface9` to serve as an example of the rotated surface code.
- (We may consider leaving the rotated surface code for arbitrary size to the future PRs.)

royess: For clarity, the term surface code seems ambiguous in various literature. In general, the surface code family can contain the toric code and many of its variants.

The code I implemented in this PR should be more precisely called "planar surface code." However, people usually use the simpler name "surface code." I should probably clarify this in the docstring to avoid ambiguity.

src/ecc/codes/toric.jl
```diff
@@ -1,4 +1,27 @@
...
+| (X) 5     6
+|--3--|--4--|
+|     7     8
+|     |     |
```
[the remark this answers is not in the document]
royess: That makes sense. Thanks!

test/test_ecc_decoder_all_setups.jl
royess: That makes sense, thanks!

Krastanov:
- Let's stick to the name "Surface".
- If the 9-qubit example you have is simply an instance of the rotated surface code, let's delete it -- no need to have a single instance of a family without having the family.

docs/src/references.bib
```diff
@@ -263,6 +263,49 @@ @article{gottesman1996class
   publisher={APS}
 }
 
+
+@article{kitaev_fault-tolerant_2003,
```
Krastanov, line 267: for consistency, I prefer if we keep all the names in "google scholar" format, e.g. `kitaev2003fault` here.

src/ecc/codes/surface.jl
Krastanov: I mostly agree. Let's keep the name `Surface` and if the 9-qubit one is just an instance of the rotated surface family, let's just delete it for now.

royess:
Thanks for your suggestions! The new commit removed the 9-qubit (rotated) Surface code and did the aforementioned formatting.

docs/src/references.bib
royess: Fixed, thanks!

src/ecc/codes/surface.jl
```diff
@@ -0,0 +1,60 @@
...
+|---4---|---5---|---6---|
+|       o       o       o
+|       |       |       |
+```
+
+The surface code has open boundary conditions compared to the toric code. To this end, we remove qubits (denoted by "o") and parity checks on the right and bottom sides.
+
+Faces like (1,4,7) have X checks, and crosses like (1,2,7) have Z checks. Due to the removal of the bottom and right sides, we have some 3-qubit checks on the boundaries.
+
+```julia
```
Krastanov, line 17: Could you turn this into a doctest, so that it is executed when testing? It helps keep the docs in sync with implementation (e.g. if at some point we have a reason to default to a different canonical form).

src/ecc/codes/toric.jl
```diff
@@ -1,4 +1,29 @@
...
+```
+
+It's important to note that the toric code has periodic boundary conditions, which means that the top and bottom sides are essentially glued together, as are the left and right sides.
+
+Faces like (1,3,5,6) have X checks, and crosses like (1,2,5,7) have Z checks.
+
+```
```
Krastanov, line 17: same comment about making it a doctest

src/ecc/codes/surface.jl
Krastanov, line 17: Could you turn this into a doctest, so that it is executed when testing? It helps keep the docs in sync with implementation (e.g. if at some point we have a reason to default to a different canonical form).

src/ecc/codes/toric.jl
```diff
@@ -1,4 +1,29 @@
...
+|--1-(Z)-2--|
+| (X) 5     6
+|--3--|--4--|
+|     7     8
+|     |     |
+```
+
+It's important to note that the toric code has periodic boundary conditions, which means that the top and bottom sides are essentially glued together, as are the left and right sides.
+
+Faces like (1,3,5,6) have X checks, and crosses like (1,2,5,7) have Z checks.
```
Krastanov, line 15:
```suggestion
Faces like `(1,3,5,6)` have X checks, and crosses like `(1,2,5,7)` have Z checks.
```

src/ecc/codes/surface.jl
```diff
@@ -0,0 +1,60 @@
+"""The planar surface code refers to the toric code [kitaev2003fault](@cite) in a 2D lattice with open boundaries.
```
Krastanov, line 1:
```suggestion
"""The planar surface code refers to the code [kitaev2003fault](@cite) in a 2D lattice with open boundaries.
```
Minor phrasing: It is not "toric" if it has open boundaries.

src/ecc/codes/surface.jl
```diff
@@ -0,0 +1,60 @@
...
+
+Illustration of a 3x2 surface code, where qubits are located on the edges:
```
Krastanov, line 3:
```suggestion
Illustration of a 3×2 surface code, where qubits are located on the edges:
```

Minor typography polish, using the unicode "times" symbol instead of the letter x.

src/ecc/codes/surface.jl
```diff
@@ -0,0 +1,60 @@
...
+
+```
+|---1--(Z)--2---|---3---|
+|  (X)  7       8       o
+|---4---|---5---|---6---|
+|       o       o       o
+|       |       |       |
+```
+
+The surface code has open boundary conditions compared to the toric code. To this end, we remove qubits (denoted by "o") and parity checks on the right and bottom sides.
```
Krastanov, line 13:
```suggestion
The surface code has open boundary conditions, unlike the toric code. To this end, we remove qubits (denoted by "o") and parity checks on the right and bottom sides.
```

src/ecc/codes/surface.jl
```diff
@@ -0,0 +1,60 @@
...
+
+Faces like (1,4,7) have X checks, and crosses like (1,2,7) have Z checks. Due to the removal of the bottom and right sides, we have some 3-qubit checks on the boundaries.
```
Krastanov, line 15:
```suggestion
Faces like `(1,4,7)` have X checks, and crosses like `(1,2,7)` have Z checks. Due to the removal of the bottom and right sides, we have some 3-qubit checks on the boundaries.
```

src/ecc/codes/toric.jl
```diff
@@ -1,4 +1,29 @@
...
+
+```
+|--1-(Z)-2--|
+| (X) 5     6
+|--3--|--4--|
+|     7     8
+|     |     |
+```
+
+It's important to note that the toric code has periodic boundary conditions, which means that the top and bottom sides are essentially glued together, as are the left and right sides.
```
Krastanov, line 13:
```suggestion
It is important to note that the toric code has periodic boundary conditions, which means that the top and bottom sides are essentially glued together, as are the left and right sides.
```

Minor style change. In modern American English, no one cares about this, but I try to stick to the older "no contractions" style.

Krastanov (changes requested):
This looks awesome! I made some minor style suggestions and a request to plug the code example docs into the test runner, but otherwise this is good to merge.

src/ecc/codes/surface.jl
royess: That makes sense, thanks!

PR #247, opened 2024-03-23
Steane-Reed-Muller code

Fe-r-oz:
Prof. Stefan,

Adding the implementation of **Steane Reed-Muller(QRM)** codes. [Paper ](https://arxiv.org/pdf/quant-ph/9608026.pdf)followed is by the Steane who discovered them.

- The algorithm is built straight from the paper. The math is closely followed.
- ` ReedMuller `codes are utilized as discussed in the paper! 
- Two implementations are attached (Table 14 and 18 of the Paper). Matrix Notation of RM (G1, G2) is different: order of rows is different in the papers that follow it. 
- This paper presents slightly different way of representing the classical RM matrix but later papers on RM as of 2020 follow different representation. 
- It is found the `isCSS` function that we implemented for CSS codes is insufficient to test whether QRM codes are CSS if they are CSS. This is because of the  X, Z, Y we have in QRM because of Dx and Dz. Please see stabilizers 14, 16, 18 in Steane's paper. (Resolved)
- On a small note, I have changed the notation to RM (m, r) instead of RM (r, m) to follow closely with the quantum version.  t is same as m. (Resolved)
- There are some ambiguities when Steane goes onto explain when code size falls to zero at k = 0, sub-optimal codes, codes for which "the method fails for k = 0", etc. (Resolved)

```
julia> parity_checks(QuantumReedMuller(1,3))
+ XXXXXXXX
+ XXXX____
+ XX__XX__
+ X_X_X_X_
+ ZZZZZZZZ
+ ZZZZ____
+ ZZ__ZZ__
+ Z_Z_Z_Z_
+ YXZ_____
+ Y_X_Z___
+ XZZ_X___

julia> parity_checks(QuantumReedMuller(2,5))
+ XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
+ XXXXXXXXXXXXXXXX________________
+ XXXXXXXX________XXXXXXXX________
+ XXXX____XXXX____XXXX____XXXX____
+ XX__XX__XX__XX__XX__XX__XX__XX__
+ X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_X_
+ XXXXXXXX________________________
+ XXXX____XXXX____________________
+ XX__XX__XX__XX__________________
+ X_X_X_X_X_X_X_X_________________
 ⋮
+ YYXXZZ__YYXXZZ__________________
+ YXZ_YXZ_YXZ_YXZ_________________
+ YZYZX_X_X_X_X_X_ZZZZ____________
+ YYXXZZ__________YYXXZZ__________
+ YXZ_YXZ_________YXZ_YXZ_________
+ YZX_X_X_ZZ______YZX_X_X_ZZ______
+ YXZ_____YXZ_____YXZ_____YXZ_____
+ Y_X_Z___Y_X_Z___Y_X_Z___Y_X_Z___
+ X___YZZZYZZZX___X___X___X___X___

```


Fe-r-oz:
Some ambiguities are now clear. The code size does fall to 0 at t = 4, r = 2 given the constraints he has set. He was trying to explain this in his paper.

equation(2)

Fe-r-oz:
Improvements:
- The algorithm is now completed.
- The bounds are added to ensure code tractibility and validity.

Focus is now towards tests.

Fe-r-oz:
Improvements:

1. The Dx Generation is improved. It now closely follows with what is expected from the paper.
2. Initial Tests are added.

Main algorithm is now less resource expensive.  (40ish lines) 

The algorithm is based on 4 points [(Page 3, Page 4)](https://arxiv.org/pdf/quant-ph/9608026.pdf). How to generate
- G1
- G2
- Dx
- Dz

I have followed these point and the math presented closely.  Kindly confirm the algorithm implementation by comparing it from the paper. 

I have some ambigities about the cases where the method fails are pointed out by Steane, when the code size falls to zero, when it is suboptimal, and what it is less optimal, or close to optimal. Does this mean that being optimal is connected with the performance of the code itself? A suboptimal code then will perform less optimally in the tests as well.

src/ecc/codes/quantumreedmuller.jl
```diff
@@ -0,0 +1,62 @@
...
+    end
+end
+
+function k(t, r)
+    return 2^r - sum(binomial.(r, 0:t - 1)) 
+end 
+
+function n(t, r)
+    return 2^r - sum(binomial.(r, 0:t)) 
+end
```
Fe-r-oz, line 27: This is unnecessary. I can t - 1 or t and make one function.

Fe-r-oz:
@Krastanov Please review the PR when you find convenient and when the time is right. I re-read the paper several times so that the standard algorithm and math is closely followed.  Thanks!

Fe-r-oz:
> Does this mean that being optimal is connected with the performance of the code itself? A suboptimal code then will perform less optimally in the tests as well.

So, I did a little investigation and it founds out the it's correct that sub optimal and less optimal codes don't perform well with the distance test. I took the examples from the paper and found out that: This really solves the ambiguities!

What's shown below is the Dx and Dz, taking their bitwise OR. The whole point of generating these matrices, as told by paper is to reduce the min distance. Conclusion presented at the bottom: 

```
QRM (t, r) => (2, 5) SUB OPTIMAL

00000000000000000000000011111111 
00000000000000000000111100001111
--------------------------------
                    11111111111 = 12

00000000000000000000111100001111
00000000000000000011001100110011 
---------------------------------
                  111111  111111 = 12

00000000000000000011001100110011 
00000000000000000101010101010101
--------------------------------
                 111 111 111 111 = 12
00000000000000000101010101010101
00000000000011110000000000001111
--------------------------------
            1111  1 1 1 1  1111 = 12  

00000000000011110000000000001111
00000000001100110000000000110011
--------------------------------
          111111          111111 = 12

00000000001100110000000000110011 
00000000010101010000000001010101
--------------------------------
         111111           111111  = 12
 
00000000010101010000000001010101 
00000011000000110000001100000011
--------------------------------
      11 1 1 1 1       11 1 1 11 = 12  
     
00000011000000110000001100000011 
00000101000001010000010100000101
--------------------------------
     111      111      111      111  = 12
 
min distance = 12 sub optimal


QRM (t, r) => (2, 4) OPTIMAL
0000000000001111 
0000000000110011
                 = 6
---------------------
0000000000110011 
0000000001010101
                 = 6
---------------------
0000000001010101 
0000001100000011
                 = 7
---------------------
0000001100000011 
0000010100000101
                 = 6
---------------------
0000010100000101 
0001000100010001
                 = 6
---------------------
0001000100010001 
0000000000111100 
                 = 7

min distance = 6  optimal 


QRM (t, r) => (1, 3) OPTIMAL
00000011 
00000101
--------
         = 3
00000101 
00010001
--------
         = 3
00010001
00000110
         = 4

min distance = 3 optimal
```

That's why Steane point why there are _sub optimal_ codes, and codes that perform _less optimally_. He says that  ( 2, 5) is "_close to optimal_". 

In his words. "
- The codes are **optimal** (have maximal k for given n and d) for t = 1 [12], and
-  **close to optimal** for small t and r (comparing with the table of codes in [5], it is seen that [[32, 10, 6]] is close to optimal). 
- They are **far from optimal** for large t and r, but should form a good starting point for deriving other codes. "

Fe-r-oz:
Improvements:
- References for consultation are added as 1999 is way back and the recent papers have utilized slight different conventions. 
- Allocation free and efficient implementation of Steane's 1999 standard reference is presented. 
- Tests are added for optimal codes.

Now, it's ready for review!

src/ecc/codes/quantumreedmuller.jl
```diff
@@ -0,0 +1,65 @@
...
+You might be interested in consulting [zhang1997quantum](@cite), [quan2018fault](@cite) and [campbell2012magic](@cite) as well.
+
+The ECC Zoo has an [entry for this family](https://errorcorrectionzoo.org/c/quantum_reed_muller)
+"""
+struct QuantumReedMuller <: AbstractECC
+    t::Int  
+    r::Int  
+    function QuantumReedMuller(t, r)
+        if t == 1 && r == 2 
+            throw(ArgumentError("Invalid parameters: t = 1, r = 2 is not quantum code as it does not satisfy Hx.Hz' + Hz.Hx' = 0 [steane1999quantum](@cite)."))
```
Krastanov, line 12:
```suggestion
            throw(ArgumentError("Invalid parameters when attempting to construct a quantum Reed-Muller code. t = 1, r = 2 is not quantum code as it does not satisfy commutativity requirements."))
```

The citation would not render in an error message so I just removed it.

src/ecc/codes/quantumreedmuller.jl
```diff
@@ -0,0 +1,65 @@
...
+        elseif t < 0 || t > 7 || r < 1 || r > 7
+            throw(ArgumentError("Invalid parameters: r must be positive and < 7 and t >= 0 and < 7 in order to obtain a valid code and to remain tractable."))
```
Krastanov, line 14:
```suggestion
            throw(ArgumentError("Invalid parameters when attempting to construct a quantum Reed-Muller code. We need  0<r<7 and 0≤t< 7 in order to obtain a valid code and to remain tractable."))
```

src/ecc/codes/quantumreedmuller.jl
```diff
@@ -0,0 +1,65 @@
...
+        else
+            new(t, r)
+        end
+    end
+end
+
+function codesize(t, r)
```
Krastanov, line 21: This function name can be a bit confusing, because it applies only to qRM codes. If you need this helper, rename it to something like `_codesize_qrm`.

src/ecc/codes/quantumreedmuller.jl
```diff
@@ -0,0 +1,65 @@
...
+    return 2^r - sum(binomial.(r, 0:t)) 
+end 
+
+function nplusk(t, r)
```
Krastanov, line 25: same comment about function name

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 
 # News
 
+## v0.9.3 - 2024-03-24
```
Krastanov, line 8: We have not actually released `v0.9.1` yet, so all of the last few ones actually need to be under the same heading.

test/test_ecc_reedmuller.jl
```diff
@@ -14,7 +14,7 @@ function binomial_coeff_sum(r, m)
     return total
 end
 
-@testset "Test RM(r, m) Matrix Rank" begin
```
Krastanov, line 17: You have changed the name here, but it does not seem anything has changed about the implementation. It still reads `ReedMuller(r,m)` in the rest of the code files. What is this mismatch due to?

src/ecc/codes/quantumreedmuller.jl
```diff
@@ -0,0 +1,65 @@
+"""The family of quantum Reed-Muller codes, as discovered by Steane in his 1999 paper [steane1999quantum](@cite).
```
Krastanov, line 1: This does not include a `iscss` function. Are qRM codes always non-CSS? You made a comment about that above, but I do not understand your comment.

src/ecc/codes/quantumreedmuller.jl
```diff
@@ -0,0 +1,65 @@
...
+   
+    Hx = vcat(G1, zeros(Int64, size(G1, 1), size(G1, 2)), Dx)
+    Hz = vcat(zeros(Int64, size(G1, 1), size(G1, 2)), G1, Dz)  
+
+    extended_Hx = Matrix{Bool}(Hx)
+    extended_Hz = Matrix{Bool}(Hz)
+
+    num_rows = size(Hx, 1)
+    fill_array = fill(UInt8(0), num_rows)
+    Stabilizer(fill_array, extended_Hx, extended_Hz)
```
Krastanov, line 64: You can skip the `fill_array` argument. The `Stabilizer` has a constructor that just takes two matrices:

image

Krastanov (changes requested):
There are a few minor things, but otherwise this looks good.

I will need to run a few comparisons against code examples as a final review.

src/ecc/codes/quantumreedmuller.jl
Fe-r-oz: Indeed, you are right. I noticed this, line 62 and 63 are not required!

src/ecc/codes/quantumreedmuller.jl
Fe-r-oz: Thanks! I named them initially based on the formulas as reminder such  n+k that gives total rows. I will give the better names

test/test_ecc_reedmuller.jl
Fe-r-oz: Yes, the implementation is not changed. It's the naming convention.
It's actually R(m, r) in the implementation, but I confused the naming them as R(m,r).  In the earliest papers such as Steane, they used (t,r). Also, in the research paper, they used (m,r) which is the same as (t,r).

src/ecc/codes/quantumreedmuller.jl
Fe-r-oz: QRM (1, 3) when t =1, and r =3 is actually Gottesman [[8,3,3]] code which is not CSS. 

I believe QRM codes are not CSS given the Dx and Dx matrices that we used to construct QRM codes. They might look CSS since G1 is horizontally concatentated with zeros on first row, and G1 is concatentated with zeros in the second row, but the Dx and Dz in the last row, when they are combined into logical operators, it becomes non - CSS.

But ECC Zoo calls them CSS, that's why I commented it as ambiguity. I mean if (1,3) is a Gottesman code, all t=1 are gottesman code, then how can they be named CSS?

Fe-r-oz:
> I will need to run a few comparisons against code examples as a final review.

That would be helpful. When making comparisons, please note the slight change in the matrix representation that Steane Used.

Note how this matrix is represented. from Left to Right.
1 1 1 1 0 0 0 0
1 1 0 0 1 1 0 0
1 0 1  0 1 0 1 0
1 1 1 1 1 1 1 1  

But in Steane, it's represented from Right to Left!
1 1 1 1 1 1 1 1  
0 0 0 0 1 1 1 1 
0 0 1 1 0 0 1 1 
0 1 0 1  0 1 0 1 

Screenshot from 2024-03-27 18-46-34

For example, in this 2020 Paper the RM (1,3) code is same as RM (3,1) given m = t.

test/test_ecc_reedmuller.jl
Krastanov: I do not think I understand. Currently, you have pieces of code that use the notation `(m,r)` and other pieces that say `(r,m)`. This seems inconsistent:
image
image

Fe-r-oz: Yes, it's (m, r) every where. I introduced the change in naming convention in this PR to match with the most widely used conventon. Widely, the notation used is (m, r) not (r, m).

I will change everywhere to (m, r)!

Fe-r-oz: Adopted (m, r) instead of (r, m) naming convention!

Fe-r-oz:
I think it's the Steane notation that's out of place with other papers.

If we agree that m = 3, and r =1 (3, 1) provides matrix shown above, We can say it any way for RM (3 ,1) , (1, 3) , r here is 1, and m =3.

But if we assume that in steane paper, where he says r = 3 and t = 1 gives the matrix that RM (3,1) (or RM (1, 3), however we want to call it) gives where r = 1, not 3!  

There was no need to change the naming convention.

Fe-r-oz:
Thanks a lot for your suggestions!

src/ecc/codes/quantumreedmuller.jl
Fe-r-oz: Please have a look at sample screenshot.

When Dx and Dz are written in Stabilizer notation, the rows that have Dx and Dz contains a lt of X,Y,Z it's not pure Xs or Zs, so it's not CSS.


Screenshot_select-area_20240327205329

test/test_ecc_reedmuller.jl
Krastanov: From the zoo:
image

According to the first of these paragraphs, the construction of a quantum RM code should be just to lines, reusing the implementation of the classical RM. According to the second paragraph, there is some non-CSS code that can then be derived by doing extra processing. Which one are you doing?

Fe-r-oz: I am following construction from the Steane Paper:

Screenshot from 2024-03-27 22-44-46

The ECC zoo is citing Steane Paper.

src/ecc/codes/quantumreedmuller.jl
Krastanov: Reading Steane's paper, equation 7 gives one family of codes and then equations 8 gives another family of codes:
image

The first one is a CSS code and the second one is not. I think you have implemented the second one?

I have the impression that the ECC Zoo wiki calls the first one "quantum Reed Muller code" and the second one "modified quantum Reed Muller code". Maybe we can keep the same nomenclature and keep both?

Maybe we can name them `CSSReedMuller` and `SteaneReedMuller` here? The `SteaneReedMuller` is what you have here.

Also, if a subfamily of these codes is the Gottesman codes, let's add a check for that. E.g.
```
for something in whatever
    @test parity_checks(SteaneReedMuller(...)) == parity_checks(Gottesman(...))
end
```

If there are differing conventions about parity check order, you can use `canonicalize` to make them the same.

Krastanov: also, be careful, some of these use the generator matrix and some use the parity check matrix -- these are related but different

Fe-r-oz: That makes sense! I have implemented `SteaneReedMuller` non-CSS codes from (8). 

Thanks for your suggestions!

Fe-r-oz: > If there are differing conventions about parity check order, you can use `canonicalize` to make them the same.

I believe first we need to convert the generator matrix G to  parity check Matrix H, right?

Screenshot_select-area_20240328002715

As directly using canonicize might not help?

```
Julia> parity_checks(QuantumReedMuller(1,3))
+ XXXXXXXX
+ XXXX____
+ XX__XX__
+ X_X_X_X_
+ ZZZZZZZZ
+ ZZZZ____
+ ZZ__ZZ__
+ Z_Z_Z_Z_
+ YXZ_____
+ Y_X_Z___
+ XZZ_X___

julia> canonicalize!(parity_checks(Gottesman(3)))
+ XZ_Y_YXZ
+ _XZY_XZY
- _ZXYZ_YX
- __ZZYYXX
+ ZZZZZZZZ

julia> canonicalize!(parity_checks(QuantumReedMuller(1,3)))

+ X__Z__ZX
+ _X_Z_ZZY
- __X___ZY
+ ___X_ZZX
+ ___ZXZ_X
+ ___Z_X_Y
- _____ZXY
+ Z__Z_ZZ_
+ _Z_Z_Z_Z
+ __ZZ__ZZ
+ ____ZZZZ

```

Thanks!

Fe-r-oz: Please have a look:

```
julia> (stab_to_gf2(canonicalize!(parity_checks(Gottesman(3)))))
5×16 Matrix{Bool}:
 1  0  0  1  0  1  1  0  0  1  0  1  0  1  0  1
 0  1  0  1  0  1  0  1  0  0  1  1  0  0  1  1
 0  0  1  1  0  0  1  1  0  1  0  1  1  0  1  0
 0  0  0  0  1  1  1  1  0  0  1  1  1  1  0  0
 0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1

julia> gf2_H_to_G(stab_to_gf2(canonicalize!(parity_checks(QuantumReedMuller(1,3)))))
5×16 Matrix{Bool}:
 1  1  1  1  1  1  1  1  0  0  0  0  0  0  0  0
 1  1  0  0  1  1  0  0  1  1  1  1  0  0  0  0
 0  1  0  1  1  0  1  0  1  1  0  0  1  1  0  0
 1  1  1  1  0  0  0  0  1  0  1  0  1  0  1  0
 0  1  1  0  0  1  1  0  0  1  1  0  1  0  0  1

```
This is the same matrix as shown in the Figure(13). As explained in  https://github.com/QuantumSavory/QuantumClifford.jl/pull/247#issuecomment-2022825082, it's just differeing convention of matrix representation.

```
 stab_to_gf2(parity_checks(QuantumReedMuller(1,3)))
11×16 Matrix{Bool}:
 1  1  1  1  1  1  1  1  0  0  0  0  0  0  0  0
 1  1  1  1  0  0  0  0  0  0  0  0  0  0  0  0
 1  1  0  0  1  1  0  0  0  0  0  0  0  0  0  0
 1  0  1  0  1  0  1  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1
 0  0  0  0  0  0  0  0  1  1  1  1  0  0  0  0
 0  0  0  0  0  0  0  0  1  1  0  0  1  1  0  0
 0  0  0  0  0  0  0  0  1  0  1  0  1  0  1  0
 1  1  0  0  0  0  0  0  1  0  1  0  0  0  0  0
 1  0  1  0  0  0  0  0  1  0  0  0  1  0  0  0
 1  0  0  0  1  0  0  0  0  1  1  0  0  0  0  0
```

Krastanov: > I believe first we need to reduce the generator matrix G of SteaneRM to parity check Matrix H of Gottesman!

The tableaux is always "the parity check matrix". Sure, some cool constructions that use conjugate codes might use the generator matrix of one code as the parity check matrix of another code. But if the paper says the two classes of codes are the same (SteaneRM with some parameter value and Gottesman with some parameter value), then only their parity check matrices are to be compared. This library as a whole does not actually provide generator matrices at all.

The SteaneRM to Gottesman comparison test is still an important correctness check. If you have some difference in convention between the two ways you have constructed those, just account for the difference in the tests.

Krastanov:
A reminder, please make one big push, not a ton of small pushes after each change. Otherwise a lot of mail spam and unnecessary (and computationally heavy) test runs are being generated each time.

src/ecc/codes/quantumreedmuller.jl
Fe-r-oz: I have changed to steane_convention and now the parity check Matrix H matches exactly with the one presented in Steane's Paper!
```

julia> stab_to_gf2(parity_checks(QuantumReedMuller(1,3)))
11×16 Matrix{Bool}:
 1  1  1  1  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  0  0  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  1  1  0  0  1  1  0  0  0  0  0  0  0  0
 0  1  0  1  0  1  0  1  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1
 0  0  0  0  0  0  0  0  0  0  0  0  1  1  1  1
 0  0  0  0  0  0  0  0  0  0  1  1  0  0  1  1
 0  0  0  0  0  0  0  0  0  1  0  1  0  1  0  1
 0  0  0  0  0  0  1  1  0  0  0  0  0  1  0  1
 0  0  0  0  0  1  0  1  0  0  0  1  0  0  0  1
 0  0  0  1  0  0  0  1  0  0  0  0  0  1  1  0

```

Fe-r-oz:
> if the paper says the two classes of codes are the same, then only their parity check matrices are to be compared. then only their parity check matrices are to be compared.


I have changed to steane convention of representation, so it will output the correct Stabilizer/H for values of t,r :)

I have a question about comparison of the two parity check matrices. How do I compare the two matrices when they have different sizes? Please let me know if I missed anything.

The SteaneRM provides at  t = 1, r = 3, this parity check Matrix H. This is exactly what is shown in the paper,

```
julia> stab_to_gf2(parity_checks(QuantumReedMuller(1,3)))
11×16 Matrix{Bool}:
 1  1  1  1  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  0  0  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  1  1  0  0  1  1  0  0  0  0  0  0  0  0
 0  1  0  1  0  1  0  1  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1
 0  0  0  0  0  0  0  0  0  0  0  0  1  1  1  1
 0  0  0  0  0  0  0  0  0  0  1  1  0  0  1  1
 0  0  0  0  0  0  0  0  0  1  0  1  0  1  0  1
 0  0  0  0  0  0  1  1  0  0  0  0  0  1  0  1
 0  0  0  0  0  1  0  1  0  0  0  1  0  0  0  1
 0  0  0  1  0  0  0  1  0  0  0  0  0  1  1  0
```

But H of Gottesman 3 is as follows:

```
julia> using QuantumClifford: canonicalize!; stab_to_gf2(canonicalize!(parity_checks(Gottesman(3))))
5×16 Matrix{Bool}:
 1  0  0  1  0  1  1  0  0  1  0  1  0  1  0  1
 0  1  0  1  0  1  0  1  0  0  1  1  0  0  1  1
 0  0  1  1  0  0  1  1  0  1  0  1  1  0  1  0
 0  0  0  0  1  1  1  1  0  0  1  1  1  1  0  0
 0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1

julia> stab_to_gf2(parity_checks(Gottesman(3)))
5×16 Matrix{Bool}:
 1  1  1  1  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1
 1  0  1  0  0  1  0  1  0  0  0  0  1  1  1  1
 1  0  1  0  1  0  1  0  0  0  1  1  0  0  1  1
 1  0  0  1  0  1  1  0  0  1  0  1  0  1  0  1
```

Fe-r-oz:
> A reminder, please make one big push, not a ton of small pushes after each change. 

Thanks for the reminder!!

Fe-r-oz:
Improvements:
 - `_Steane_convention_qrm `is added to match exactly with steane way of representation!
- Tests for Gottesman Family are added. `Steane convention` for Gottesman (Table 14 of the paper) is used to represent H as `canonicalize! `or s`tab_to_gf2 `was reprsent different conventon from steane's. 
-  Named changed to `SteaneReedMuller ` as this is a non-CSS code
-  The `docstring` is carefully edited as well.

In Steane's convention, 
```
julia> parity_checks(SteaneReedMuller(1,4))
+ XXXXXXXXXXXXXXXX
+ ________XXXXXXXX
+ ____XXXX____XXXX
+ __XX__XX__XX__XX
+ _X_X_X_X_X_X_X_X
+ ZZZZZZZZZZZZZZZZ
+ ________ZZZZZZZZ
+ ____ZZZZ____ZZZZ
+ __ZZ__ZZ__ZZ__ZZ
+ _Z_Z_Z_Z_Z_Z_Z_Z
+ ______ZZ_X_X_XZY
+ _____ZXY_____ZXY
+ ___Z_X_Y___Z_X_Y
+ ___Y_Z_Y_Z_X___X
```

Thanks for your comments. I learned various new cool functions that we can use in QuantumClifford!

test/test_ecc_steanereedmuller.jl
```diff
@@ -0,0 +1,96 @@
...
+    Hx, Hz = split(H)
+    gH = generate_gH(1, 3, Hx, Hz)
+
+    #Gottesman [[8, 3, 3]]
+    @test gH  == [1 1 1 1 1 1 1 1 0 0 0 0 0 0 0 0;
+                  0 0 0 0 0 0 0 0 1 1 1 1 1 1 1 1;
+                  0 0 0 0 1 1 1 1 0 0 1 1 0 0 1 1;
+                  0 0 1 1 0 0 1 1 0 1 0 1 0 1 0 1;
+                  0 1 0 1 0 1 0 1 0 0 1 1 1 1 0 0] 
+
```
Krastanov, line 68: I am confused by this check. Why is it so elaborate? Shouldn't it be simply:

```
for some_parameter in some_list
    @test parity_checks(SteaneReedMuller(...)) == parity_checks(Gottesman(...))
end
```

Krastanov: The hardcoded matrices you compare to are a bit dangerous. How can one know they are correctly copied from wherever they are copied?

test/test_ecc_steanereedmuller.jl
```diff
@@ -0,0 +1,96 @@
...
+        stab = parity_checks(SteaneReedMuller(t, r))
+        H = stab_to_gf2(stab)
+        Hx, Hz = split(H)
+        computed_distance = min_distance(Hx, Hz)
+        expected_distance = 2^t + 2^(t - 1)
+        @test computed_distance == expected_distance
+    end
+end
+
+function generate_gH(t, r, Hx, Hz)
```
Krastanov, line 36: What is this for? Seems very complicated.

src/ecc/codes/steanereedmuller.jl
```diff
@@ -0,0 +1,71 @@
...
+
+function _codesize_qrm(t, r)
+    return 2^r - sum(binomial.(r, 0:t)) 
+end 
+
+function _nplusk_qrm(t, r)
+    return (2^r + 2^r) - sum(binomial.(r, 0:t)) -  sum(binomial.(r, 0:t - 1)) 
+end 
+
+function _steane_convention_qrm(mat)
```
Krastanov, line 29: What is this for? Any reason to prefer one convention over the other? Generally, I would prefer simpler code, even if it leads to a slightly different convention.

If this is for testing purposes (e.g. to compare a code of different family with different conventions), then it should live in the test file.

test/test_ecc_steanereedmuller.jl
```diff
@@ -0,0 +1,96 @@
...
+    w = typemax(Int)
+    for (x, z) in zip(eachrow(Hx), eachrow(Hz))
+        r = map(|, x, z)
+        w = min(w, sum(r))
+    end
+    return w 
+end
+
+@testset "Test QRM(t, r) Matrix Minimum Distance" begin
+    for (t, r) in [(1, 3), (2, 4)]
```
Krastanov, line 26: Why are only these parameters tested?

src/ecc/codes/steanereedmuller.jl
```diff
@@ -0,0 +1,71 @@
...
+    end
+    e = (_codesize_qrm(t, r) - Dx_rows - t) == 0 || (_codesize_qrm(t, r) - Dx_rows - t) < 0 ? t : _codesize_qrm(t, r) - Dx_rows - t
+    Dz[m, :] = circshift(Dx[1, :], -e) 
+    
+    Hx = vcat(G1, zeros(Int64, size(G1, 1), size(G1, 2)), Dx)
+    Hz = vcat(zeros(Int64, size(G1, 1), size(G1, 2)), G1, Dz)
+    extended_Hx = Matrix{Bool}(Hx)
+    extended_Hz = Matrix{Bool}(Hz)
+    Stabilizer(extended_Hx, extended_Hz)
+end
```
Krastanov, line 71: A lot of file-ending issues again. Set your editor to take care of these. See this for explanation https://stackoverflow.com/questions/5813311/whats-the-significance-of-the-no-newline-at-end-of-file-log

test/test_ecc_steanereedmuller.jl
Krastanov: The reason I am worried about "elaborate" tests, is because if the test is not simple then you also need to worry about whether there are bugs in the test itself.

Krastanov (changes requested):
There are a couple of places where I do not follow what the code or tests are doing and a few places where simplifications seem to be necessary.

test/test_ecc_steanereedmuller.jl
Fe-r-oz: (1,3), and (2,4) are optimal codes. Their minimum distance is 3 and 6. And the test checks that, by checking the matrix H and by formula.

I have explained this in detail here: https://github.com/QuantumSavory/QuantumClifford.jl/pull/247#issuecomment-2019847986

Steane has explained that (2,5) is a less optimal code. I included the tests for optimal codes that these are the codes that obey the minimum distance bound which is 2^t + 2^(t-1) that is set in his paper.

Please have a look at the comment. I have thoroughly investigated and explained that.

src/ecc/codes/steanereedmuller.jl
Fe-r-oz: This is the standard convention that is used by Steane in his paper. If we don't use steane convention, people would be confused.  

By using this convention, it would help to maintain the code in the long run as it would be stabilizer can be readily verifiable from the Steane paper.  Please see this comment: 

I have explained this a bit further here: https://github.com/QuantumSavory/QuantumClifford.jl/pull/247#issuecomment-2022825082

test/test_ecc_steanereedmuller.jl
Fe-r-oz: The hardcoded matrix for j = 3 are taken from Steane's paper, **Table 15** where he represents the stabilizer for Gottesman [8,3,3]] code. 

Please https://github.com/QuantumSavory/QuantumClifford.jl/pull/247#issuecomment-2024103577

The reason that this test will not work is that the size of Parity Check Matrix of Steane-Reed-Muller Code is different from the size of the Gottesman codes for the corresponding values of j. 

> @test parity_checks(SteaneReedMuller(...)) == parity_checks(Gottesman(...))

Steane explains in Table 14, 15, that the Stabilizer (1,3) of Steane-Reed-Muller Code is the Gottesman Code for j =3. But they have different sizes.
This is because Steane has appended Dx and Dz matrices, and we can extract the Gottesman code for the corresponding  j  from the Stabilizer (1,3) of Steane-Reed-Muller Code. Please have a look at the paper as well. 

I have explained why canoncilize and stab_to_gf2 would help here because the Gottesman code has a different basis as we had to do row operations to get to the convention that is represented in Gottesman thesis . In steane's paper, he has used a different convention for representing Gottesman codes. And through that convention, he laids the concept that Gottesman codes are subset of when stabilizer has t=1, r=3, 4, 5 ..

test/test_ecc_steanereedmuller.jl
Fe-r-oz: Since the Stabilizer of Steane-Reed-Muller code for t=1,  r=3 has a size of 11 x 16 and the size of Gottesman code for j =3 is 5x8. They can't be compared even if we do canonicallize or stab_to_gf2. 

So, we need to extract the code from the Stabilizer of Steane-Reed-Muller code. We note that Dx and Dz are used to minimize the minimum distance from 2^t to 2^t-1 so they serve that purpose. When we focus on Hx and Hx of the Stabilizer of Steane-Reed-Muller code, we can see how we can extract the Gottesman code! This is a general extraction routine that will work for any value of r = 3, onwards given that t=1.

src/ecc/codes/steanereedmuller.jl
Fe-r-oz: I noticed these. Sorry, the file-ending issues will be removed. I didn't commit and push as that would have caused expensive tests.

test/test_ecc_steanereedmuller.jl
Fe-r-oz: Table 15 for Gottesman [[8,3,3]] is taken from Steane Paper: https://arxiv.org/pdf/quant-ph/9608026.pdf

test/test_ecc_steanereedmuller.jl
Fe-r-oz: Steane points out that the method he has described will result in less optimal codes, sub optimal codes along with the optimal codes. We can distinguish between the using the minimum distance test.

 The (1,3), (2,4) pass the min distance test because they are optimal codes, as and (2,5) is close to optimal, but as we increase the value of t and r, the codes becomes sub-optimal. This is also presented in the conclusion of his paper.

Krastanov: > Please have a look at the comment. I have thoroughly investigated and explained that.

These comments will not be there when someone is trying to understand the tests. It is important to give this context to future readers in a succinct and easy to understand fashion.

src/ecc/codes/steanereedmuller.jl
Krastanov: I disagree that this makes things easier. This makes things harder to understand and review.

Correctness checks should be based on automated tests and property checking, not on manual verification with something in a paper (on the other hand verifying with the paper is a good initial check during development). After all, how do you know the paper does not have a typo.

test/test_ecc_steanereedmuller.jl
Krastanov: > Steane explains in Table 14, 15, that the Stabilizer (1,3) of Steane-Reed-Muller Code is the Gottesman Code for j =3. But they have different sizes

Then compare the appropriate subsection of the code

> I have explained why canoncilize and stab_to_gf2 would help here because the Gottesman code has a different basis as we had to do row operations to get to the convention that is represented in Gottesman thesis . In steane's paper, he has used a different convention for representing Gottesman codes. And through that convention, he laids the concept that Gottesman codes are subset of when stabilizer has t=1, r=3, 4, 5

If there is some canonicalization necessary then `canonicalize!` and potentially reordering of the columns should be enough. This currently looks much too complicated to be trusted as a test.

test/test_ecc_steanereedmuller.jl
Krastanov: Why is all of this necessary if you just need to extract some subtableau from the SteaneReedMuller code? You can just write `some_tableau[1:n, :]` to extract the first `n` rows.

test/test_ecc_steanereedmuller.jl
Fe-r-oz: Please see this comment as well when the code size (k = 0 falls to zero) : https://github.com/QuantumSavory/QuantumClifford.jl/pull/247#issuecomment-2016707379.

Krastanov:
Generally, aim to have the code be selfcontained. As you are preparing the pull requests you are making a lot of comments and separate commits which generate a lot of emails. I generally do not read these as it is difficult to distinguish what is important and what is just an autogenerated message or an unimportant tangential topic. Have your code be selfcontained so it can be understood on its own down the line, when someone is reading it in the future without having access to a long informal conversation in github (which would be difficult to read anyway).

Fe-r-oz:
> Generally, aim to have the code be selfcontained. I generally do not read these as it is difficult to distinguish what is important and what is just an autogenerated message or an unimportant tangential topic.

Thanks for your comments. I actually included those comments as notes as I read the steane paper as there were many points that were to be noted.  And as I explored the paper more, it gathered up more notes. I believe it's better to have notes in a word document than github conversations

test/test_ecc_steanereedmuller.jl
Fe-r-oz: > You can just write some_tableau[1:n, :] to extract the first n rows.

>Then compare the appropriate subsection of the code

It's not just as simple extract as it would appear. Please see the the Stabilizer for Reed-Muller-Code for (1,3) and compare with the Gottesman Stabilizer that he has presented for [[8,3,3]], we can easily see that it's not just simply extracting rows. Does it seem that simple extraction `n` rows is work? 

 We have to carefully extract them, delete the Dx and Dz matrices, and then from the Gottesman rows. using canonicalize will not work in this case.  I have tested using canonicalize to see whether it works in this context or not before making the function.

And the function correctly outputs the tableaue that the Table 15 presents.

Krastanov:
You are free to keep notes however it is most convenient for you. Just make sure that things are available to the people that will be reading your code in the future, long after you have moved to other things (e.g. 10 or 20 years from now).

src/ecc/codes/steanereedmuller.jl
Fe-r-oz: I have included the minimum distance test as a correctness test. Tests only works work correct codes, not the less optimal codes, sub optimal codes or the codes for which the method fails. All of this is pointed out by steane. 

I included the steane convention just to resemble with the way of the paper. If you don't want to include the convention, it's okay. 

Not using `steane_convention_qrm ` would changes the order and result the Gottesman code extraction as Steane as shown the Gottesman Code in his Convention. Please see the table **15**. https://arxiv.org/pdf/quant-ph/9608026.pdf

 And we extract rows without using the convention would result in more complication if we canonicalize. I have tried using the canoniclize and I have presented its output  in the comment:

```
 1  0  0  1  0  1  1  0  0  1  0  1  0  1  0  1
 0  1  0  1  0  1  0  1  0  0  1  1  0  0  1  1
 0  0  1  1  0  0  1  1  0  1  0  1  1  0  1  0
 0  0  0  0  1  1  1  1  0  0  1  1  1  1  0  0
 0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1

julia> stab_to_gf2(parity_checks(Gottesman(3)))
5×16 Matrix{Bool}:
 1  1  1  1  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1
 1  0  1  0  0  1  0  1  0  0  0  0  1  1  1  1
 1  0  1  0  1  0  1  0  0  0  1  1  0  0  1  1
 1  0  0  1  0  1  1  0  0  1  0  1  0  1  0  1

```
This is how how steane represents Gottesman codes. Kindly please see the comment: https://github.com/QuantumSavory/QuantumClifford.jl/pull/247#issuecomment-2024103577

Fe-r-oz: Gottesman codes can be extracted carefully if they are presented in the notation Steane represents the Stabilizer for Reed-Muller Codes. 

Please see:

Screenshot from 2024-03-29 01-10-34

Fe-r-oz: ` generate_gH(t, r, Hx, Hz)` takes the value of `t,` and r, along with `Hx `and `Hz` and outputs the Gottesman Codes. There is a reason why the parameters are Hz and Hx and why we are not not passing H directly. 

Simply extracting rows of the Stabilizer Tableaue will not work. This can be seen if we see both the stabilizer of SRM code and corresponding Gottesman code.

Krastanov:
It seems the following might be one of the main source of miscommunication is the following:

This PR provides the following parity check matrix
image

And the paper gives this for the corresponding generator matrix
image

There seems to be a mismatch here. If I am reading the paper correctly, you have confused the generator matrix and the parity check matrix.

Fe-r-oz:
> If I am reading the paper correctly, you have confused the generator matrix and the parity check matrix.

Nope, I have not because if we go to the next pages, he represents them as Parity check matrices, not Generator matrices:

Please see SRM (2, 4) , SRM (2, 5)

Screenshot from 2024-03-29 01-28-35

Krastanov:
Whatever the paper is saying, your code is definitely not doing that. Check for instance the example I picked.

- Is `SteaneReedMuller(1,3)` supposed to be a valid code? What is its parity check matrix according to the paper?
- Why does `parity_checks(SteaneReedMuller(1,3))` has more rows than columns? A valid code should have fewer rows than columns.

Fe-r-oz:
> Whatever the paper is saying, your code is definitely not doing that. Check for instance the example I picked.

The code is definitely doing that. I can prove that pretty easily.

> * Is `SteaneReedMuller(1,3)` supposed to be a valid code? What is its parity check matrix according to the paper?
In the paper, Steane says the TOTAL ROWS are N +K. So, for t=1, r=3, the total rows must be 11

> * Why does `parity_checks(SteaneReedMuller(1,3))` has more rows than columns? A valid code should have fewer rows than columns.
Please see the answer above.

Fe-r-oz:
Please see:

Screenshot_select-area_20240329013736

Krastanov:
What your are talking about is the generator matrix, not the parity check matrix.

Please answer the following:

- Can the parity check matrix of a classical or quantum error correcting code have more rows than columns?
- Is that property fulfilled by the parity check matrices you have implemented here?

Fe-r-oz:
>>Why does parity_checks(SteaneReedMuller(1,3)) has more rows than columns?  

The stabilizer for 1,3 Does not have more rows than columns. Please see, the stabilizer does not have more rows than columns. 

```
ulia> stab_to_gf2(parity_checks(SteaneReedMuller(1, 3)))
11×16 Matrix{Bool}:
 1  1  1  1  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  0  0  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  1  1  0  0  1  1  0  0  0  0  0  0  0  0
 0  1  0  1  0  1  0  1  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1
 0  0  0  0  0  0  0  0  0  0  0  0  1  1  1  1
 0  0  0  0  0  0  0  0  0  0  1  1  0  0  1  1
 0  0  0  0  0  0  0  0  0  1  0  1  0  1  0  1
 0  0  0  0  0  0  1  1  0  0  0  0  0  1  0  1
 0  0  0  0  0  1  0  1  0  0  0  1  0  0  0  1
 0  0  0  1  0  0  0  1  0  0  0  0  0  1  1  0
```

Please explain to me when he says the **'Stabilizer of the  code',** what does he mean then? He has shown the method of how to construct the stabilizer for the Steane-Reed-Muller codes and I have followed the paper. 

The "stabilizer of the codes at the given (1,3), (2,4), (2,5) have following properties. 

Screenshot_select-area_20240329014525

Fe-r-oz:
Rows of the Parity check Matrix for` SteaneReedMuller `codes are not greater than its columns.

I have attached a few examples: The Gottesman (3) is also shown.
julia> stab_to_gf2(parity_checks(SteaneReedMuller(1, 3)))
```
11×16 Matrix{Bool}:  11 
 1  1  1  1  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  0  0  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  1  1  0  0  1  1  0  0  0  0  0  0  0  0
 0  1  0  1  0  1  0  1  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1
 0  0  0  0  0  0  0  0  0  0  0  0  1  1  1  1
 0  0  0  0  0  0  0  0  0  0  1  1  0  0  1  1
 0  0  0  0  0  0  0  0  0  1  0  1  0  1  0  1
 0  0  0  0  0  0  1  1  0  0  0  0  0  1  0  1
 0  0  0  0  0  1  0  1  0  0  0  1  0  0  0  1
 0  0  0  1  0  0  0  1  0  0  0  0  0  1  1  0

julia> stab_to_gf2(parity_checks(SteaneReedMuller(1, 4)))
14×32 Matrix{Bool}:        
 1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  1  1  1  1  0  0  0  0  1  1  1  1  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  1  1  0  0  1  1  0  0  1  1  0  0  1  1  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  1  0  1  0  1  0  1  0  1  0  1  0  1  0  1  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  1  1  1  1  0  0  0  0  1  1  1  1
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  1  1  0  0  1  1  0  0  1  1  0  0  1  1
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  1  0  1  0  1  0  1  0  1  0  1  0  1  0  1
 0  0  0  0  0  0  0  0  0  1  0  1  0  1  0  1  0  0  0  0  0  0  1  1  0  0  0  0  0  0  1  1
 0  0  0  0  0  0  1  1  0  0  0  0  0  0  1  1  0  0  0  0  0  1  0  1  0  0  0  0  0  1  0  1
 0  0  0  0  0  1  0  1  0  0  0  0  0  1  0  1  0  0  0  1  0  0  0  1  0  0  0  1  0  0  0  1

```

```
julia> stab_to_gf2(parity_checks(Gottesman(3)))

5×16 Matrix{Bool}:  
 1  1  1  1  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1
 1  0  1  0  0  1  0  1  0  0  0  0  1  1  1  1
 1  0  1  0  1  0  1  0  0  0  1  1  0  0  1  1
 1  0  0  1  0  1  1  0  0  1  0  1  0  1  0  1
```

src/ecc/codes/steanereedmuller.jl
Fe-r-oz: I have resolved the file ending locally. When the commt and push happens, these will be resolved.

Krastanov:
> The parity check matrix for 1,3 **Does not** have more rows than columns. Please see, the parity check matrix H does not have more rows than columns.

You are making a mistake in something pretty basic here. Please review some of the resources on error correcting codes before we continue with this conversation. 

Here is again the parity check tableau that your implementation is generating:

image

This has 8 columns, corresponding to 8 qubits. It has 11 rows, corresponding to 11 parity checks. If this is a code, it would have `n = 8` physical qubits and `k = 8-11 = -3` logical qubits, which obviously does not make sense.

I guess you are looking at the matrix in binary (X and Z part side by side), but then you need two columns per qubit - one for the X operations and one for the Z operations.

Krastanov:
Moreover, the rows of your parity check matrix do not commute (e.g. the two last rows). This is another sign that the code is definitely wrong. I am actually very surprised that the tests did not catch this... I thought it was being checked automatically.

Fe-r-oz:
I am very surprised. It seems that you have not reviewd the Steane paper fully. 

Everything that I said is from Steane Paper. The implementation I present is correct.

Kindly please review the paper, all the equations and the properties that he presents. It seems you have skimmed everything I said and are not wiliing to see that

```

julia> stab_to_gf2(parity_checks(SteaneReedMuller(1, 3)))
11×16 Matrix{Bool}:
 1  1  1  1  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  0  0  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  1  1  0  0  1  1  0  0  0  0  0  0  0  0
 0  1  0  1  0  1  0  1  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1
 0  0  0  0  0  0  0  0  0  0  0  0  1  1  1  1
 0  0  0  0  0  0  0  0  0  0  1  1  0  0  1  1
 0  0  0  0  0  0  0  0  0  1  0  1  0  1  0  1
 0  0  0  0  0  0  1  1  0  0  0  0  0  1  0  1
 0  0  0  0  0  1  0  1  0  0  0  1  0  0  0  1
 0  0  0  1  0  0  0  1  0  0  0  0  0  1  1  0

julia> parity_checks(SteaneReedMuller(1, 3))
+ XXXXXXXX
+ ____XXXX
+ __XX__XX
+ _X_X_X_X
+ ZZZZZZZZ
+ ____ZZZZ
+ __ZZ__ZZ
+ _Z_Z_Z_Z
+ _____ZXY
+ ___Z_X_Y
+ ___X_ZZX




```

Fe-r-oz:
> This has 8 columns, corresponding to 8 qubits. It has 11 rows, corresponding to 11 parity checks. If this is a code, it would have `n = 8` physical qubits and `k = 8-11 = -3` logical qubits, which obviously does not make sense.

I guess the miscommnication is the how the stabilizer is used in QuantumCLifford and how Steane is presenting them. I agree this will have error when we do circuit encoding because of the logic that you are using.

Krastanov:
I will have to close this for now. If you want to reopen the discussion please be explicit in explaining the following:

- Can an error correcting code have parity checks that do not commute?
- Can an error correcting code have more parity checks than number of physical qubits?
- What is the difference between a generator matrix and a parity check matrix?

Fe-r-oz:
> Can an error correcting code have parity checks that do not commute?

Commutativity property has to be satisfied. That's the reason why SteaneReedMuller (1,2) is not a quantum code.

>Can an error correcting code have more parity checks than number of physical qubits?
No.
 
What is the difference between a generator matrix and a parity check matrix?
Remember when you said,
> The tableaux is always "the parity check matrix. Agree?

So, this should give us H or G? 
```
julia> stab_to_gf2(parity_checks(Gottesman(3)))
5×16 Matrix{Bool}:  
 1  1  1  1  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1
 1  0  1  0  0  1  0  1  0  0  0  0  1  1  1  1
 1  0  1  0  1  0  1  0  0  0  1  1  0  0  1  1
 1  0  0  1  0  1  1  0  0  1  0  1  0  1  0  1

```

Similarly,
this gives us the parity check matrix H for SteaneMullerCode:


Do you disagree that this is not the parity check Matrix? 
```
julia> stab_to_gf2(parity_checks(SteaneReedMuller(1, 3)))
11×16 Matrix{Bool}:
 1  1  1  1  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  0  0  1  1  1  1  0  0  0  0  0  0  0  0
 0  0  1  1  0  0  1  1  0  0  0  0  0  0  0  0
 0  1  0  1  0  1  0  1  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  1  1  1  1  1  1  1  1
 0  0  0  0  0  0  0  0  0  0  0  0  1  1  1  1
 0  0  0  0  0  0  0  0  0  0  1  1  0  0  1  1
 0  0  0  0  0  0  0  0  0  1  0  1  0  1  0  1
 0  0  0  0  0  0  1  1  0  0  0  0  0  1  0  1
 0  0  0  0  0  1  0  1  0  0  0  1  0  0  0  1
 0  0  0  1  0  0  0  1  0  0  0  0  0  1  1  0
```
You asked me what's the parity check Matrix for parity_checks(SteaneReedMuller(1, 3))). You can see it above. 

SteaneReedMuller (1,3) 

n = 8
k = 3  (Please note that k = 3 based on the formula that is given in the steane paper)
Screenshot_select-area_20240329041746

d = 3
The problem is the inherent definition of the Stabilizer and how it's used in the encoding. 

I have read the Preskill notes, I know the differences. It seems that you have issue with Steane construction of the Paper. 
Screenshot_select-area_20240329042453

I have followed his approach, but according to you, I have not followed his appraoch.
> Whatever the paper is saying, your code is definitely not doing that

You already have a bias and mistrust that I am not following his paper even when I say many times that I am.

Fe-r-oz:
Also, I believe the miscommunication arose because of the differeing conventions and how Steane approach the paper. 

That could be the reason. I am not sure why this miscommunication arose

Fe-r-oz:
I am curious what should be the Stabilizer output of `SteaneReedMuller `is t =1, r = 3 or at any instance if we don't follow the Steane approach? As per your comments, SteaneReedMuller(1,3) is wrong Stabilizer that he presents it in the paper.

Fe-r-oz:
Dear Professor,

I would like to open this request. My apologies for causing inconvenience.

Krastanov:
Fe-r-oz, I do not think we should attempt to do a GSoC together, I would not be an effective mentor for you. You are extremely talented, but without you contemplating the possibility that you can be wrong about something, I would not be able to trust that you have thoroughly checked your work. Both in this PR and in the previous PR on Gottesman codes you made mistakes, I pointed out the mistakes by focusing on fundamental properties that the codes have to obey, i.e. checks that are so universal that they should be true no matter the code, and you doubled down on saying that the papers say whatever you are saying. You certainly have implemented something based on these papers, but the interpretation of what you have implemented is wrong, the bigger picture understanding is missing.

In this particular case, I have multiple times already said that you are producing generator matrices instead of producing parity check matrices. You have not shown understanding of the difference between these, which would have been fine on its own, but doubling down on your claims without checking is not fine. There is nothing wrong with Steane's approach, but the approach you are implementing simply does not produce the parity check matrix (it seems it is an intermediary step to producing the parity check matrix).

I want to make sure that you get credit for the things you have already implemented. Please let me know how can I list you as one of the contributors at https://eccbench.areweentangledyet.com/contributors/

Fe-r-oz:
Dear @Krastanov ,

Please let's continue conversation on slack.

Fe-r-oz:
Please open the PR so that we can finish it . Please note that The PR on isCSS and classical ReedMuller went very well and I also want this PR to end well. 

I believe the way Steane notation presented his paper is the biggest source of miscommunication. 

I request your to please commit your time this summer as your mentorship is much needed. I am really committed as I tried best as I opened back to back 4 PRs in one month to demonstrate my commitment. Of course, I did mistakes in two PRs,  and I am grateful for your time and commitment. 

Unfortunately, in this PR, there was miscommunication, and I am sorry for that. I hope to continue working on this PR. 

Please open this PR, I will incorporate your comments, and will focus on the universal properties when making tests. Please be rest assured . Suddenly closing the PR might not be the best route, please re-open so it can be finished. 

Please don't punish me this severely for the mistake. I will readily improve and not repeat the mistakes.

Fe-r-oz:
> (it seems it is an intermediary step to producing the parity check matrix).

That's why I asked you yesterday whether we can convert Generator Matrices to Parity Matrices in Quantum Clifford.

I believe the algorithm is creating the Stabilizer of the code, based on the Generator Matrix, and the Tables 14, 16, 18 are all Generator Matrices.  

We need to convert them to parity check matrices and then generate the correct Stabilizer?

Fe-r-oz:
My assumption was that we can construct the Stabilizer in a Generator Matrix Notation, and we can use stab_to_gf2 to get the parity check matrix. But as you pointed out that the Stabilizer Tableau is the parity check. Also, in the implementations,, since we were using extended_hx and extended hz, as in CSS, I thought that Stabilizer actually is constructed in the Generator Matrix notation, since the Stabilizer constructor takes two input (Hx, Hz).  That's was the reason why I assumed Generator Matrix is used to construct the Stabilizer, and we can get the parity check matrix via function. Since the design method was about separating the Matrices with X operators (Hx) and Z operators (Hz), I thought the stabilizer presented the generator matrix since for example XXXXXXXX => 1 1 1 1 1 1 1 1  0 0 0 0 0 0 0 0

Fe-r-oz:
Dear @Krastanov

Please see this as well 

He and Steane is calling them parity-check matrices, but you are phrasing them as Generator Matrics!?

Screenshot from 2024-03-29 15-19-34

If we follow the above definition, Steane's is presenting parity-check matrics not the Generator Matrices!

> And the paper gives this for the corresponding generator matrix

Paper is talking about Parity check matrices, not the Generator Matrices!
Screenshot from 2024-03-29 15-21-23

Kindly Please let me now your comments.

Krastanov:
I am locking this after a private conversation with Feroz

PR #252, opened 2024-04-02
Fixing errors with generation of uncycle codes

Benzillaist:
A few small changes were made that will improve the .ECC module:
- Fixed errors with the generation of Unicycle codes:
- Added a few tests to check the validity of bicycle and unicycle codes
- Added docstrings to bicycle and unicycle codes that include examples of how to use the codes


Krastanov:
It seems this PR has lost steam. Seems pretty close to being ready for merging though. I will close it for now, but if you want to pick it back up just reopen it and we can continue the discussion.

PR #253, opened 2024-04-02
configure two gate noise in various code evaluation setups

ismoldayev:
Implemented add_two_qubit_gate_noise for NaiveSyndromeECCSetup.

### DIFF
--- a/src/ecc/decoder_pipeline.jl
+++ b/src/ecc/decoder_pipeline.jl
@@ -108,8 +108,8 @@ end
 
 function physical_ECC_circuit(H, setup::ShorSyndromeECCSetup)
     prep_anc, syndrome_circ, n_anc, syndrome_bits = shor_syndrome_circuit(H)
-    noisy_syndrome_circ = []
 
+    noisy_syndrome_circ = []
     for op in syndrome_circ
         push!(noisy_syndrome_circ, op)
         for noise_op in add_two_qubit_gate_noise(op, setup.two_qubit_gate_noise)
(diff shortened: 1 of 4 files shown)


src/ecc/decoder_pipeline.jl
```diff
@@ -81,12 +81,27 @@ end
 
 function physical_ECC_circuit(H, setup::NaiveSyndromeECCSetup)
     syndrome_circ, n_anc, syndrome_bits = naive_syndrome_circuit(H)
-    noisy_syndrome_circ = syndrome_circ # add_two_qubit_gate_noise(syndrome_circ, gate_error)
-    mem_error_circ = [PauliError(i, setup.mem_noise) for i in 1:nqubits(H)];
+    function add_two_qubit_gate_noise(circuit, gate_error)
```
Krastanov, line 84: We can restructure this in a bit more idiomatic, multiple dispatch way.

I would suggest doing something more like

```
function add_two_qubit_noise(g, gate_error)
    return ()
end
function add_two_qubit_noise(g::AbstractTwoQubitOperator, gate_error)
    return (PauliError(affectedqubits(op), gate_error), )
end

function physical_ECC_circuit...
    ...
    for op in circuit
        push!(...)
        for noiseop in add_two_qubit_gate_noise(op, gate_error)
            push!(...)
        end
    end
end
```

The advantage of this change is that now folks can simply define their own types and their own `add_two_qubit_gate_noise` without having to modify `physical_ECC_circuit`. We provided a reusable abstract interface.

Krastanov (changes requested):
Thanks for the submission, this is a good start! I am suggesting some restructuring and then we can merge it.

Krastanov:
I will mark this is a draft. When you have addressed the change and you have passing tests, re-request a review from me and make it not a draft anymore.

Ping me if you have questions or if something is unclear.

PR #256, opened 2024-04-09
Convolutional code

Fe-r-oz:
Classical Convolutional Code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 
 # News
 
+## v0.9.2 - 2024-04-09
+
+- Added the classical convolutional code to the ECC module
+
 ## v0.9.2 - 2024-04-02
 
 - Provide more configuration options in the `PyBeliefProp` decoders.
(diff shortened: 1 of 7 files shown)


Fe-r-oz:
Classical Convolutional codes are fully implemented using Nemo.  I have followed the Field Theoretic Implementation that was described in a paper, so that way, they can be described via Abstract Algebra.

TODO: look for properties of the convolutional codes that can be tested via FieldPolynomialRings.

Fe-r-oz:
TODO: I am exploring about Direct Summands and how to check whether a Free Module is a Direct Summand given the Finite Field.

Fe-r-oz:
Progress Update:

I had correspondence with authors of the paper of the code itself.  I read the book and there is no mention of "free modules" at all, but they defined the it in terms of modules.

They have provided a definition in terms of Number theory that differs from what is provided in the book so I have requested further details.

Fe-r-oz:
I have opened an feature request at `https://github.com/esabo/CodingTheory/issues/33` for these codes in `CodingTheory.jl `to monitor the progress on these codes.

PR #263, opened 2024-04-19
Adding classical Bose–Chaudhuri–Hocquenghem code to ECC module

Fe-r-oz:
- Adding classical Bose–Chaudhuri–Hocquenghem code to ECC module


Fe-r-oz:
Thanks for your comments. I have significantly improved the doc-strings and made 1 commit/push.

I hope you like the descent from Finite Fields back to the Boolean Field when reviewing!

Fe-r-oz:
Results:

```
Test Summary:          | Time
ecc_decoder_all_setups | None  27.7s
Test Summary: | Pass  Total  Time
ecc_encoding  |  238    238  0.9s
Test Summary: |  Pass  Total  Time
ecc_gottesman | 24562  24562  0.2s
Test Summary:  | Pass  Total  Time
ecc_reedmuller |   19     19  0.8s
Test Summary: | Pass  Total  Time
ecc_bch       |   36     36  2.7s
Test Summary: | Pass  Total  Time
ecc_syndromes |  176    176  3.5s
Test Summary: | Pass  Total  Time
ecc_throws    |    4      4  0.0s
Test Summary: |Time
precompile    | None  0.3s
┌ Warning: The operation is of a type that can not be unified, defaulting to slower runtime dispatch
│   typeof(g) = NoiseOp{UnbiasedUncorrelatedNoise{Float64}, 1}
└ @ QuantumClifford ~/Desktop/New/bch/QuantumClifford.jl/src/sumtypes.jl:77
┌ Warning: The operation is of a type that can not be unified, defaulting to slower runtime dispatch
│   typeof(g) = NoiseOp{UnbiasedUncorrelatedNoise{Float64}, 1}
└ @ QuantumClifford ~/Desktop/New/bch/QuantumClifford.jl/src/sumtypes.jl:77
Test Summary: | Pass  Total  Time
pauliframe    |   93     93  6.0s
Test Summary: | Pass  Total  Time
allocations   |   39     39  7.4s
[ Info: SetupBuildDirectory: setting up build directory.
[ Info: Doctest: running doctests.
[ Info: Skipped ExpandTemplates step (doctest only).
[ Info: Skipped CrossReferences step (doctest only).
[ Info: Skipped CheckDocument step (doctest only).
[ Info: Skipped Populate step (doctest only).
[ Info: Skipped RenderDocument step (doctest only).
Test Summary: | Pass  Total   Time
doctests      |    1      1  14.6s
Skipping Base.active_repl
Skipping Base.active_repl_backend
Test Summary: | Pass  Total   Time
aqua          |   11     11  19.0s
     Testing QuantumClifford tests passed 
```

Fe-r-oz:
@Krastanov, Could you please help review the open PRs? Thanks.

Fe-r-oz:
BCH code (15, 2)

Parity Check Matrix: H 
```
julia> parity_checks(BCH(15,2))
8×15 Matrix{Bool}:
 1  0  0  0  1  0  0  1  1  0  1  0  1  1  1
 0  1  0  0  1  1  0  1  0  1  1  1  1  0  0
 0  0  1  0  0  1  1  0  1  0  1  1  1  1  0
 0  0  0  1  0  0  1  1  0  1  0  1  1  1  1
 1  0  0  0  1  1  0  0  0  1  1  0  0  0  1
 0  0  0  1  1  0  0  0  1  1  0  0  0  1  1
 0  0  1  0  1  0  0  1  0  1  0  0  1  0  1
 0  1  1  1  1  0  1  1  1  1  0  1  1  1  1
```
Screenshot_select-area_20240430050759

HField Matrix:
```

2×15 Matrix{FqFieldElem}:
 1  a    a^2        a^3      a + 1              a^2 + a  a^3 + a^2  …  a^3 + a^2 + 1  a^3 + 1
 1  a^3  a^3 + a^2  a^3 + a  a^3 + a^2 + a + 1  1        a^3           a^3 + a        a^3 + a^2 + a + 1
```
Screenshot_select-area_20240430050859


julia> GF2, a = finite_field(2, 4, "a")
(Finite field of degree 4 and characteristic 2, a)


```

julia> a^42
a^3 + a^2 + a + 1

julia> a^39
a^3 + a

julia> a^12
a^3 + a^2 + a + 1

julia> a^9
a^3 + a

julia> a^14
a^3 + 1

julia> a^13
a^3 + a^2 + 1

```

src/ecc/codes/classical/bch.jl
```diff
@@ -0,0 +1,93 @@
...
+
+The ECC Zoo has an [entry for this family](https://errorcorrectionzoo.org/c/q-ary_bch)
+"""
+struct BCH <: ClassicalCode
+    n::Int 
+    t::Int #Error correction capability; t bits can be corrected
+
+    function BCH(n, t)
+        if n < 7 || n > 500 || t < 0 || t > 2^(ceil(Int, log2(n + 1)) - 1)
+            throw(ArgumentError("Invalid parameters: 'n' and 't' must be positive, and 'r' must be >= to 3. Additionally, 'n' is >= to 7 since n = 2ͬ - 1 and 't' < 2^(r - 1), to obtain a valid code and to tractable."))
```
Krastanov, line 25: what is `r`? It does nto seem to be one of the parameters

Krastanov: More generally, this error message is very confusing. Maybe split it in a few separate checks, each with its own error message and explanation.

test/test_ecc_bch.jl
```diff
@@ -0,0 +1,39 @@
+using Test
+using Nemo: ZZ, residue_ring, matrix, finite_field, GF, minpoly, coeff, lcm, FqPolyRingElem, FqFieldElem, is_zero
+using QuantumClifford
+using QuantumClifford.ECC
+using QuantumClifford.ECC: AbstractECC, BCH
+
+@testset "Test BCH(n, t) generator polynomial g(x) universal property: mod(x^n - 1, g(x)) == 0" begin
+    n_cases = [7, 15, 31, 63, 127, 255]
```
Krastanov, line 8: Where do these come from? Why only them?

Krastanov (changes requested):
Thanks for setting this up! I finished a first round of a review. There are some superficial comments, but also a bunch of fairly important comments related to testing and error messages.

I also tried to run a few examples, and I am getting some strange results for some parameter regimes. E.g. the following:

```
julia> parity_checks(BCH(7,4)) |> size
(12, 7)
```

This does not make sense to me. A parity check matrices should have fewer rows than columns. This should probably be explicitly part of some form of universal tests -- the rank of the matrix should be equal to the number of rows it has which should be smaller than the number of columns, for any code.

Krastanov:
> BCH code (15, 2)
> 
> Parity Check Matrix: H
> 
> ```
> julia> parity_checks(BCH(15,2))
> 8×15 Matrix{Bool}:
>  1  0  0  0  1  0  0  1  1  0  1  0  1  1  1
>  0  1  0  0  1  1  0  1  0  1  1  1  1  0  0
>  0  0  1  0  0  1  1  0  1  0  1  1  1  1  0
>  0  0  0  1  0  0  1  1  0  1  0  1  1  1  1
>  1  0  0  0  1  1  0  0  0  1  1  0  0  0  1
>  0  0  0  1  1  0  0  0  1  1  0  0  0  1  1
>  0  0  1  0  1  0  0  1  0  1  0  0  1  0  1
>  0  1  1  1  1  0  1  1  1  1  0  1  1  1  1
> ```
> 
> Screenshot_select-area_20240430050759
> 
> HField Matrix:
> 
> ```
> 
> 2×15 Matrix{FqFieldElem}:
>  1  a    a^2        a^3      a + 1              a^2 + a  a^3 + a^2  …  a^3 + a^2 + 1  a^3 + 1
>  1  a^3  a^3 + a^2  a^3 + a  a^3 + a^2 + a + 1  1        a^3           a^3 + a        a^3 + a^2 + a + 1
> ```
> 
> Screenshot_select-area_20240430050859
> 
> julia> GF2, a = finite_field(2, 4, "a") (Finite field of degree 4 and characteristic 2, a)
> 
> ```
> 
> julia> a^42
> a^3 + a^2 + a + 1
> 
> julia> a^39
> a^3 + a
> 
> julia> a^12
> a^3 + a^2 + a + 1
> 
> julia> a^9
> a^3 + a
> 
> julia> a^14
> a^3 + 1
> 
> julia> a^13
> a^3 + a^2 + 1
> ```

This should be part of the tests, not just a comment on the pull requests (it is one of the todos I have listed in the review). Make sure these tests have references to where the example is taken from.

Fe-r-oz:
> I also tried to run a few examples, and I am getting some strange results for some parameter regimes. E.g. the following:
> 
> ```
> julia> parity_checks(BCH(7,4)) |> size
> (12, 7)
> ```
> 
> This does not make sense to me. A parity check matrices should have fewer rows than columns. 

I made a mistake in the setting the bound of t. After reading the slides, the  m >= 3 and t < 2^(m -1 ).  n = Block length: n = 2^m − 1.

When m =3, we have n=7, and GF(2^3).

When n =7 is actually 2^3 - 1, GF(8), so m = 3. Now, t must be < 2^(3 - 1) so t < 4. (7, 4) is not a valid instance of the code and the bound is set incorrectly.  I have set it to t <= 2^(m - 1) so that why the invalid instance (7, 4) is not throwing the error.

So, the m =3, the valid codes are (7, 1),  (7, 2)  and (7, 3) as t should be less than 2^(m - 1) as per the formula!
```
Right: 1 -> m- tuple
julia> parity_checks(BCH(7,1))
FqFieldElem[1 a a^2 a + 1 a^2 + a a^2 + a + 1 a^2 + 1]
3×7 Matrix{Bool}:
 1  0  0  1  0  1  1
 0  1  0  1  1  1  0
 0  0  1  0  1  1  1

Right:  2 -> 2 m- tuples
julia> parity_checks(BCH(7,2))
FqFieldElem[1 a a^2 a + 1 a^2 + a a^2 + a + 1 a^2 + 1; 1 a + 1 a^2 + 1 a^2 a^2 + a + 1 a a^2 + a]
6×7 Matrix{Bool}:
 1  0  0  1  0  1  1
 0  1  0  1  1  1  0
 0  0  1  0  1  1  1
 1  1  1  0  1  0  0
 0  1  0  0  1  1  1
 0  0  1  1  1  0  1

Right: Right:  3 -> 3 m- tuples
julia> parity_checks(BCH(7,3))
FqFieldElem[1 a a^2 a + 1 a^2 + a a^2 + a + 1 a^2 + 1; 1 a + 1 a^2 + 1 a^2 a^2 + a + 1 a a^2 + a; 1 a^2 + a + 1 a + 1 a a^2 + 1 a^2 + a a^2]
9×7 Matrix{Bool}:
 1  0  0  1  0  1  1
 0  1  0  1  1  1  0
 0  0  1  0  1  1  1
 1  1  1  0  1  0  0
 0  1  0  0  1  1  1
 0  0  1  1  1  0  1
 1  1  1  0  1  0  0
 0  1  1  1  0  1  0
 0  1  0  0  1  1  1

Wrong, invalid code
julia> parity_checks(BCH(7,4))
FqFieldElem[1 a a^2 a + 1 a^2 + a a^2 + a + 1 a^2 + 1; 1 a + 1 a^2 + 1 a^2 a^2 + a + 1 a a^2 + a; 1 a^2 + a + 1 a + 1 a a^2 + 1 a^2 + a a^2; 1 1 1 1 1 1 1]
12×7 Matrix{Bool}:
 1  0  0  1  0  1  1
 0  1  0  1  1  1  0
 0  0  1  0  1  1  1
 1  1  1  0  1  0  0
 0  1  0  0  1  1  1
 0  0  1  1  1  0  1
 1  1  1  0  1  0  0
 0  1  1  1  0  1  0
 0  1  0  0  1  1  1
 1  1  1  1  1  1  1
 0  0  0  0  0  0  0
 0  0  0  0  0  0  0
```

test/test_ecc_bch.jl
Fe-r-oz: Block length: n = 2^m − 1, hence testing all examples of the BCH code in this way, Since we can vary t-bits.

```
m = 3, n = 7
m = 4, n = 15
m = 5, n = 31
m = 6, n = 63
m = 7, n = 127
m = 8, n = 255
```

src/ecc/codes/classical/bch.jl
Fe-r-oz: r is the same as m which is the degree of the finite field. Alright

test/test_ecc_bch.jl
Krastanov: Noted. If I understand your constructor `BCH(...)` takes `n`, but only some `n` are valid. I think it would be better to take the underlying parameter `m` as input for the constructor. I get that `n` - the block size - is a number of significant importance, but here that number is derived, it is not the starting point. A user can always do `code_n(BCH(...))` to figure out what the size is.

Consider the toric code for example. In its constructor we use the latice size, not the number of qubits.

src/ecc/codes/classical/bch.jl
```diff
@@ -0,0 +1,93 @@
...
+.                  .                         .                         .                                           .
+1            (α ^ (2*t - 1)) ^ 1       (α ^ (2*t - 1)) ^ 2       (α ^ (2*t - 1)) ^ 3     ...              (α ^ (2*t - 1)) ^ (n - 1) 
+
+Note: The entries of matrix over field elements are in GF(2^m). Each element in GF(2^m) can be represented by a m-tuple/binary column of length m over GF(2). If each entry of H is replaced by its corresponding m-tuple/binary column of length m over GF(2) arranged in column form, we obtain a binary parity check matrix for the code.
+
+You might be interested in consulting [bose1960further](@cite) as well.
+
+The ECC Zoo has an [entry for this family](https://errorcorrectionzoo.org/c/q-ary_bch)
+"""
+struct BCH <: ClassicalCode
```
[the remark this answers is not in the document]
Fe-r-oz: Definitely.

test/test_ecc_bch.jl
Fe-r-oz: I see. That makes better sense.

src/ecc/codes/classical/bch.jl
```diff
@@ -0,0 +1,93 @@
...
+2. Error Correction: A BCH code with generator polynomial `g(x)` can correct up to `t` errors in a codeword of length `2^m - 1`.
+3. Minimal Polynomials: `g(x)` is the least common multiple (LCM) of the minimal polynomials `φ_i(x)` of `α^i` for `i = 1` to `2^t`.
+
+Minimal Polynomial:
+
+- The minimal polynomial of a field element `α` in GF(2^m) is the polynomial of the lowest degree over GF(2) that has `α` as a root. It represents the simplest polynomial relationship between `α` and the elements of GF(2).
+
+Least Common Multiple (LCM):
+
+- The LCM of two or more polynomials `f_i(x)` is the polynomial with the lowest degree that is a multiple of all `f_i(x)`. It ensures that `g(x)` has all the roots of `φ_i(x)` for `i = 1` to `2^t`.
```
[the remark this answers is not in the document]
Fe-r-oz: Finding Minimal Polynomial and LCM are specific to BCH generator polynomial.  Every distinct polynomial code has a unique way of describing its generator polynomial, gx.

test/test_ecc_bch.jl
```diff
@@ -0,0 +1,39 @@
+using Test
+using Nemo: ZZ, residue_ring, matrix, finite_field, GF, minpoly, coeff, lcm, FqPolyRingElem, FqFieldElem, is_zero
+using QuantumClifford
+using QuantumClifford.ECC
+using QuantumClifford.ECC: AbstractECC, BCH
```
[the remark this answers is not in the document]
Fe-r-oz: I checked that n - k is == (degree of the generator polynomial, gx) == rank.  So using one line of code, we can check the rank, self consistency of H as well correctness of gx like:

`@test rank(parity_checks(BCH(m, t))) == degree(generator_polynomial(BCH(m, t))`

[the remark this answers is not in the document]
Krastanov: be careful with this. I think you are not checking the binary rank, because the default `rank` converts to floating point.

Consider this case where the rank should be 2:
```
julia> rank([true true false; true false true; false true true])
3
```

You should check rank by first making a GF(2) matrix, e.g. like here https://github.com/QuantumSavory/QuantumClifford.jl/blob/d6018f8b3531119ff81cbe002e9256999e452c58/src/entanglement.jl#L204-L205

Fe-r-oz:
Thanks for comprehensive review. I have incorporated all your suggestions.

```
Test Summary:          | Time
ecc_decoder_all_setups | None  26.6s
Test Summary: | Pass  Total  Time
ecc_encoding  |  238    238  0.6s
Test Summary: |  Pass  Total  Time
ecc_gottesman | 24562  24562  0.2s
Test Summary:  | Pass  Total  Time
ecc_reedmuller |   19     19  0.8s
Test Summary: | Pass  Total  Time
ecc_bch       |   51     51  2.5s
Test Summary: | Pass  Total  Time
ecc_syndromes |  176    176  3.2s
Test Summary: | Pass  Total  Time
ecc_throws    |    4      4  0.0s
Test Summary: |Time
precompile    | None  0.2s
Test Summary: | Pass  Total  Time
pauliframe    |  106    106  5.5s
Test Summary: | Pass  Total  Time
allocations   |   39     39  6.3s

Test Summary: | Pass  Total   Time
doctests      |    1      1  14.3s
Skipping Base.active_repl
Skipping Base.active_repl_backend
Test Summary: | Pass  Total   Time
aqua          |   11     11  19.4s
     Testing QuantumClifford tests passed 

```

test/test_ecc_bch.jl
```diff
@@ -0,0 +1,55 @@
...
+                                           1  0  0  0  1  1  0  0  0  1  1  0  0  0  1;
+                                           0  0  0  1  1  0  0  0  1  1  0  0  0  1  1;
+                                           0  0  1  0  1  0  0  1  0  1  0  0  1  0  1;
+                                           0  1  1  1  1  0  1  1  1  1  0  1  1  1  1]
+
+    GF2ͬ, a = finite_field(2, 4, "a")
+    GF2x, x = GF(2)["x"]
+    #examples taken from [https://web.ntpu.edu.tw/~yshan/BCH_code.pdf]   
+    @test generator_polynomial(BCH(4, 2)) == x ^ 8 + x ^ 7 +  x ^ 6 +  x ^ 4 + 1
+    @test generator_polynomial(BCH(4, 3)) == x ^ 10 + x ^ 8+  x ^ 5 +  x ^ 4 +  x ^ 2 + x + 1
```
Fe-r-oz, line 53: Please add a space after + here. Don't want to push as it will create another expensive test run.

src/ecc/codes/classical/bch.jl
```diff
@@ -0,0 +1,103 @@
...
+    end
+end
+
+"""
+Generator Polynomial of BCH Codes
+
+This function calculates the generator polynomial `g(x)` of a t-bit error-correcting BCH code of length `(n)` of `2ᵐ - 1` over the finite Galois field GF(2).
+
+generator_polynomia(BCH(m, t)):
+- `m` (Integer): The positive integer defining the code length (`2ᵐ - 1`).
```
Fe-r-oz, line 49: Tiny correction for line 49:
- - `m` (Integer): The positive integer defining the degree of the finite (Galois) field, GF(2ᵐ)

test/test_ecc_bch.jl
```diff
@@ -0,0 +1,55 @@
...
+        end
+        # BCH code is cyclic as its generator polynomial, gx divides xⁿ - 1, so mod (xⁿ - 1, gx) = 0 
+        for t in 1:2
+            gx = generator_polynomial(BCH(m, t))
+            GF2x, x = GF(2)["x"] 
+            @test mod(x^n - 1, gx) == 0
+        end
+    end
+
+    #example taken from [https://web.ntpu.edu.tw/~yshan/BCH_code.pdf]
```
Fe-r-oz, line 39: Please use this link at line 39 during cleanup:
 https://personal.oss.unist.hr/~mnizetic/ZASTITNO%20LINIJSKI%20KODIRANJE/SEMINARSKI%20RADOVI/CH06.pdf

instead of this: https://web.ntpu.edu.tw/~yshan/BCH_code.pdf

Krastanov: I am happy to review, but these are things you should finish if you want this merged. I am grateful for the contribution, but I can not spend time on finishing tasks that you already know need to be done.

Fe-r-oz: Noted. Also, I am grateful for your review.

Krastanov:
I see you have requested a review but then you also made quite a few changes after the request. It is not uncommon for this to happen as you think of some other things to improve. I usually try to wait a day if I am unsure I whether I will be making more changes, and request the review only after that day has passed.

I am converting this back to a draft so that you have a way to signal when you are done. Convert it to non-draft and request a review when you think it is ready.

src/ecc/codes/classical/bch.jl
```diff
@@ -0,0 +1,105 @@
...
+"""
+BCH(m, t):
+- `m` (Integer): The positive integer defining the degree of the finite (Galois) field, GF(2ᵐ).
+- `t` (Integer): The positive integer specifying the number of correctable errors (`t`).
+"""
+struct BCH <: AbstractPolynomialCode
+    m::Int 
+    t::Int 
+    function BCH(m, t)
+        if m < 3 || m > 10 || t < 0 || t >= 2^(m - 1)
```
Krastanov, line 35: I would suggest removing the `m>10` check and correspondingly update the error message. Code with 100 000 physical bits is unreasonable, but this seems to limit it to 1000-ish.

Fe-r-oz:
> I am converting this back to a draft so that you have a way to signal when you are done. Convert it to non-draft and request a review when you think it is ready.

Yeah, waiting a day sounds better. A lot of "I could also do this or that" signals are popping" so that's why a quite a few stylistic changes are being done today. I will wait a day to make sure the cosmic background disappear!

src/ecc/codes/classical/bch.jl
```diff
@@ -0,0 +1,105 @@
...
+
+The ECC Zoo has an [entry for this family](https://errorcorrectionzoo.org/c/q-ary_bch).
+"""
+
+abstract type AbstractPolynomialCode <: ClassicalCode end
+
+"""
+BCH(m, t):
+- `m` (Integer): The positive integer defining the degree of the finite (Galois) field, GF(2ᵐ).
+- `t` (Integer): The positive integer specifying the number of correctable errors (`t`).
```
Krastanov, line 29:
```suggestion
`BCH(m, t)`
- `m`: The positive integer defining the degree of the finite (Galois) field, GF(2ᵐ).
- `t`: The positive integer specifying the number of correctable errors.
```

Removing repetition and ensuring code is formatted

src/ecc/codes/classical/bch.jl
```diff
@@ -0,0 +1,105 @@
...
+"""
+struct BCH <: AbstractPolynomialCode
+    m::Int 
+    t::Int 
+    function BCH(m, t)
+        if m < 3 || t < 0 || t >= 2 ^ (m - 1)
+            throw(ArgumentError("Invalid parameters: 'm' and 't' must be positive. Additionally, 'm' ≥ 3 and 't' < 2ᵐ ⁻ ¹ to obtain a valid code and to tractable."))
```
Krastanov, line 36:
```suggestion
            throw(ArgumentError("Invalid parameters: `m` and `t` must be positive. Additionally, ensure `m ≥ 3` and `t < 2ᵐ ⁻ ¹` to obtain a valid code."))
```

formatting and grammar

src/ecc/codes/classical/bch.jl
```diff
@@ -0,0 +1,105 @@
...
+        end
+        new(m, t)
+    end
+end
+
+"""
+Generator Polynomial of BCH Codes
+
+This function calculates the generator polynomial `g(x)` of a `t`-bit error-correcting BCH code of length `(n)` of `2ᵐ - 1` over the finite Galois field GF(2).
```
Krastanov, line 45:
```suggestion
This function calculates the generator polynomial `g(x)` of a `t`-bit error-correcting BCH code of binary length `n = 2ᵐ - 1`. The binary code is derived from a code over the finite Galois field GF(2).
```

formatting and disambiguation about what is binary and what is not

src/ecc/codes/classical/bch.jl
```diff
@@ -0,0 +1,105 @@
...
+
+generator_polynomial(BCH(m, t)):
+- `m` (Integer): The positive integer defining the degree of the finite (Galois) field, GF(2ᵐ).
+- `t` (Integer): The positive integer specifying the number of correctable errors (`t`).
```
Krastanov, line 49:
```suggestion
`generator_polynomial(BCH(m, t))`

- `m`: The positive integer defining the degree of the finite (Galois) field, GF(2ᵐ).
- `t`: The positive integer specifying the number of correctable errors.
```

code formatting and removing repetition

src/ecc/codes/classical/bch.jl
```diff
@@ -0,0 +1,105 @@
...
+
+Description:
+
+The generator polynomial `g(x)` is the fundamental polynomial used for encoding and decoding BCH codes. It has the following properties:
+
+1. Roots: It has `α`, `α²`, `α³`, ..., `α²ᵗ` as its roots, where `α` is a primitive element of the Galois Field GF(2ᵐ).
+2. Error Correction: A BCH code with generator polynomial `g(x)` can correct up to `t` errors in a codeword of length `2ᵐ - 1`.
+3. Minimal Polynomials: `g(x)` is the least common multiple (LCM) of the minimal polynomials `φᵢ(x)` of `αⁱ` for `i = 1` to `2ᵗ`.
```
Krastanov, line 57:
```suggestion
3. Minimal Polynomials: `g(x)` is the least common multiple (LCM) of the minimal polynomials `φᵢ(x)` of `αⁱ` for `i` from `1` to `2ᵗ`.
```

src/ecc/codes/classical/bch.jl
```diff
@@ -0,0 +1,105 @@
...
+
+Minimal Polynomial:
+
+- The minimal polynomial of a field element `α` in GF(2ᵐ) is the polynomial of the lowest degree over GF(2) that has `α` as a root. It represents the simplest polynomial relationship between `α` and the elements of GF(2).
```
Krastanov, line 61: I do not think I understand what this sentence means: `It represents the simplest polynomial relationship between `α` and the elements of GF(2).`

Could you clarify it or remove it?

test/test_ecc_bch.jl
```diff
@@ -0,0 +1,51 @@
...
+            @test designed_distance(H) >= 2 * t
+            # n - k == degree of generator polynomial, `g(x)` == rank of binary parity check matrix, `H`.
+            mat = matrix(GF(2), parity_checks(BCH(m, t)))
+            computed_rank = rank(mat)
+            @test computed_rank == degree(generator_polynomial(BCH(m, t)))
+            @test code_k(BCH(m, t)) == n - degree(generator_polynomial(BCH(m, t)))
+            # BCH code is cyclic as its generator polynomial, `g(x)` divides `xⁿ - 1`, so `mod (xⁿ - 1, g(x))` = 0.
+            gx = generator_polynomial(BCH(m, t))
+            GF2x, x = GF(2)["x"] 
+            @test mod(x ^ n - 1, gx) == 0
```
Krastanov, line 34: really great tests!

test/test_ecc_bch.jl
```diff
@@ -0,0 +1,51 @@
...
+        end
+    end
+
+    #example taken from [https://personal.oss.unist.hr/~mnizetic/ZASTITNO%20LINIJSKI%20KODIRANJE/SEMINARSKI%20RADOVI/CH06.pdf].
```
Krastanov, line 38:
```suggestion
    #example taken from https://personal.oss.unist.hr/~mnizetic/ZASTITNO%20LINIJSKI%20KODIRANJE/SEMINARSKI%20RADOVI/CH06.pdf
```

test/test_ecc_bch.jl
```diff
@@ -0,0 +1,51 @@
...
+    @test parity_checks(BCH(4, 2))  ==    [1  0  0  0  1  0  0  1  1  0  1  0  1  1  1;
+                                           0  1  0  0  1  1  0  1  0  1  1  1  1  0  0;
+                                           0  0  1  0  0  1  1  0  1  0  1  1  1  1  0;
+                                           0  0  0  1  0  0  1  1  0  1  0  1  1  1  1;
+                                           1  0  0  0  1  1  0  0  0  1  1  0  0  0  1;
+                                           0  0  0  1  1  0  0  0  1  1  0  0  0  1  1;
+                                           0  0  1  0  1  0  0  1  0  1  0  0  1  0  1;
+                                           0  1  1  1  1  0  1  1  1  1  0  1  1  1  1]
+    GF2x, x = GF(2)["x"]
+    #examples taken from [https://web.ntpu.edu.tw/~yshan/BCH_code.pdf].
```
Krastanov, line 48:
```suggestion
    #examples taken from https://web.ntpu.edu.tw/~yshan/BCH_code.pdf
```

test/test_ecc_bch.jl
```diff
@@ -0,0 +1,51 @@
...
+    @test generator_polynomial(BCH(4, 2)) == x ^ 8 + x ^ 7 +  x ^ 6 +  x ^ 4 + 1
+    @test generator_polynomial(BCH(4, 3)) == x ^ 10 + x ^ 8 +  x ^ 5 +  x ^ 4 +  x ^ 2 + x + 1
```
Krastanov, line 50: I assume most of those are from page 13? Could you add a few more examples from there, going into the large t values? No need to expand the polynomials by hand, you can just have a few predefined variables and write the expressions exactly as they are in page 13 so that it is easy for a reader to compare them.

For these same ones, could you also test the k value (by computing the rank) and the t value (by computing the distance). These would be very solid correctness checks.

Krastanov (changes requested):
This looks very polished! I left a few minor comments and I think this is pretty much the last round of review.

- Some minor comments on grammar and style in the documentation with suggested changes. Feel free to directly accept the suggested changes (I now know that this triggers separate new commits - I guess we will have to live with that)
- Requests for some additional tests
- This is probably the one that is technically most difficult. Check out the "Downgrade" test run in github. It seems to be failing. The "Downgrade" run forces all of our dependencies to be the **lowest** possible version, in order to ensure that our package does not cause problems when installed in pre-existing potentially old environments. You are now using features of Nemo that are not used by the rest of our library, and these features are not available in all Nemo versions that we list as compatible. Could you please look at the Project.toml file and delete the bottom half of compatible versions in the list. If the Downgrade test still fails for the same Nemo-related reason, just delete another half, etc, until it stops failing. This is a pretty important test to ensure stability of the system.

src/ecc/codes/classical/bch.jl
Fe-r-oz: The second line is redundant and vague, so let's remove it.

test/test_ecc_bch.jl
Fe-r-oz: Nice suggestions, I have included them.

Also, I am testing for t is 1:2 for m 3:15, I will test for t is 1:m for each m, so that larger values of ts for larger m can be included in the above test as well. 

Testing larger ts for larger m takes some time though. 
```
julia> include("test_ecc_bch.jl")
Test Summary:                   | Pass  Total     Time
Testing properties of BCH codes |  508    508  5m31.1s
Test.DefaultTestSet("Testing properties of BCH codes", Any[], 508, false, false, true, 1.715256468799204e9, 1.715256799918867e9, false, "/home/Transcendence/Desktop/New/bch/QuantumClifford.jl/test/test_ecc_bch.jl")
```

src/ecc/codes/classical/bch.jl
```diff
@@ -0,0 +1,122 @@
...
+
+- Minimal Polynomials and Conjugates: Multiplying by a constant (like `2ˡ`) doesn't affect the position of an element relative to its conjugates within the finite field. Therefore, the minimal polynomials `(φᵢ(x))` for even powers are identical to those for corresponding odd powers `(φᵢ"'"(x))`.
+
+- By exploiting this conjugate relationship, BCH code constructions focus on odd power values of elements. This avoids redundant calculations and simplifies the process, leading to more efficient BCH code construction.
+
+Conway polynomial:
+
+- The finite Galois field `GF(2ᵐ)` can have multiple distinct primitive polynomials of the same degree due to existence of several irreducible polynomials of that degree, each generating the field through different roots. 
+
+- Nemo.jl uses [Conway polynomial](https://en.wikipedia.org/wiki/Conway_polynomial_(finite_fields)), a standard way to represent the primitive polynomial for finite Galois fields `GF(pᵐ)` of degree `m`, where `p` is a prime number.
```
Krastanov, line 82: I am not quite sure why this is part of the documentation. Could you run me through why you added it (e.g. why is this useful for the user)?

src/ecc/codes/classical/bch.jl
```diff
@@ -0,0 +1,122 @@
...
+
+Leveraging Conjugates:
+
+- Even Power Factorization: Even powers `(i)` of an element `(α)` can be factored as `i = i' * 2ˡ`, where `i'` is odd and `l > 1`.
+
+- Conjugacy of Even Powers: `α` raised to the even power `(αⁱ)` is equivalent to `(αⁱ"'") * 2ˡ`, which is a conjugate of `αⁱ`.
+
+- Minimal Polynomials and Conjugates: Multiplying by a constant (like `2ˡ`) doesn't affect the position of an element relative to its conjugates within the finite field. Therefore, the minimal polynomials `(φᵢ(x))` for even powers are identical to those for corresponding odd powers `(φᵢ"'"(x))`.
+
+- By exploiting this conjugate relationship, BCH code constructions focus on odd power values of elements. This avoids redundant calculations and simplifies the process, leading to more efficient BCH code construction.
```
Krastanov, line 76: Is this really about BCH construction? Most of these remarks sound pretty general (about how one should do any group algebra calculations), but then there is one or two comments that make it sound like this is something specifically done to make BCH construction better, which seems misleading.

test/test_ecc_bch.jl
```diff
@@ -0,0 +1,90 @@
+using Test
+using LinearAlgebra
+using QuantumClifford
+using QuantumClifford.ECC
+using QuantumClifford.ECC: AbstractECC, BCH, generator_polynomial
+using Nemo: ZZ, residue_ring, matrix, finite_field, GF, minpoly, coeff, lcm, FqPolyRingElem, FqFieldElem, is_zero, degree, defining_polynomial, is_irreducible 
+
+"""
+- To prove that `t`-bit error correcting BCH code indeed has minimum distance of at least `2 * t + 1`, it is shown that no `2 * t` or fewer columns of its binary parity check matrix `H` sum to zero. A formal mathematical proof can be found on pages 202 and 203 of https://personal.oss.unist.hr/~mnizetic/ZASTITNO%20LINIJSKI%20KODIRANJE/SEMINARSKI%20RADOVI/CH06.pdf.
```
Krastanov, line 9: This seems to be from a textbook of some kind. Do you happen to know what the textbook name is? Asking because otherwise this will be lost in a few years when this website link dies.

test/test_ecc_bch.jl
```diff
@@ -0,0 +1,90 @@
...
+    @test generator_polynomial(BCH(6, 10)) == generator_polynomial(BCH(6, 7)) * (1 + x + x ^ 2 + x ^ 4 + x ^ 6)
+    @test generator_polynomial(BCH(6, 11)) == generator_polynomial(BCH(6, 10)) * (1 + x + x ^ 2)
+    @test generator_polynomial(BCH(6, 13)) == generator_polynomial(BCH(6, 11)) * (1 + x ^ 5 + x ^ 6)  
+    @test generator_polynomial(BCH(6, 15)) == generator_polynomial(BCH(6, 13)) * (1 + x ^ 2 + x ^ 3)
+    
+    test_cases = [(6, 1), (6, 2), (6, 3), (6, 4), (6, 5), (6, 6), (6, 7), (6, 10), (6, 11), (6, 13), (6, 15)]
+    results = [57 51 45 39 36 30 24 18 16 10 7]
+    i = 0
+    for (m, t) in test_cases
+        i += 1
```
Krastanov, line 86:
```suggestion
    for (i, (m, t)) in enumerate(test_cases)
```

test/test_ecc_bch.jl
Krastanov: It seems to be this (out of print) book: Error Control Coding by Shu Lin and Daniel Costello

Krastanov (approved):
Looks good! I will make a few small changes (check the comments and the upcoming commits) and if tests pass I will merge.

You had mentioned that in some situations tests take long time. A convenient workaround that does not sacrifice the quality of the tests is to do something like:

```
for t in rand(all_test_cases, n) # where n is some small number of samples, e.g. 5 or 10
```

instead of

```
for t in all_test_cases
```

That way everything gets tested eventually but you do not need to worry about things being too expensive each time.

Of course, that should be done only after someone has actually executed all the tests at least once, e.g. during developing the package.

Thanks for taking care of this submission. It is quite polished now.

test/test_ecc_bch.jl
```diff
@@ -0,0 +1,88 @@
...
+    @test generator_polynomial(BCH(6, 2)) == generator_polynomial(BCH(6, 1)) * (1 + x ^ 2 + x ^ 4 + x ^ 5 + x ^ 6)
+    @test generator_polynomial(BCH(6, 3)) == generator_polynomial(BCH(6, 2)) * (1 + x + x ^ 6)
+    @test generator_polynomial(BCH(6, 4)) == generator_polynomial(BCH(6, 3)) * (1 + x ^ 3 + x ^ 6)
+    @test generator_polynomial(BCH(6, 5)) == generator_polynomial(BCH(6, 4)) * (1 + x + x ^ 3)
+    @test generator_polynomial(BCH(6, 6)) == generator_polynomial(BCH(6, 5)) * (1 + x + x ^ 2 + x ^ 5 + x ^ 6)
+    @test generator_polynomial(BCH(6, 7)) == generator_polynomial(BCH(6, 6)) * (1 + x + x ^ 4 + x ^ 5 + x ^ 6)
+    @test generator_polynomial(BCH(6, 10)) == generator_polynomial(BCH(6, 7)) * (1 + x + x ^ 2 + x ^ 4 + x ^ 6)
+    @test generator_polynomial(BCH(6, 11)) == generator_polynomial(BCH(6, 10)) * (1 + x + x ^ 2)
+    @test generator_polynomial(BCH(6, 13)) == generator_polynomial(BCH(6, 11)) * (1 + x ^ 5 + x ^ 6)  
+    @test generator_polynomial(BCH(6, 15)) == generator_polynomial(BCH(6, 13)) * (1 + x ^ 2 + x ^ 3)
```
Krastanov, line 80: Where do these come from? I understand that a different initial irreducible polynomial is used so there will be differences, but I am not sure where all of the example polynomials you are using here come from. If you simply reused your code to get these polynomials, this stops being a very good check, because it is not checking against an external source.

src/ecc/codes/classical/bch.jl
Fe-r-oz: About conjugates:
- Even powers (of primitive elements, a), (i.e., φ_i(x)) are redundant and not included since even powers can be factored as i = i' * 2^l (i' odd, l > 1). 
- Interestingly, α^i (α raised to even power) is a conjugate of α^i'. Hence, constructing the minimal polynomial vector, we only consider odd power of primitive element. 
- This helps explain why even power of primitive element 'a' are not iterated over when constructing minimal polynomial. 

About Conway Polynomial:
- The Nemo dev told me today that Nemo utilizes Conway polynomials as a way of representing the primitive polynomials for larger fields. This is different from what is used in some books/slides, that's why I thought it would be better to add this note.
- This will help answer the change in notation that we see for primitive polynomial for GF(2^6) and other fields where books/slides use  p(z) = z^6 + z + 1, and the Nemo gives us primitive polynomial to be p(z) ==  z ^ 6 + z ^ 4 + z ^ 3 + z + 1. They both are irreducible polynomials, which makes them primitive polynomials.

src/ecc/codes/classical/bch.jl
Fe-r-oz: Yes, please see the above comment with regards to this.

test/test_ecc_bch.jl
Fe-r-oz: Yeah, that's the book! Although, the pages are 149  to 150 now.

test/test_ecc_bch.jl
Fe-r-oz: I used the same examples for polynomials which is m =6, t = 1, 2, 3, 4, 5, 6, 7, 10, 11, 13, 15  as given in the table in the [slides](https://web.ntpu.edu.tw/~yshan/BCH_code.pdf), but I had to multiply with a different suitable polynomial because of difference in primitive polynomials for GF(2^6). The approach is similar to what is used in the table in the slides. We see that the BCH(m, t + 1) == BCH(m, t) * (multiply by some polynomial). 

We can still verify the values of k and checking the minimum designed distance which satisfies with the results regardless of change in notation of primitive polynomials. The final call is up to you whether we should include this check or not. 

The difference in primitive polynomials is because:
- Conway polynomials are used to represent primitive polynomials for finite fields in computer algebra. Therefore, Nemo.jl uses conway polynomials, which are different from the examples primitive polynomials found in some books/slides.
- In this case, the table in the slides uses p(z) = z^6 + z + 1 as primitive polynomial, instead of using conway polynomial. The books Error Control Coding by Shu Lin and Daniel Costello also uses p(z) = z^6 + z +1. 
- The Nemo dev told me that standard way of representing primitive polynomials, is by Conway polynomials, in compute algebra systems. Hence, Nemo uses p(z) == z ^ 6 + z ^ 4 + z ^ 3 + z + 1 as primitive polynomial which is its conway polynomial.  Both are irreducible , so the change in representation

src/ecc/codes/classical/bch.jl
Fe-r-oz: The second statement of second line about efficiency is a homegrown comment\afterthought.

'This avoids redundant calculations and simplifies the process, leading to more efficient BCH code construction.' 

If this comment/second line about efficiency is misleading, then please remove it.

test/test_ecc_bch.jl
Fe-r-oz: Pages 168, 169 since PDF(total 624 pages) reader starts counting pages from the first/title pages. 
Pages 1 of chapter 1 starts at page 20 so in this case, pages are 149 to 150.

test/test_ecc_bch.jl
Krastanov: > but I had to multiply with a different suitable polynomial

How were these multiplicative factors derived? The reason for my question is so that I understand whether the test is independent from the already existing implementation you are contributing (e.g. from a textbook) or if they were generated with the assumption that your code is already correct.

src/ecc/codes/classical/bch.jl
Krastanov: noted, I will undo some of the deletings

src/ecc/codes/classical/bch.jl
Krastanov: With this explanation it is not misleading, but its original meaning was not as clear. I will update the text.

test/test_ecc_bch.jl
Fe-r-oz: These factors are minimal polynomials for GF(2^6).  We can use minpoly function as it is or use it's results:

```
julia> GF2x, x = GF(2)["x"]
(Univariate polynomial ring in x over GF(2), x)

julia> GF2ʳ, a = finite_field(2, 6, "a")
(Finite field of degree 6 and characteristic 2, a)


julia> generator_polynomial(BCH(6, 5)) == generator_polynomial(BCH(6, 4)) * minpoly(GF2x, a^9)
true

julia> generator_polynomial(BCH(6, 4)) == generator_polynomial(BCH(6, 3)) * minpoly(GF2x, a^7)
true

julia> generator_polynomial(BCH(6, 3)) == generator_polynomial(BCH(6, 2)) * minpoly(GF2x, a^5)
true

julia> generator_polynomial(BCH(6, 2)) == generator_polynomial(BCH(6, 1)) * minpoly(GF2x, a^3)
true

julia> generator_polynomial(BCH(6, 6)) == generator_polynomial(BCH(6, 5)) * minpoly(GF2x, a^11)
true

julia> generator_polynomial(BCH(6, 10)) == generator_polynomial(BCH(6, 7)) * minpoly(GF2x, a^15)
true

julia> generator_polynomial(BCH(6, 11)) == generator_polynomial(BCH(6, 10)) * minpoly(GF2x, a^21)
true

julia> generator_polynomial(BCH(6, 13)) == generator_polynomial(BCH(6, 11)) * minpoly(GF2x, a^23)
true

julia> generator_polynomial(BCH(6, 15)) == generator_polynomial(BCH(6, 13)) * minpoly(GF2x, a^27)
true
```

More formally,

The power of a in minpoly(GF2x, a^(n_t + n_(t - 1) - x )) where x = difference between t terms - 1 (if any difference)

In the case when you want to derive a factor for (6, 15) from (6, 13) the power is  n_15 + n_13 - 1. 1 is difference between 15 - 13 - 1 since we are skipping a^14

For generally, 
```

julia> generator_polynomial(BCH(6, 20)) == generator_polynomial(BCH(6, 15)) * minpoly(GF2x, a^31)
true

t =  (n_20 + n_15 ) -  difference between (20 -15 -1).   
t = 4
```

Krastanov: Yeah, converting the tests to use the minpoly function would make it more obvious why the tests should be trusted. Could you make that change?

Krastanov: and have the power in the minpoly come from a formula, not just having it hardcoded

Fe-r-oz: >  Could you make that change?

Alrighty.

Fe-r-oz:
> Apparently, I forgot to change the page numbers from 202, 203 to 168, 169.

Fixed the page numbers as well. I have polished this up as well!

Krastanov:
This is great! I removed a minor part of the docstring that I do not understand well. I do not want to keep you waiting on a merge over such a minor thing, given that the rest seems already very polished. If you consider that part of the docstring important, we can discuss it over a separate pull request.

Do you want me to start reviewing a specific next pull request? I see you have submitted a few, but I am not sure which one you would like to start with.

Krastanov:
Woohoo!

Fe-r-oz:
> Do you want me to start reviewing a specific next pull request? 

Please review Reed-Solomon next!

PR #275, opened 2024-05-16
Adding family of classical binary [[2ʳ - 1, 2ʳ - 1 - r, 3]] Hamming c…

Fe-r-oz:
Adding family of classical binary [[2ʳ - 1, 2ʳ - 1 - r, 3]] Hamming codes along with Hamming/Sphere-packing, Gilbert-Varshamov, Griesmer and other bounds. I believe it would be nice to have these bounds defined. 

Please do let me know that I have defined the bounds at the correct places, meaning in ECC.jl whenever you review this PR. Thanks!

### DIFF
--- a/docs/src/references.bib
+++ b/docs/src/references.bib
@@ -379,3 +379,21 @@
   year={2024},
   publisher={Pearson}
 }
+
+@article{hamming1950error,
+  title={Error detecting and error correcting codes},
+  author={Hamming, Richard W},
+  journal={The Bell system technical journal},
+  volume={29},
+  number={2},
+  pages={147--160},
+  year={1950},
+  publisher={Nokia Bell Labs}
+}
+
+@book{huffman2010fundamentals,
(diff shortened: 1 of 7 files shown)


nzy1997:
Everything in this PR looks good. The failed tests appear to be unrelated.

Fe-r-oz:
Just another minor note regarding CECCs: attaching below a general set of guidelines Stefan provided, which will be helpful when importing classical CECCs from other libraries, as they outline consistency checks we can apply to the imported check matrices.

> - checking the rank of the binary matrices and verifying we do not have redundant rows
> - checking that each column of a binary matrix has at least one non-zero entry
> - finding some code database and comparing the binary matrices generated here to the matrices from the database (probably one of the most valuable options, although it also has potential complications if the database uses different bit order conventions) 
> - similarly to the database option, finding another piece of software that also generates these codes and running comparisons between the two

PR #277, opened 2024-05-18
 Bug fix to the `parity_checks(ReedMuller(r, m))` along with `RecursiveReedMuller` code for cross-reference

Fe-r-oz:
Back then, I didn't had access to this book Fundamentals of Error Correcting codes by Huffman, that's why I skipped some details about Reed-Muller codes as now I am reading it now.

I have polished this PR #244 and added more details from the literature. This provides RM a complete introduction! Tested for all the cases.

@Krastanov Kindly please handle this small PR before reviewing ShortenedMDS! Thanks!


src/ecc/ECC.jl
```diff
@@ -1,6 +1,7 @@
 module ECC
 
 using LinearAlgebra
+using LinearAlgebra: I
```
Fe-r-oz, line 4: This is to avoid conflict warnings like QuantumClifford and LinearAlgebra both use I.

Fe-r-oz:
Please review so that we can apply the fix as soon as feasible!

I have used reduced row echelon form test to compare all the cases of RecursiveReedMuller with ReedMuller,

`@test rref(generator(ReedMuller(r, m))) == rref(generator(RecursiveReedMuller(r, m)))`

The generator for RecursiveReedMuller is widely used as subroutine for construction of new quantum codes. Therefore, added this as well.

src/ecc/ECC.jl
```diff
@@ -69,6 +70,8 @@ In a [polynomial code](https://en.wikipedia.org/wiki/Polynomial_code), the gener
 """
 function generator_polynomial end
 
+function generator end
```
Krastanov, line 73: What is `generator`? Probably needs a docstring. Do many families of codes have this defined for them?

src/ecc/codes/classical/recursivereedmuller.jl
```diff
@@ -0,0 +1,71 @@
...
+G(r, m - 1) & G(r, m - 1) \\
+0 & G(r - 1, m - 1)
+\end{bmatrix} 
+```
+
+Here, the matrix 0 denotes an all-zero matrix with dimensions matching `G(r - 1, m - 1)`. This recursive approach facilitates the construction of higher-order Reed-Muller codes based on the generator matrices of lower-order codes.
+
+The dimension of `RM(m - r - 1, m)` equals the dimension of the dual of `RM(r, m)`. Thus, `RM(m - r - 1, m) = RM(r, m)^⊥` which indicates that the parity check matrix of `RM(r, m)` is the generator matrix for `RM(m - r - 1, m)`.
+"""
+abstract type ClassicalCode end
```
Krastanov, line 19: This docstring does not seem to be meant for `ClassicalCode`

src/ecc/codes/classical/recursivereedmuller.jl
```diff
@@ -0,0 +1,71 @@
...
+
+struct RecursiveReedMuller <: ClassicalCode
+    r::Int
+    m::Int
+
+    function RecursiveReedMuller(r, m)
+        0 ≤ r ≤ m || throw(ArgumentError("Invalid parameters: r must be non-negative, r ≤ m and m - r - 1 ≥ 0 in order to obtain a valid code."))
```
Krastanov, line 26: m-r-1>=0 is not being tested here, right? Do you mean r<m instead of r<=m?

src/ecc/codes/classical/recursivereedmuller.jl
```diff
@@ -0,0 +1,71 @@
...
+end
+
+"""
+This function generates the generator matrix, `G`, for RecursiveReedMuller`(RecursiveReedMuller(r, m))` error-correcting codes. 
+
+`generator(RecursiveReedMuller(r, m))`:
+- `m`: Positive integer representing the message length.
+- `r`: The order of the Reed-Muller code. Must satisfy `0 ≤ r ≤ m`.
+"""
+function _recursiveReedMuller(r::Int, m::Int)
```
Krastanov, line 38: not sure if this docstring is at the correct function... The function cided in the docstring is not this one

src/ecc/codes/classical/recursivereedmuller.jl
```diff
@@ -0,0 +1,71 @@
...
+    return _recursiveReedMuller(c.r, c.m)
+end
+
+"""
+This function generates the parity check matrix, `H`, for RecursiveReedMuller`(RecursiveReedMuller(r, m))` error-correcting codes. 
+
+`parity_checks(RecursiveReedMuller(r, m))`:
+- `m`: Positive integer representing the message length.
+- `r`: The order of the Reed-Muller code. Must satisfy `0 ≤ r ≤ m`.
+"""
```
Krastanov, line 62: This docstring is unnecessary if `RecursiveReedMuller` is already documented. We have document `parity_checks` at the location of the function declaration. No need to document every method with repetitive docstrings.

src/ecc/codes/classical/reedmuller.jl
```diff
@@ -1,8 +1,16 @@
 """The family of Reed-Muller codes, as discovered by Muller in his 1954 paper [muller1954application](@cite) and Reed who proposed the first efficient decoding algorithm [reed1954class](@cite).
 
+Let `m` be a positive integer and `r` a nonnegative integer with `r ≤ m`. These linear codes, denoted as `RM(r, m)`, have order `r` (where `0 ≤ r ≤ m`) and codeword length `n` of `2ᵐ`.
+
+Two special cases exist:
+    1. `RM(0, m)`: This is the `0ᵗʰ`-order `RM` code, similar to the binary repetition code with length `2ᵐ`. It's characterized by a single basis vector containing all ones.
+    2. `RM(m, m)`: This is the `mᵗʰ`-order `RM` code. It encompasses the entire field `F(2ᵐ)`, representing all possible binary strings of length `2ᵐ`.
```
Krastanov, line 7: Are these two special cases part of the test suite?

src/ecc/codes/classical/reedmuller.jl
```diff
@@ -1,8 +1,16 @@
...
+
 You might be interested in consulting [raaphorst2003reed](@cite), [abbe2020reed](@cite), and [djordjevic2021quantum](@cite) as well.
 
-The ECC Zoo has an [entry for this family](https://errorcorrectionzoo.org/c/reed_muller)
+The dimension of `RM(m - r - 1, m)` equals the dimension of the dual of `RM(r, m)`. Thus, `RM(m - r - 1, m) = RM(r, m)^⊥` which indicates that the parity check matrix of `RM(r, m)` is the generator matrix for `RM(m - r - 1, m)`.
```
Krastanov, line 11: Are these properties part of the test suite?

src/ecc/codes/classical/recursivereedmuller.jl
Krastanov: Actually, there seems to be a much bigger problem. You are defining `ClassicalCode` in multiple locations. There is another abstract type like this defined in the next file.

src/ecc/codes/classical/reedmuller.jl
```diff
@@ -12,27 +20,51 @@ struct ReedMuller <: ClassicalCode
     m::Int
 
     function ReedMuller(r, m)
-        if r < 0 || m < 1 || m >= 11
-            throw(ArgumentError("Invalid parameters: r must be non-negative and m must be positive and < 11 in order to obtain a valid code and to remain tractable"))
-        end
+        0 ≤ r ≤ m || throw(ArgumentError("Invalid parameters: r must be non-negative and r ≤ m. Additionally, m must be positive and < 11 in order to obtain a valid code and to remain tractable"))
```
Krastanov, line 23: The m<11 check is deleted (which seems like a good idea) but the error message does not seem to be updated to remove the unnecessary warning

src/ecc/codes/classical/reedmuller.jl
```diff
@@ -12,27 +20,51 @@ struct ReedMuller <: ClassicalCode
...
-    m=c.m
-    xi = [variables_xi(m, i) for i in 0:m - 1]
-    row_matrices = [reduce(vmult, [xi[i + 1] for i in S], init = ones(Int, 2^m)) for s in 0:r for S in combinations(0:m - 1, s)]
+"""
+This function generates the generator matrix, `G`, for Reed-Muller `(RM(r, m))` error-correcting codes. 
+
+`generator(ReedMuller(r, m))`:
+- `m`: Positive integer representing the message length.
+- `r`: Nonnegative integer less than or equal to `m`, specifying the code's order.
+"""
```
Krastanov, line 42: We should probably have one single `generator` docstring at the place the function is first introduced. No need to have a repetitive new docstring for it, if the `ReedMuller` type has already been documented well.

src/ecc/codes/classical/reedmuller.jl
```diff
@@ -12,27 +20,51 @@ struct ReedMuller <: ClassicalCode
...
+end
+
+"""
+This function generates the parity check matrix, `H`, for Reed-Muller `(RM(r, m))` error-correcting codes. 
+
+`parity_checks(ReedMuller(r, m))`:
+- `m`: Positive integer representing the message length.
+- `r`: Nonnegative integer less than or equal to `m`, specifying the code's order.
+"""
+function parity_checks(c::ReedMuller)
```
Krastanov, line 62: similarly, no need for this docstring, same comment as mentioned a couple of times above

test/Project.toml
```diff
@@ -20,6 +20,7 @@ Quantikz = "b0d11df0-eea3-4d79-b4a5-421488cbf74b"
 QuantumInterface = "5717a53b-5d69-4fa3-b976-0bf2f97ca1e5"
 QuantumOpticsBase = "4f57444f-1401-5e15-980d-4471b28d5678"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+RowEchelon = "af85af4c-bcd5-5d23-b03a-a909639aa875"
```
Krastanov, line 23: Nemo.jl seems to already provide a rref function (see http://nemocas.github.io/Nemo.jl/v0.6/matrix/#reduced-row-echelon-form ) -- let's use theirs, as it is a more established package that is less probable to disappear.

test/test_ecc_reedmuller.jl
```diff
@@ -1,68 +1,77 @@
...
-                                             1 1 1 1 1 1 1 1 0 0 0 0 0 0 0 0;
-                                             1 1 1 1 0 0 0 0 1 1 1 1 0 0 0 0;
-                                             1 1 0 0 1 1 0 0 1 1 0 0 1 1 0 0;
-                                             1 0 1 0 1 0 1 0 1 0 1 0 1 0 1 0;
-                                             1 1 1 1 0 0 0 0 0 0 0 0 0 0 0 0;
-                                             1 1 0 0 1 1 0 0 0 0 0 0 0 0 0 0;
-                                             1 0 1 0 1 0 1 0 0 0 0 0 0 0 0 0;
-                                             1 1 0 0 0 0 0 0 1 1 0 0 0 0 0 0;
-                                             1 0 1 0 0 0 0 0 1 0 1 0 0 0 0 0;
-                                             1 0 0 0 1 0 0 0 1 0 0 0 1 0 0 0]
```
Krastanov, line 67: These are pretty suspicious. Were the previously implemented results wrong?

Are these examples supposed to be the generator matrices or the parity check matrices?

test/test_ecc_reedmuller.jl
```diff
@@ -1,68 +1,77 @@
...
+function designed_distance(matrix, m, r)
+    for row in eachrow(matrix)
+        count = sum(row)
+        if count >= 2 ^ (m - r)
+            return true
+        end
     end
-    return total
+    return false
 end
```
Krastanov, line 17: I am not sure I understand this function. What is it checking for?

test/test_ecc_reedmuller.jl
```diff
@@ -1,68 +1,77 @@
...
             @test computed_rank == expected_rank
+            @test designed_distance(H, m, r) == true
+            @test rate(ReedMuller(r, m)) == sum(binomial.(m, 0:r)) / 2 ^ m
+            @test code_n(ReedMuller(r, m)) == 2 ^ m
+            @test code_k(ReedMuller(r, m)) == sum(binomial.(m, 0:r))
+            @test distance(ReedMuller(r, m)) == 2 ^ (m - r)
+            if m - r - 1 ≥ 0 
+                @test rref(parity_checks(ReedMuller(r, m))) == rref(parity_checks(RecursiveReedMuller(r, m)))
+            end
+            @test rref(generator(ReedMuller(r, m))) == rref(generator(RecursiveReedMuller(r, m)))
```
Krastanov, line 35: This is a really neat and helpful set of checks!

Why is there an if statement? What is the problem with m=r?

Krastanov (changes requested):
This is a pretty helpful set of additional implementation for cross reference. I did a quick first round of review -- there are some stylistic requests for improvement and a couple of questions on definitions and correctness checks.

src/ecc/ECC.jl
Fe-r-oz: It's the Generator Matrix for Reed-Muller codes. In most of the papers, the people show the generator matrices instead of the parity check matrices.  

Also,  the code conversion paper used the generator for `recursivereedmuller `codes, so adding a function for generator

In the paper I used, the matrices they used were generators: https://arxiv.org/pdf/2002.03317. (Page 5) I read the paper carefully this time 👀

src/ecc/codes/classical/reedmuller.jl
Fe-r-oz: Yep,` r` starts at `0` at ends at m, so we test both` (0, m) `and `(m, m)`

```
    for m in 3:10
        for r in 0:m
```

test/test_ecc_reedmuller.jl
Fe-r-oz: The results were correct but I mistook generators matrices for parity check matrices. 

I found this out when I was reading the [code conversion paper ](https://arxiv.org/pdf/1112.2417)and authors were talking about 'generators' and soon I realized, I had implemented generator matrices. Sorry for this inconvenience.

test/test_ecc_reedmuller.jl
Fe-r-oz: It's checking that the minimum hamming distance must be at least greater than/equal to `2^(m - r).`

test/test_ecc_reedmuller.jl
Fe-r-oz: If statement make sure that `r` is not  `< 0` since the new r for parity check matrix is` m - r - 1.`

We do check in struct whether r is less than 0, but since for parity check matrix, we select r as  `m - r - 1`, and m  as m `(m - r - 1, m)`, for some cases near larger values, `m - r - 1` is evaluated to `< 0 `

So, even with this set of checks  `0 ≤ r ≤ m`, it's possible to get   `m - r - 1` < 0 which shouldn't be the case. That's why I added the if check. Is there any better way to do this?

src/ecc/codes/classical/reedmuller.jl
Krastanov: These case are being generated as part of the tests for all RM codes, but the very special properties of these edge cases are not tested, are they? E.g. The fact that RM(0.m) is the repretition code and the fact that RM(m,m) are the entire field?

Such edge cases make for pretty good correctness tests.

test/test_ecc_reedmuller.jl
Krastanov: The difference between parity check matrix and generator matrix is very big -- I would not say these were correct, this is a serious error that we should be careful to guard against. Are there any automated consistency tests that we can put it so that such errors are automatically detected in the future? (it is not enough to say that one would be careful in the future -- carefulness is never enough to protect against bugs)

For instance, do we automatically test that the number of rows and columns fulfills correctly the necessary relationships with code_n, code_k and others?

src/ecc/codes/classical/reedmuller.jl
Fe-r-oz: Indeed. Explicit checks would be nice. I will add them.

test/test_ecc_reedmuller.jl
Krastanov: But if the first row has a weight bugger than 2^(m-r) this would immediately return true, even if the second row has weight smaller than 2^(m-r)? Is that the intended behavior?

test/test_ecc_reedmuller.jl
Fe-r-oz: > For instance, do we automatically test that the number of rows and columns fulfills correctly the necessary relationships with code_n, code_k and others?

Yeah, that makes sense. And during the time of implementation, we should check what's the parity checks method is returning, and have true or false. Similar to `isCSS,` maybe `isParityCheck `method.

src/ecc/codes/classical/reedmuller.jl
Fe-r-oz: I will add this explicit check of checking one dimension of `RM(m - r - 1, m)` to dual of `RM(r, m)`

test/test_ecc_reedmuller.jl
Fe-r-oz: I think better to check all rows first, and then return `true` would be better.

src/ecc/codes/classical/recursivereedmuller.jl
Fe-r-oz: m-r-1 is the r for parity check matrix. And r should be always >=0 (hence this condition: m-r-1>=0). 

I will add this check here.

test/test_ecc_reedmuller.jl
Fe-r-oz: >Is there any better way to do this? 

Rather than if statement during tests, better to include the check in the definition that `m - r - 1`.

test/Project.toml
Fe-r-oz: Nemo dev told me better to use `Nemo.echelon_form(QQ, matrix)` rather than `Nemo.rref` as latter is going be to removed in later versions.

test/test_ecc_reedmuller.jl
Krastanov: Yup, that would be better, but I do not think that is what you function is doing, right?

If that is what it is supposed to be doing, the simplest fix is to check for failure in the loop and return false, and only if the whole loop finishes without a failure, you return true (kinda the opposite of what is happening here)

Fe-r-oz: I have fixed the entire function! I realized it. Exactly!

src/ecc/codes/classical/reedmuller.jl
```diff
@@ -1,38 +1,55 @@
...
+"""
 struct ReedMuller <: ClassicalCode
     r::Int
     m::Int
 
     function ReedMuller(r, m)
-        if r < 0 || m < 1 || m >= 11
-            throw(ArgumentError("Invalid parameters: r must be non-negative and m must be positive and < 11 in order to obtain a valid code and to remain tractable"))
+        if r < 0 || r > m || m > 11
+            throw(ArgumentError("Invalid parameters: r must be non-negative, r ≤ m and m - r - 1 ≥ 0. Additionally, m must be positive and < 11 in order to obtain a valid code and to remain tractable."))
```
Krastanov, line 21: I think this was in the previous review, but maybe I forgot to add it:

The argument error message and the check do not seem to actually match.

Also, isn't r ≤ m and m - r - 1 ≥ 0 equivalent to just r<m?

I think it is best to remove the <11 constraint. It seems a bit too stringent and it really depends on the amount of RAM the user has.

Krastanov:
I probably will be able to merge this shortly if I do not find any new issues besides what I just mentioned. Could you update the changelog so that it is ready for merge? Mention there is a bug fix to the parameter_checks matrix (if this was a public type this would have been a breaking change, so it is good to document these clearly)

src/ecc/codes/classical/reedmuller.jl
Fe-r-oz: You mentioned in in the previous review.

`r > m` make sure that `r<m` because if  `r > m` error message will be thrown.


```
@test_throws ArgumentError ReedMuller(-1, 3)
@test_throws ArgumentError ReedMuller(1, 0)
@test_throws ArgumentError ReedMuller(4, 2)

@test_throws ArgumentError RecursiveReedMuller(-1, 3)
@test_throws ArgumentError RecursiveReedMuller(1, 0)
@test_throws ArgumentError RecursiveReedMuller(4, 2)
```

I will remove the  <11 constraint.

Fe-r-oz: I forgot to update the argument error message, which is my mistake.

Fe-r-oz:
This was changelog of 244 "Implemented the classical Reed-Muller code"

I think RM was not exported, so I am not sure about Public Type/Breaking change part.

So, maybe `Bug fix to parity_checks(ReedMuller(r, m)) of classical Reed-Muller code along with RecursiveReedMuller code implementation for cross-reference`

Please feel free to update CHANGELOG accordingly.

Fe-r-oz:
###  Upcoming PR discussion

Good news, @Krastanov  :) 

Sometime ago, I discussed a paper about[ code conversion between Quantum Reed Muller code and Steane Code](https://arxiv.org/pdf/1403.2734). Turns out that the Quantum Reed Muller is really simple!  Now, I understand what ECCZOO is talking about in their [QRM page](https://errorcorrectionzoo.org/c/quantum_reed_muller)!

There is a lot of rigorous testing that I am planning to do with these quantum codes and I think the entire paper can be implemented as upcoming PR. The parameters of the QRM are: [[n = 2^m − 1, k = 1, d = 3]]. Now, this are CSS Quantum Reed Muller codes! 

Please checkout the two examples! Example of the popular 15 qubit QRM code that is used in magic state distillation: From visual inspection, we can see it as two copies to steane code.  Steanecode is QRM(3)

```
#notes
# RM(1,m): dual is RM(m -2, m), generator(RM(m -2, m)) = H_m
# Shortened RM_bar(1, m): first row and first column are deleted from generator(RM(1, m))
julia> RM14 = generator(RecursiveReedMuller(1, 4));

julia> RM24 = generator(RecursiveReedMuller(4 - 2, 4)); #dual

julia> QRM = CSS(RM14[2:end, 2:end], RM24[2:end, 2:end])
CSS(Bool[1 0 … 0 1; 0 1 … 1 1; 0 0 … 1 1; 0 0 … 1 1], Bool[1 0 … 0 0; 0 1 … 1 0; … ; 0 0 … 1 1; 0 0 … 1 1])

julia> Stabilizer(QRM)
+ X_X_X_X_X_X_X_X
+ _XX__XX__XX__XX
+ ___XXXX____XXXX
+ _______XXXXXXXX
+ Z___Z___Z___Z__
+ _Z___Z___Z___Z_
+ __Z___Z___Z___Z
+ ___ZZZZ____ZZZZ
+ ____Z_Z_____Z_Z
+ _____ZZ______ZZ
+ _______ZZZZZZZZ
+ ________Z_Z_Z_Z
+ _________ZZ__ZZ
+ ___________ZZZZ

julia> naive_encoding_circuit(parity_checks(Stabilizer(QRM)))

43-element Vector{QuantumClifford.AbstractOperation}:
 sCNOT(15,5)
 sCNOT(15,6)
 sCNOT(15,7)
 sCNOT(15,9)
 sCNOT(15,10)
 sCNOT(15,12)
 sHadamard(1)
 sZCX(1,5)
 sZCX(1,6)
 sZCX(1,8)
 sZCX(1,9)
 sZCX(1,11)
 sZCX(1,13)
 sZCX(1,15)
 sHadamard(2)
 sZCX(2,5)
 sZCX(2,7)
 ⋮
 sZCX(3,12)
 sZCX(3,13)
 sZCX(3,14)
 sZCX(3,15)
 sHadamard(4)
 sZCX(4,9)
 sZCX(4,10)
 sZCX(4,11)
 sZCX(4,12)
 sZCX(4,13)
 sZCX(4,14)
 sZCX(4,15)
 sSWAP(8,4)
 sSWAP(7,4)
 sSWAP(6,4)
 sSWAP(5,4)
 sSWAP(4,3) 
```

More exciting is this fact that QuantumReedMuller(3) is Steane Code. The proof is as follows: 
```

julia> RM13 = generator(RecursiveReedMuller(1, 3));

julia> RM23 = generator(RecursiveReedMuller(3 - 2, 3));

julia> QRM = CSS(RM14[2:end, 2:end], RM24[2:end, 2:end])
CSS(Bool[1 0 … 0 1; 0 1 … 1 1; 0 0 … 1 1; 0 0 … 1 1], Bool[1 0 … 0 0; 0 1 … 1 0; … ; 0 0 … 1 1; 0 0 … 1 1])

julia> QRM_steane = CSS(RM13[2:end, 2:end], RM23[2:end, 2:end])
CSS(Bool[1 0 … 0 1; 0 1 … 1 1; 0 0 … 1 1], Bool[1 0 … 0 1; 0 1 … 1 1; 0 0 … 1 1])

julia> Steane = Stabilizer(QRM_steane)
+ X_X_X_X
+ _XX__XX
+ ___XXXX
+ Z_Z_Z_Z
+ _ZZ__ZZ
+ ___ZZZZ

julia> canonicalize!(parity_checks(Steane7()))
+ X_X_X_X
+ _XX__XX
+ ___XXXX
+ Z_Z_Z_Z
+ _ZZ__ZZ
+ ___ZZZZ

julia> canonicalize!(parity_checks(Steane7())) == Stabilizer(QRM_steane)
true

```
I am very excited to work on this as soon as the RecursivePR is finished!

Fe-r-oz:
Hello @Krastanov, the quantum CSS Reed Muller code PR is ready but I am waiting for this one to be finished so that I can use the constructors introuced in this PR in next one! 

Please review this one so I can polish the next quantum code PR for review 😬 I am excited to submit quantum codes from classical codes finally!

Fe-r-oz:
Hello @Krastanov, just a small reminder that the current implementation of `ReedMuller` needs a bug fix as it's bad and ugly at the moment.  Please help review the PR so bug fix can be applied soon.

Thank you for the valuable feedback that helped improved the PR significantly.

In this PR, I have made sure to test two different representations (binomial polynomial rep vs recursive) for cross-reference and added several correctness tests and provided complete documentation.

Fe-r-oz:
I tried to rename the branch to `FixRM` instead of `PolishRM` and the branch got deleted! Hopefully it's restored. 

I was following your comments that "its good to document changes clearly" so I thought let's rename the branch. 🙄

Fe-r-oz:
Removing `CHANGELOG `to prevent merge conflicts.

```
changelog:

- **(fix)** Bug fix to the `parity_checks(ReedMuller(r, m))` of classical `Reed-Muller` code along with `RecursiveReedMuller` code implementation for cross reference.
```

Krastanov (changes requested):
Regrettably, this needs a rebase after the recent rework of the test suite. Sorry for the runaround, I know this is pretty annoying.

Fe-r-oz:
No problem at all. `Testitems` is cool! 

Rebased. Thanks for the reminder.

Krastanov (changes requested):
this looks great! Sorry for the slow review. After a few minor fixes to the docstrings it should be good to merge.

src/ecc/codes/classical/recursivereedmuller.jl
```diff
@@ -0,0 +1,60 @@
...
+- `G(r - 1, m - 1)`: Generator matrix of `RM(r - 1, m - 1)`
+- `G(r, m - 1)`: Generator matrix of `RM(r, m - 1)`
+
+The generator matrix `G(m, r)` of `RM(m, r)` is formulated as follows in matrix notation:
+
+```math
+G(m, r) = \begin{bmatrix}
+G(r, m - 1) & G(r, m - 1) \\
+0 & G(r - 1, m - 1)
+\end{bmatrix}
```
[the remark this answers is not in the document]
Fe-r-oz: Thanks for your question! I have cross verified with repos that use ```math env for a lot of math expressions. When we use this, it seems we don't use double slashes, although, if we are not using, double slashes absolutely make sense.  Also, I checked - double slashes are not rendered within math environment.

Hopefully, we will see a pretty matrix with math in dev doc. I will cross verify from dev doc as well.

Fe-r-oz:
Thank you for your helpful suggestions. 

P.S. The `build` fail is due to upstream `Persistant tasks` test fail from `QuantumSavory.jl`.

Fe-r-oz:
@Krastanov, pinging for review.

Krastanov:
Thank you for fixing this

PR #278, opened 2024-05-21
Stabilizer group function completion and testing

IsaacP1234:
Currently unfinished
Goal is to finish group tableaux functions and write tests for them


src/grouptableaux.jl
```diff
@@ -0,0 +1,255 @@
...
+    return s[1:r,:]
+end
+g =path_graph(3)
+s = Stabilizer(g)
+plot = graphplot(g)
+print(s)
+println("group")
+print(groupify(s))
+println(get_generating_set(groupify(s)))
+println("tested")
```
Krastanov, line 61: This is code that runs at library definition time, while Kenneth probably meant it as some manual checks he was doing while this was a separate non-library file for experimentation. It is currently causing errors during compilation.

Krastanov:
I assume #293 is supposed to supersede this? Closing for now if that is the case.

PR #279, opened 2024-05-22
ReedSolomon codes as subfamily of q-ary BCH

Fe-r-oz:
In Draft Stage: Completed the Generator Polynomial. Goal is to implement them as subfamily of BCH codes. 

ReedSolomon codes are introduced as subfamily of BCH codes with n  = q - 1. So, q is never 2.

### DIFF
--- a/docs/src/references.bib
+++ b/docs/src/references.bib
@@ -379,3 +379,49 @@
   year={2024},
   publisher={Pearson}
 }
+
+@article{reed1960polynomial,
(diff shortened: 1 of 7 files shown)


Fe-r-oz:
The q-ary (non-binary) codes are not needed. As for Reed Solomon codes, specifically as subcode of BCH, consider using `CodingTheory.jl `as cyclic codes are better developed there.

PR #280, opened 2024-05-22
q-ary/Non-binary BCH codes

Fe-r-oz:
Reed Solomon are non-binary BCH codes (q-ary BCH) where q is a prime number other than 2). 

This PR generalizes already implemented binary BCH codes to q-ary BCH codes. q-ary BCH codes can be used as Binary BCH codes or non-Binary BCH codes.


Fe-r-oz:
The q-ary (non-binary) codes are not needed.

PR #282, opened 2024-05-25
`[[2ʲ, 2ʲ - 2j - 2, 4]]` Gottesman code

Fe-r-oz:
I found a small quantum code, aka a Distance 4 code from Gottesman 1997 thesis. This class of  [[2ʲ, 2ʲ - j - 2, 3]]` Gottesman codes is derived from [[2ʲ, 2ʲ - 2j - 2, 4]] Gottesman codes . The method is taken from the his thesis:

- The first two generators are the Pauli-`X` and Pauli-`Z` operators acting on all qubits, represented by `Mₓ` and `Mz`, respectively. 
- The next `j` generators correspond to `M₁` through `Mⱼ`, which are directly inherited from the `[[2ʲ, 2ʲ - j - 2, 3]]` Gottesman  code's stabilizers. This inclusion ensures that `S` retains the inherent distance-three property of the original Gottesman code. 

So, this is Gottesman(4) that we have in the library
Screenshot_select-area_20240526041636
- The final `j` generators are defined as `Nᵢ = RMᵢR`, where `i` ranges from `1` to `j`. Here, `R` signifies a Hadamard Rotation operation applied to all `2ʲ` qubits, and `Mᵢ` refers to one of the existing generators from the second set `(M₁ to Mⱼ)`.

Finally, Hadamard transform turns X operator to Z and Z operator to X, so for 1 to j Rows of (M₁ to Mⱼ)
Screenshot_select-area_20240526041620


Fe-r-oz:
Thanks, @Krastanov, for providing clarification in https://github.com/QuantumSavory/QuantumClifford.jl/issues/287!

Gottesman applies H to the state to get Family of Distance 4 codes.

Previously, I was using 

```
 for rows in 1:c.j
       for cols in 1:2^c.j
             if Hⱼ[rows][cols] == (true, false) # X
                  Hⱼ[rows, cols] = (false, true) # X -> Z
             elseif Hⱼ[rows][cols] == (false, true) # Z
                  Hⱼ[rows, cols] = (true, false) # Z -> X
             end 
         end
     end
```

which can now be changed

```
    for qᵢ in 1:nqubits(Hⱼ)
        apply!(Hⱼ, sHadamard(qᵢ))
    end
```

Fe-r-oz:
display

display

src/ecc/codes/gottesman4.jl
```diff
@@ -0,0 +1,25 @@
...
+
+The stabilizer generators of the original `[[2ʲ, 2ʲ - j - 2, 3]]` Gottesman codes are incorporated to create a set of generators for this distance-four code. The resulting stabilizer set, denoted by S, incorporates the following elements: The first two generators are the Pauli-`X` and Pauli-`Z` operators acting on all qubits, represented by `Mₓ` and `Mz`, respectively. The next `j` generators correspond to `M₁` through `Mⱼ`, which are directly inherited from the `[[2ʲ, 2ʲ - j - 2, 3]]` Gottesman  code's stabilizers. This inclusion ensures that `S` retains the inherent distance-three property of the original Gottesman code. The final `j` generators are defined as `Nᵢ = RMᵢR`, where `i` ranges from `1` to `j`. Here, `R` signifies a Hadamard Rotation operation applied to all `2ʲ` qubits, and `Mᵢ` refers to one of the existing generators from the second set `(M₁ to Mⱼ)`. By incorporating the stabilizers of a distance-three code, the constructed set `S` inherently guarantees a minimum distance of three for the resulting distance-four Gottesman code.
```
Krastanov, line 3: A see also to `Gottesman` and an explanation about the different would be useful.

PR #289, opened 2024-06-11
Concatenated quantum code

royess:
Implement the concatenated quantum code, which combines two quantum codes into a large one.

### DIFF
--- a/docs/src/references.md
+++ b/docs/src/references.md
@@ -36,7 +36,7 @@ For quantum code construction routines:
 - [chao2018quantum](@cite)
 - [kitaev2003fault](@cite)
 - [fowler2012surface](@cite)
-- [concatenated1996knill](@cite)
+- [knill1996concatenated](@cite)
 
 For classical code construction routines:
 - [muller1954application](@cite)
--- a/docs/src/references.md
+++ b/docs/src/references.md
@@ -36,6 +36,7 @@ For quantum code construction routines:
 - [chao2018quantum](@cite)
 - [kitaev2003fault](@cite)
 - [fowler2012surface](@cite)
+- [concatenated1996knill](@cite)
 
 For classical code construction routines:
 - [muller1954application](@cite)
--- a/src/ecc/ECC.jl
+++ b/src/ecc/ECC.jl
@@ -8,7 +8,7 @@ using DocStringExtensions
 using Combinatorics: combinations
 using SparseArrays: sparse
 using Statistics: std
-using Nemo: ZZ, residue_ring, matrix, finite_field, GF, minpoly, coeff, lcm, FqPolyRingElem, FqFieldElem, is_zero, degree, defining_polynomial, is_irreducible 
+using Nemo: ZZ, residue_ring, matrix, finite_field, GF, minpoly, coeff, lcm, FqPolyRingElem, FqFieldElem, is_zero, degree, defining_polynomial, is_irreducible
 
 abstract type AbstractECC end
 
@@ -20,6 +20,7 @@ export parity_checks, parity_checks_x, parity_checks_z, iscss,
     CSS,
     Shor9, Steane7, Cleve8, Perfect5, Bitflip3,
     Toric, Gottesman, Surface,
+    concat,
     evaluate_decoder,
     CommutationCheckECCSetup, NaiveSyndromeECCSetup, ShorSyndromeECCSetup,
     TableDecoder,
--- a/src/ecc/ECC.jl
+++ b/src/ecc/ECC.jl
@@ -19,7 +19,7 @@ export parity_checks, parity_checks_x, parity_checks_z, iscss,
     RepCode,
     CSS,
     Shor9, Steane7, Cleve8, Perfect5, Bitflip3,
-    Toric, Gottesman, Surface,
+    Toric, Gottesman, Surface, Concat,
     concat,
     evaluate_decoder,
     CommutationCheckECCSetup, NaiveSyndromeECCSetup, ShorSyndromeECCSetup,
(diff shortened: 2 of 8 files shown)


royess:
I am a bit unsure about the right way to write the `iscss` function for the concatenated quantum code.

A concatenated quantum can be either CSS or non-CSS. In this PR, I make `iscss` return true if the inner and outer codes are both CSS. This can make tests pass but is not a necessary condition in general; two non-CSS codes can also generate a CSS, concatenated code.

I would appreciate some input. Also, please help review this PR. @Krastanov

src/ecc/codes/concat.jl
```diff
@@ -0,0 +1,45 @@
...
+The construction is the following: replace each qubit in code c₂ with logical qubits encoded by code c₁.
+The resulting code will have n = n₁ * n₂ qubits and k = k₁ * k₂ logical qubits.
+"""
+struct Concat <: AbstractECC
+    c₁::AbstractECC
+    c₂::AbstractECC
+end
+
+"""The function returns the parity checks of the concatenated code."""
+function concat(c₁, c₂)
```
Krastanov, line 12: I think this should be just `parity_checks(Concat(c1, c2))`. I want to avoid introducing new single-purpose functions to the public API.

src/ecc/codes/concat.jl
```diff
@@ -0,0 +1,45 @@
...
+code_n(c::Concat) = code_n(c.c₁) * code_n(c.c₂)
+
+code_k(c::Concat) = code_k(c.c₁) * code_k(c.c₂)
+
+function parity_checks(c::Concat)
+    concat(c.c₁, c.c₂)
+end
+
+function iscss(c::Concat)
+    iscss(c.c₁) && iscss(c.c₂) # but it is possible that c.c₁ and c.c₂ are non-css but c is css
```
Krastanov, line 44: we can return `nothing` for the unknown cases, we just need to document that possibility where `iscss` is originally declared

src/ecc/ECC.jl
```diff
@@ -19,7 +19,8 @@ export parity_checks, parity_checks_x, parity_checks_z, iscss,
     RepCode,
     CSS,
     Shor9, Steane7, Cleve8, Perfect5, Bitflip3,
-    Toric, Gottesman, Surface,
+    Toric, Gottesman, Surface, Concat,
+    concat,
```
Krastanov, line 23: I want to keep the public API as small as possible. Let's remove `concat` given that it is single purpose

src/ecc/codes/concat.jl
```diff
@@ -0,0 +1,45 @@
+"""
+Code concatenation of two quantum codes [concatenated1996knill](@cite)--the inner code c₁ and the outer code c₂.
```
Krastanov, line 2:
```suggestion
`Concat(c₁, c₂)` is a code concatenation of two quantum codes [concatenated1996knill](@cite)

The inner code c₁ and the outer code c₂.
```

Otherwise c* is not defined. Also adding an empty line to separate header from rest of docstring.

src/ecc/codes/concat.jl
```diff
@@ -0,0 +1,45 @@
...
+The construction is the following: replace each qubit in code c₂ with logical qubits encoded by code c₁.
+The resulting code will have n = n₁ * n₂ qubits and k = k₁ * k₂ logical qubits.
```
Krastanov, line 4:
```suggestion
The resulting code will have `n = n₁ × n₂` qubits and `k = k₁ × k₂` logical qubits.
```

Krastanov (changes requested):

Krastanov:
Converting back to draft to make it easier for me to track my review backlog. Do not hesitate to convert it back to non-draft and to ping me for another review.

test/test_ecc_codeproperties.jl
```diff
@@ -22,7 +22,7 @@ end
 @testset "is CSS" begin
     for code in all_testablable_code_instances()
         H = parity_checks(code)
-        @test iscss(code) == is_css_matrix(H)
+        @test iscss(code) in (is_css_matrix(H), nothing)
```
Krastanov, line 25: I changed this from a list to a tuple to avoid allocations. Not important in this place, but a good habit to keep.

src/ecc/codes/concat.jl
```diff
@@ -0,0 +1,47 @@
...
+
+code_n(c::Concat) = code_n(c.c₁) * code_n(c.c₂)
+
+code_k(c::Concat) = code_k(c.c₁) * code_k(c.c₂)
+
+function iscss(c::Concat)
+    if iscss(c.c₁) && iscss(c.c₂)
+        true
+    end
+    return nothing # if c.c₁ or c.c₂ are non-CSS; in this case, `Concat(c₁, c₂)` can still be CSS
```
Krastanov, line 46: I added this because I prefer having explicit returns, otherwise the code is a bit more complicated to read.

docs/src/references.md
```diff
@@ -36,6 +36,7 @@ For quantum code construction routines:
 - [chao2018quantum](@cite)
 - [kitaev2003fault](@cite)
 - [fowler2012surface](@cite)
+- [concatenated1996knill](@cite)
```
Krastanov, line 39: the doc CI fails because this is the wrong ref name

src/ecc/codes/concat.jl
```diff
@@ -0,0 +1,47 @@
+"""
+`Concat(c₁, c₂)` is a code concatenation of two quantum codes [concatenated1996knill](@cite).
```
Krastanov, line 2: the doc CI fails because this is the wrong ref name

PR #290, opened 2024-06-14
`[[k + 4, k, 2]]` quantum code

Fe-r-oz:
Implemented [[k + 4, k, 2]] CSS quantum codes, also known as HCode introduced by Cody Jones in his 2013 Paper: [Multilevel distillation of magic states for quantum computing](https://arxiv.org/pdf/1210.3388)

### DIFF
--- a/docs/src/references.bib
+++ b/docs/src/references.bib
@@ -379,3 +379,14 @@
   year={2024},
   publisher={Pearson}
 }
+
+@article{jones2013multilevel,
+  title={Multilevel distillation of magic states for quantum computing},
+  author={Jones, Cody},
+  journal={Physical Review A},
+  volume={87},
+  number={4},
+  pages={042305},
+  year={2013},
+  publisher={APS}
+}
--- a/docs/src/references.md
+++ b/docs/src/references.md
@@ -36,6 +36,7 @@ For quantum code construction routines:
 - [chao2018quantum](@cite)
 - [kitaev2003fault](@cite)
 - [fowler2012surface](@cite)
+- [jones2013multilevel](@cite)
 
 For classical code construction routines:
 - [muller1954application](@cite)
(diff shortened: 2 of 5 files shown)


Fe-r-oz:
Please consider reviewing, @Krastanov. Tested instances via ecc_base. [ECC Zoo](https://errorcorrectionzoo.org/c/quantum_h) has an entry about HCodes! Thanks!

Fe-r-oz:
Decoder gives `max(e...) < noise / 4` even for small instances, k =2 and k= 4. Turning it into Draft. When I tested locally,  the assertion was not happening. 

```
julia> parity_checks(HCode(2))
+ XXXX__
+ XX__XX
+ ZZZZ__
+ ZZ__ZZ

julia> parity_checks(HCode(4))
+ XXXX____
+ XX__XXXX
+ ZZZZ____
+ ZZ__ZZZZ

```

Fe-r-oz:
```
julia> parity_checks(HCode(2))
+ XXXX__
+ XX__XX
+ ZZZZ__
+ ZZ__ZZ

X1X2X3X4II
X1X2IIX5X6
Z1Z2Z3Z4II
Z1Z2IIZ5Z6
```
Authors: Our distillation circuit is based on a family of concatenated quantum codes (H code) that possess a transversal Hadamard operation, enabling each of these codes to distill the eigenstate of the Hadamard operator.

Checking graphs for the smallest HCode whose Stabilizer is shown above. What do these graphs tells us about their decoding performance? There is no error in the code_properties though. But it raises then question, why is their decoding performance like this?


display

display

display

Fe-r-oz:
Based on the graphs shown above, the  `test_ecc_decoder_setups.jl` have bad performance for these codes. These codes seem important for distillation protocols as mentioned in the paper though.

Closing the PR because adding them to `ecc_base` will result in decoding errors,  from `test_ecc_decoder_setups.jl`

If these codes are deemed important, for distillation purpose, please feel free to reopen it. Otherwise, I am not sure whether rebase will help based on their decoding graphs.

Krastanov:
adding codes with low rates or low distances or even codes with good figures of merit but unknown decoders would all be valid and useful additions. They would just need to be tested separately for correctness

Fe-r-oz:
Thank you, I will keep that in mind, we need to add quantum Golay code that has terrible performance with Table decoder :) and these codes! I will make sure to test them each in the future!

PR #293, opened 2024-06-18
Some group theory tools for Quantum error correction

IsaacP1234:
Implemented groupify function to find group of Paulis generated by a generating set.

Implemented get_generating_set function to find minimal generating set of a group of Paulis.

Implemented normalize function to find all Paulis that commute with a given set of paulis.

Implemented center function to find all Paulis within a set that commute with all other elements of that set.

Added united tests for all four functions.


src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+
+
+#
+
+# center
+# normalizer
+# `invert' arbitrary groups
+
+"""
+input: (generating) set of PauliOperator
```
Krastanov, line 16: Check other docstrings in the library, e.g. at http://qc.quantumsavory.org/stable/API/ and follow a similar style.

E.g. a very brief one-line header followed by a paragraph of narative text explaining what is happening. The input and output description can be just part of the text, no need for separate bullets if the function is simple.

src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+julia> groupify(S"XZ XZ")
++ __
++ XZ
++ ZX
++ YY
+```
+"""
+function groupify(s::Stabilizer) # add checks?
+    n = length(s)
+    group = zero(Stabilizer, 2^n, nqubits(s))
```
Krastanov, line 28: Use `Tableau` instead of `Stabilizer` and remove the corresponding TODO.

src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+    if r == 0
+        gs = zero(Stabilizer, 1, nqubits(s))
+        if (1im * zero(Stabilizer, 1, nqubits(s))[1]) in s
+            gs[1] = 1im * gs[1]
+        elseif (-1 * zero(Stabilizer, 1, nqubits(s))[1]) in s
+            gs[1] = -1 * gs[1]
+        end
+        return gs
+    else
+        return s[1:r, :]
```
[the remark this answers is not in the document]
Krastanov: which is fine if given what we said elsewhere about dealing only with real phases

src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+    end
+    return group # TODO use Tableau instead of Stabilizer
+end
+
+
+"""
+For a given non-necessarily-minimal generating set represented as a tableau `s`,
+return the minimal generating set.
+"""
+function get_generating_set(s) # TODO potentially rename to `generator_tableau` or `generating_set`
```
Krastanov, line 44: let's rename it to `minimal_generating_set`

src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+            end
+        end
+    end
+end =#
+
+# s = S"XZ ZX IY"
+# s = groupify(s)
+# println("")
+# print(get_generating_set(s))
+# ##
```
Krastanov, line 101: remove unfinished things from the pull request so that it can be merged

src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+"""
+Returns the full pauli group of the given length. Phases besides + are ignored by default, but can be included by setting phases = true. 
+"""
+function pauligroup(n, phases=false)
+    if phases
+        s = zero(Stabilizer, 4^(n + 1), n)
```
Krastanov, line 107: let's use a `Tableau` here

Any Tableau can be converted to a Stabilizer by calling `Stabilizer(the_tableau)`

src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+
+# s = S"XZ ZX IY"
+# s = groupify(s)
+# println("")
+# print(get_generating_set(s))
+# ##
+"""
+Returns the full pauli group of the given length. Phases besides + are ignored by default, but can be included by setting phases = true. 
+"""
+function pauligroup(n, phases=false)
```
Krastanov, line 105:
```suggestion
function pauligroup(n; phases=false)
```

That way it is a keyword argument instead of a default argument (makes things a bit more legible)

src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+        end
+    end
+
+    return Stabilizer(normalizer)
+end
+
+"""
+For a given s, returns the subset of elements in s that commute with all elements in s.
+"""
+function center(s::Stabilizer)  # hilariously inefficient. centerify? centrify?
```
Krastanov, line 163: how about naming this to `centralizer`?

src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+            end
+        end
+    end
+    return s
+end
+
+"""
+For a given s, find all Pauli elements of the number of qubits that commute with all elements in s
+"""
+function normalize(s::Stabilizer)
```
Krastanov, line 139: How about renaming it to `normalizer`?

src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+    end
+end =#
+
+# s = S"XZ ZX IY"
+# s = groupify(s)
+# println("")
+# print(get_generating_set(s))
+# ##
+"""
+Returns the full pauli group of the given length. Phases besides + are ignored by default, but can be included by setting phases = true.
```
Krastanov, line 103:
```suggestion
Returns the full Pauli group of the given length. Phases besides `+` are ignored by default, but can be included by setting `phases = true`. 
```

src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+            for (j, p) in enumerate(P)
+                s[i, j] = p
+            end
+        end
+    end
+    return s
+end
+
+"""
+For a given s, find all Pauli elements of the number of qubits that commute with all elements in s
```
Krastanov, line 137: not very clear what this says, could you clarify the docstring?

src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+        if commutes == 0
+            push!(normalizer, p)
+        end
+    end
+
+    return Stabilizer(normalizer)
+end
+
+"""
+For a given s, returns the subset of elements in s that commute with all elements in s.
```
Krastanov, line 161:
```suggestion
For a given `s`, return the subset of elements in `s` that commute with all elements in `s`.
```

src/grouptableaux.jl
Krastanov: Should this take `Stabilizer` because it is meant to be about a state, or `Tableau` because it is for any set of Pauli operators or about an arbitrary Pauli subgroup?

src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+
+# scenter = center(s)
+# print(scenter)
+# println(canonicalize!(s)[1:,])
+
+# function center(s::Stabilizer) # centerify? centrify?
+#     #  # returns pairs of non-commuting generators
+# end
+
+# function
```
Krastanov, line 205: clean up

src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+
+
+
+function contract(s::PauliOperator, subset)
```
Krastanov, line 209: probably needs a better name and a docstring

src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+        return nothing
+    end
+end
+
+#= function contract(𝒮::Stabilizer, subset)  # change mathcal notation since not a group?
+    # TODO: contraction on a generating set does not yield a generating set of the contraction of <generating set>!
+    return Stabilizer([P for P in contract.(𝒮, (subset,)) if !isnothing(P)])  # make more efficient
+end =#
+
+function delete(s::PauliOperator, subset)
```
Krastanov, line 222: Should be the standard function `Base.deleteat!`, not a new function `delete`. If it is being implemented, it should probably go in the file where `setindex` and `getindex` are defined.

src/grouptableaux.jl
```diff
@@ -0,0 +1,284 @@
...
+# println(canonicalize!(contracted)[1:4])
+
+# println("---")
+# using QuantumClifford.ECC
+
+# println(canonicalize!(parity_checks(Perfect5())[:,[1, 2, 3, 5, 4]]))
+# println(typeof(groupify(S"XZ ZX")))
+
+
+# # md = MixedDestabilizer(s); groupify(copy(vcat(stabilizerview(md), logicalxview(md), logicalzview(md)))) x
```
Krastanov, line 284: clean up

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@
 
 # News
 
+## v0.9.2
+
+Implemented groupify function to find group of Paulis generated by a generating set
+Implemented get_generating_set function to find minimal generating set of a group of Paulis
+Implemented normalize function to find all Paulis that commute with a given set of paulis
+Implemented center function to find all Paulis within a set that commute with all other elements of that set
+Added united tests for all four functions
```
Krastanov, line 14: You can leave CHANGELOG unchanged to avoid merge conflicts until we are done with the reviews. You probably want to rebase this on master:

```
git checkout master
git pull # potentially with specifying the upstream remote
git checkout some_group_functions
git rebase master
```

Krastanov (changes requested):
Looks like a great start! Let's do some final polish so that we have what is currently available merged, before we worry about more sophisticated functionality. Doing small merges often is much easier for me than doing large pull requests.

Converting to draft for now. Convert back to non-draft and re-request a review when ready.

@KDGoodenough , could you check whether the tests seem reasonable for the functionality you have in mind?

src/grouptableaux.jl
IsaacP1234: Because stabilizer groups are commutative, running this function on a stabilizer would return the input, so I think it should take in a Tableau. As for the output, the output will be commutative, so I assume returning it as a Stabilizer type is correct.

src/grouptableaux.jl
IsaacP1234: Almost all the functions in the API have the one-line header and then some code examples. For simpler functions like groupify and pauligroup, should I include the paragraph portion?

Krastanov: yup, `groupify` will not be simple to someone new to the concept

src/grouptableaux.jl
Krastanov: probably in `paulioperators`

IsaacP1234:
Thank you for the feedback, I think I have implemented it all. 
CHANGELOG left blank to avoid merge conflicts.
Fails JET tests because there are 7 additional warnings, however these are because it interprets indexing a Tableau as a Tableau and not a PauliOperator. I didn't change the limit on warnings in test_jet because that results in a merge conflict. 
I did not rebase this into the master branch of my fork since the pull request seems to work without it and I am not sure what it would help with.

src/grouptableaux.jl
```diff
@@ -0,0 +1,239 @@
...
+Constructs an iterable containing the remaining qubits by taking the difference
+between the given set of qubits and the qubits in the given Stabilizer. Then return 
+the part of the given Stabilizer with the remaining vertical indices.
+
+'''jldoctest
+julia> delete(S"XYZ YZX ZXY", [1,3])
++ Y
++ Z
++ X
+'''
```
Krastanov, line 232:
```
See also: [`traceout!`](@ref)
```

src/grouptableaux.jl
```diff
@@ -0,0 +1,239 @@
...
+end   
+
+"""
+Return the given stabilizer without all the qubits in the given iterable.
+
+Constructs an iterable containing the remaining qubits by taking the difference
+between the given set of qubits and the qubits in the given Stabilizer. Then return 
+the part of the given Stabilizer with the remaining vertical indices.
+
+'''jldoctest
```
Krastanov, line 227: fix to backticks

src/grouptableaux.jl
Krastanov: add a "see also" pointing to this function in `traceout!`

Krastanov (changes requested):
could you let me know what the new jet issues are (just post some screenshots)

IsaacP1234:
> could you let me know what the new jet issues are (just post some screenshots)

image

src/grouptableaux.jl
```diff
@@ -0,0 +1,239 @@
...
++ YY
+```
+"""
+function groupify(s::Stabilizer)
+    n = length(s)
+    group = zero(Tableau, 2^n, nqubits(s))
+    for i in 0:2^n-1
+        for (digit_order, j) in enumerate(digits(i, base=2, pad=n))
+            if j == 1
+                group[i+1] *= s[digit_order]
```
Krastanov, line 24: the JET warning stems from the fact that the compiler can not prove that `group[i+1]` and `s[digit_order]` are both `PauliOperator`. The weird thing is that this seems to be due to the fact that it can not prove that `i` is integer, which seems to be not proven because it can not prove that `n` is an integer, because it does not know that the output of `length` is an integer.

The root solution would be to make sure the compiler knows that `length` can only be an integer (wherever it is defined), but just as a workaround for the moment, I am curious, could you just annotate with `n = length(s)::Int` and see whether that particular JET warning disappears.

src/grouptableaux.jl
```diff
@@ -0,0 +1,239 @@
...
++ X_
++ Z_
++ Y_
+ ⋮
+-iXY
+-iZY
+-iYY
+'''
+"""
+function pauligroup(n; phases=false)
```
Krastanov, line 94: annotate `n::Int` to see whether the compiler and JET will stop complaining

Krastanov (changes requested):
Thanks for taking care of this!
I am just responding to the JET reports for the moment. They seem like false positives, but could you please try the following annotations to help the compiler to not raise these false positives. I will do a final review after we have checked that.

When you are done with this, please also make sure to re-request a code review (should be towards the start of the page on the right) to make sure this goes on my automated reading list.

IsaacP1234:
Adding a see also from trace! to delete_columns causes the docs to fail to build, I think because the project_trace_reset.jl docs get created before the group_tableaux.jl ones. 

Additionally, the build fails, but it seems to be an error with the second part, QuantumSavory, and not QuantumClifford. Let me know if this is not the case and I will debug it.

IsaacP1234:
CHANGELOG:
Implemented  `groupify` function to generate stabilizer group from generating set.
Implemented `minimal_generating_set` function to find the minimal generating set of a Stabilizer.
Implemented `pauligroup` function to generate the full Pauli group of a certain length.
Implemented `normalizer` function to generate all Paulis that commute with a set of Paulis
Implemented `centralizer` function to find a subset of a set of Paulis such that each element in the subset commutes with each element in the set. 
Implemented `contractor` function to find a subset of Paulis in a stabilizer that have an identity operator on a certain qubit. 
Implemented `delete_columns` function to remove the operators corresponding to a certain qubit from all Paulis in a Stabilizer.

IsaacP1234:
@Krastanov re-pinging for review as it is now fully complete but a review request is currently active.

src/grouptableaux.jl
```diff
@@ -0,0 +1,246 @@
...
+    center = typeof(t[1])[]
+    for P in t
+        commutes = 0
+        for Q in t
+            if comm(P, Q) == 0x01
+                commutes = 1
+                break
+            end
+        end
+        if commutes == 0
```
Krastanov, line 183: This does not seem to match the description. I would have expected only `+X` to be returned.

Krastanov (commented):
minor doc cleanup

Krastanov (changes requested):
I made a few documentation changes. Generally, the documentation should say what something is and why it exists. It should say "how" only if there is something particularly non-trivial and only if the previous two questions are already fully addressed. Frequently that "how" might be just in the code comments. If the "how" is mostly obvious from the code, it should probably not even be a comment. The reasoning behind this is two-fold: (1) the user probably does not care exactly how something is implemented, rather they care what it is; (2) the implementation might change without anyone remembering the update the documentation.

I do have one question for clarification on the contract function. Otherwise the rest looks great.

IsaacP1234:
I have removed the groupify functionality from contractor. PR should be all good.

Krastanov (approved):

Krastanov:
Thanks, @IsaacP1234 , this was great!

@KDGoodenough , let's see what you are planning to use this for!

PR #295, opened 2024-06-18
beginning of biased pauli noise for PF sim

amicciche:
This is the code that was mentioned in our Zulip messages. One note, it seemed to be that line 41 in the noise.jl, `    n = nqubits(s)` wasn't doing anything, so I went ahead and deleted that in addition to the functions I added. 

As mentioned over Zulip, because of the error, this code is untested, aside from being able to generate the object from the Julia REPL.


amicciche:
I'm now having this error when I try to use QuantumClifford:

image

Krastanov:
it is fixed now

amicciche:
The `applynoise!()` that takes a `PauliFrame` seems to work correctly, however the way the constructors are written for `PauliError`, one can't do something like the following (which will probably a common enough case to warrant fixing):
image
I tried to fix this by changing Integer to Real on line 37 of noise.jl, but this caused an error. Here:
image

Krastanov:
What error did the change to Real cause?

Krastanov:
marking this as a draft just to make my review queue a bit more manageable. Mark it as finished when ready for review.

amicciche:
> What error did the change to Real cause?

When I change it to Real:
image

I get this error:
image

Krastanov:
Float is a Real so when you do `PauliNoise(x::Real, y::Real, z::Real) = PauliNoise(float(x), float(y), float(z))` you end up with infinite recursion.

A cleaner solution than both what I have written and what you have written is to specify the type for the constructor:
```
julia> struct A{T} a::T end

julia> A(a::Real) = A{Float64}(a)
A

julia> A(1)
A{Float64}(1.0)
```

Or even better, something that does not hardcode Float64

```
julia> struct A{T} a::T end

julia> function A(a::Real) 
           a = float(a)
           return A{typeof(a)}(a)
       end
A

julia> A(1)
A{Float64}(1.0)

julia> A(BigInt(1))
A{BigFloat}(1.0)
```

So the cleanest solution should be something along the lines of
```
function PauliNoise(px::Real, py::Real, pz::Real)
    px, py, pz = float.((px, py, pz))
    px, py, pz = promote(px, py, pz)
    T = typeof(px)
    return PauliNoise{T}(px, py, pz)
end
```

amicciche:
Ah, I see! I'll try to fix this later today, using what you suggested, thanks!

amicciche:
Your code worked as provided! Thanks! With that, I think everything works. I checked for a few codes that doing PauliError(qubit, p) after decoding results in the same logical vs physical error rate plots as PauliError(qubit, p/3, p/3, p/3). For some CSS codes, I also checked that the X and Z logical error corresponding to those simulations also matches doing PauliError(qubit, p*2/3, 0, 0) and PauliError(qubit, 0, 0, p*2/3)  correspondingly (and the other X/Z logical error is 0). Similarly, PauliError(qubit, 0, p*2/3, 0) gives the same performance as PauliError(qubit, p).

src/noise.jl
```diff
@@ -74,6 +99,14 @@ function PauliError(qubits,p)
     NoiseOp(PauliNoise(p), qubits)
 end
 
+function PauliError(q::Int, px, py, pz)
+    NoiseOp(PauliNoise(px,py,pz), [q])
```
Krastanov, line 103:
```suggestion
    NoiseOp(PauliNoise(px,py,pz), (q,))
```

no need to allocate a vector

src/noise.jl
```diff
@@ -37,8 +50,8 @@ function PauliNoise(p)
     UnbiasedUncorrelatedNoise(p)
 end
 
+
```
Krastanov, line 53:
```suggestion
```

spurious whitespace

Krastanov (changes requested):
could you also update the changelog

Krastanov (approved):

Issue #291, opened 2024-06-17
Type conversion error in TableDecoder


royess:
I want to report a type conversion bug in TableDecoder.

It should be reproduced by:
```julia
using QuantumClifford
using QuantumClifford.ECC
import QuantumClifford.ECC: parity_checks, code_k, code_n

struct TestCode end

parity_checks(c::TestCode) = S"""
+ XY__Z_ZZXY_XXYXYX_ZZ
- ZXYZY_X____YZZ_YXZZY
- Y_X__X_Y_Y_ZYYX__Z_Z
+ YZ__XY_X_____Y_XYZ__
+ XZYYYY_ZX_XYZ___YYZX
+ XXXXY__ZZZY_YYYZXZXZ
- X_ZXYXXZ_ZZXZYXZYYYX
+ XXYYYZZYZY_ZY__XZZYY
+ Y_YYXYZZ____XYZ__YY_
- X_ZYZXXY_XZ_XY__XXZX
+ XYY_YYZZ_Z__XZ___XY_
- X_ZXZ_YZZ_ZZZYY_Z_XX
- YYZ__X_Z_Z_YX_XY__ZY
- _YXZXZYZYYZ_ZYX__XZ_
+ _XXX_YYXXZZ_YY_ZZXYZ
- YXXZXZ_X_Z_XZZ_ZZXZX
- XZXXZ___XYXXYXXZZXY_
- Z_XYYYZY_XYZZXXXZYXZ
- XZX_X_Y_Z_ZXYX_Y_X_Z
"""

code_n(c::TestCode) = 20
code_k(c::TestCode) = 1

c = TestCode()

# check it is a valid code
for ps1 in parity_checks(c)
    for ps2 in parity_checks(c)
        @assert comm(ps1, ps2) == 0x0
    end
end

noise = 0.001

d = TableDecoder
s = ShorSyndromeECCSetup(noise, 0)

e = evaluate_decoder(d(c), s, 100000)
```

The error reads:
```
ERROR: LoadError: MethodError: Cannot `convert` an object of type 
  ClassicalXOR{17} to an object of type 
[...]

Issue #245, opened 2024-03-22
Test failure report: Perturbative expansion Purification examples, Comparison to MC


royess:
"Perturbative expansion Purification examples, Comparison to MC" seems to fail randomly. I tried twice, one pass, one failure.

The code I tested is the master branch version [7ff9f39](https://github.com/QuantumSavory/QuantumClifford.jl/commit/7ff9f39bb0cbb6d7a22ce63663b371633158e530), modified in the QEC module, which should not affect this test.

Test log for "Perturbative expansion Purification examples, Comparison to MC":
```
Comparison to MC: Test Failed at /home/yuxuan/Documents/Repos/QuantumClifford.jl/test/test_noisycircuits.jl:70
  Expression: compare(mc, pe, false_success_stat)

Stacktrace:
 [1] macro expansion
   @ /snap/julia/95/share/julia/stdlib/v1.10/Test/src/Test.jl:672 [inlined]
 [2] macro expansion
   @ ~/Documents/Repos/QuantumClifford.jl/test/test_noisycircuits.jl:70 [inlined]
 [3] macro expansion
   @ /snap/julia/95/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [4] macro expansion
   @ ~/Documents/Repos/QuantumClifford.jl/test/test_noisycircuits.jl:57 [inlined]
 [5] macro expansion
   @ /snap/julia/95/share/julia/stdlib/v1.10/Test/src/Test.jl:1577 [inlined]
 [6] top-level scope
   @ ~/Documents/Repos/QuantumClifford.jl/test/test_noisycircuits.jl:56
Test Summary:                                  | Pass  Fail  Total  Time
noisycircuits                                  |   90     1     91  5.0s
  Noisy Gates                                  |    2            2  1.1s
  Monte Carlo Purification examples            |    9            9  0.4s
  Perturbative expansion Purification examples |   11     1     12  2.7s
    Comparison to MC                           |    8     1      9  1.0s
    Symbolic                                   |    3            3  1.7s
  Measurements                                 |   53           53  0.5s
  Classical Bits                               |   15           15  0.1s
ERROR: LoadError: Some tests did not pass: 90 passed, 1 failed, 0 errored, 0 broken.
in expression starting at /home/yuxuan/Documents/Repos/QuantumClifford.jl/test/runtests.jl:52
ERROR: Package QuantumClifford errored during testing
```

royess:
[...]

PR #298, opened 2024-06-20
add random Clifford circuit codes

royess:
Add implementations of random Clifford circuit codes, whose connectivity can be all-to-all or brickwork in some dimensions.

Some notes:
1. While preparing this PR, I found that [this notebook](https://github.com/QuantumSavory/QuantumClifford.jl/blob/master/docs/src/notebooks/Stabilizer_Codes_Based_on_Random_Circuits.ipynb) have implemented a 1D brickwork random circuit codes. This PR provides more connectivity choices and uses `AbstractOperations[]` of sparse gates for the generated circuit, which should be more efficient and allow visualization.
2. I do not include random circuit codes in `ecc_decoder_all_setups` in this PR. The reason is that to ensure a good code parameter, the code has to contain more than 10 qubits, which exceeds the size that `TableDecoder` can efficiently handle. Random circuit codes are not CSS and are currently unsupported by `PyMatchingDecoder`.
3. For a reference, I did some more tests in [royess/qecc-devlog/test_random_circuit.ipynb](https://github.com/royess/qecc-devlog/blob/master/test_random_circuit.ipynb), including visualizing the encoding circuits and decoding via TableDecoder. In the future, we may consider including these contents as tutorials or documentation.


src/ecc/codes/random_circuit.jl
```diff
@@ -0,0 +1,93 @@
...
+
+"""
+Random Clifford circuit code.
+
+The code is generated by a random Clifford circuit `circ` that encode a subset of qubits `encode_qubits` into logical qubits.
+The connectivity of the random circuit can be either all-to-all [brown2013short](@cite) or brickwork in some dimensions [gullans2021quantum](@cite).
+Each gate in the random circuit is a random 2-qubit Clifford gate.
+"""
+struct RandomCircuitCode <: AbstractECC
+    arrange::NTuple{N,Int} where {N} # the spatial arrangement of qubits
```
Krastanov, line 12: you can turn these comments into per-field docstrings (julia supports per-field docstrings and they render better in various introspection tools).

src/ecc/codes/random_circuit.jl
```diff
@@ -0,0 +1,93 @@
...
+    connect::Union{Val{:alltoall},Val{:brickwork}} # the gate connectivity
+    circ::Vector{QuantumClifford.AbstractOperation} # the random generated encoding circuit
+    encode_qubits::AbstractArray # the qubits to be encoded
+    # TODO it will be nicer if we can use CartesianIndex for encode_qubits here,
+    # but its conversion to LinearIndex is limited, not supporting non-one step.
+
+    function RandomCircuitCode(rng::AbstractRNG, n::Int, connect::Val{:alltoall}, ngates::Int, k::Int)
```
Krastanov, line 19: I think I prefer if all these inner constructors become normal constructors. It is valuable to still make it possible to access something like the default constructor, even if that constructor is undocumented and non-public.

But maybe I am missing some other reason to do that, just let me know.

src/ecc/codes/random_circuit.jl
```diff
@@ -0,0 +1,93 @@
...
+    function RandomCircuitCode(rng::AbstractRNG, arrange::NTuple{N,Int} where {N}, connect::Val{:brickwork}, nlayers::Int, encode_qubits::AbstractArray)
+        new(arrange, Val(:brickwork), random_brickwork_clifford_circuit(rng, arrange, nlayers), encode_qubits)
+    end
+
+    function RandomCircuitCode(arrange::NTuple{N,Int} where {N}, connect::Val{:brickwork}, nlayers::Int, encode_qubits::AbstractArray)
+        new(arrange, Val(:brickwork), random_brickwork_clifford_circuit(arrange, nlayers), encode_qubits)
+    end
+end
+
+function random_brickwork_clifford_circuit(rng::AbstractRNG, arrange::NTuple{N,Int} where {N}, nlayers::Int)
```
Krastanov, line 44: Do we want to support periodic and non-periodic boundary conditions?

src/ecc/codes/random_circuit.jl
```diff
@@ -0,0 +1,93 @@
...
+"""
+Random Clifford circuit code.
+
+The code is generated by a random Clifford circuit `circ` that encode a subset of qubits `encode_qubits` into logical qubits.
+The connectivity of the random circuit can be either all-to-all [brown2013short](@cite) or brickwork in some dimensions [gullans2021quantum](@cite).
+Each gate in the random circuit is a random 2-qubit Clifford gate.
+"""
+struct RandomCircuitCode <: AbstractECC
+    arrange::NTuple{N,Int} where {N} # the spatial arrangement of qubits
+    connect::Union{Val{:alltoall},Val{:brickwork}} # the gate connectivity
```
Krastanov, line 13: Do we want to support 2d brickworks (the qubits are layed out in a square lattice instead of a line)?

royess:
@Krastanov, please help review this PR; thanks! (The failed CI check is related to #245 )

src/ecc/codes/random_circuit.jl
```diff
@@ -0,0 +1,93 @@
...
+        new((n,), Val(:alltoall), random_all_to_all_clifford_circuit(n, ngates), encode_qubits)
+    end
+
+    function RandomCircuitCode(rng::AbstractRNG, arrange::NTuple{N,Int} where {N}, connect::Val{:brickwork}, nlayers::Int, encode_qubits::AbstractArray)
+        new(arrange, Val(:brickwork), random_brickwork_clifford_circuit(rng, arrange, nlayers), encode_qubits)
+    end
+
+    function RandomCircuitCode(arrange::NTuple{N,Int} where {N}, connect::Val{:brickwork}, nlayers::Int, encode_qubits::AbstractArray)
+        new(arrange, Val(:brickwork), random_brickwork_clifford_circuit(arrange, nlayers), encode_qubits)
+    end
```
Krastanov, line 41: I am not super comfortable that all these constructors are nondeterministic. In my mind, a constructor is supposed to take some data and organize it in a nice datastructure.

I think these should not be constructors, rather `rand` methods. To make a random Int, you do not call the constructor Int(). You call `rand(Int)`.

On the other hand here you basically need `rand(RandomCircuitCode, params...)`, which is not the `rand` API either.

I guess this means that we need `random_circuit_code` function.

What are your thoughts on this?

Krastanov (changes requested):
This is great! You mentioned how it does not work well with many decoders. How does it perform with belief propagation and OSD decoders?

src/ecc/codes/random_circuit.jl
royess: The 2d brickworks (or higher) are supported and can be constructed by passing `arrange=(5,5)`, for example. I think I should make this better documented.

src/ecc/codes/random_circuit.jl
royess: It will be easy technically.

I am unsure whether adding more options to `RandomCircuitCode::connect` will make the code a bit clumsy.

src/ecc/codes/random_circuit.jl
royess: I also think using a `random_circuit_code` will be better.

Further, we may also consider decoupling between (1) generating a random circuit and (2) constructing a code from a given encoding circuit. In this way, part (1) should be put outside the ECC submodule; part (2) can be a function that converts an encoding circuit to a code of type `AbstractECC`.

In this way, a struct type `RandomCircuitCode` also does not seem necessary. The public API will be `random_circuit(...)` and `AbstractECC(circ, encoding_qubits)`.

Will this be a better design?

royess: A downside of removing `RandomCircuitCode` is that one can no longer get the circuit from such a code, which is sometimes convenient.

Alternatively, we may still have such a type with a maybe better name, e.g., `CircuitCode`.

Krastanov: I do not think we should remove the struct, just the randomness in the constructor. `CircuitCode` sounds good.

royess:
> You mentioned how it does not work well with many decoders.

Actually, `TableDecoder` can give a rather low logical error rate, as I tested in [this notebook](https://github.com/QuantumSavory/QuantumClifford.jl/blob/master/docs/src/notebooks/Stabilizer_Codes_Based_on_Random_Circuits.ipynb).

However, we need random circuit codes with more than 10 qubits, e.g., 20 qubits, to ensure the randomly generated codes have a distance larger than 0. This will make `TableDecoder` inefficient and trigger warnings due to #291.

> How does it perform with belief propagation and OSD decoders?

BP and BP-OSD in `PyQDecoders.jl` (as well as earlier mentioned `PyMatchingDecoder` ) are not applicable to random circuit codes because they are non-CSS.

```
"Codes of type CircuitCode do not have separate X and Z parity checks, either because they are not a CSS code and thus inherently do not have separate checks, or because its separate checks are not yet implemented in this library."

Stacktrace:
 [1] parity_checks_x(code::CircuitCode)
   @ QuantumClifford.ECC ~/Documents/Repos/QuantumClifford.jl/src/ecc/ECC.jl:41
 [2] QuantumCliffordPyQDecodersExt.PyBeliefPropOSDecoder(c::CircuitCode; maxiter::Nothing, bpmethod::Nothing, errorrate::Nothing, osdmethod::Nothing, osdorder::Int64)
   @ QuantumCliffordPyQDecodersExt ~/Documents/Repos/QuantumClifford.jl/ext/QuantumCliffordPyQDecodersExt/QuantumCliffordPyQDecodersExt.jl:51
 [3] QuantumCliffordPyQDecodersExt.PyBeliefPropOSDecoder(c::CircuitCode)
   @ QuantumCliffordPyQDecodersExt ~/Documents/Repos/QuantumClifford.jl/ext/QuantumCliffordPyQDecodersExt/QuantumCliffordPyQDecodersExt.jl:50
 [4] PyBeliefPropOSDecoder(args::CircuitCode; kwargs::@Kwargs{})
   @ QuantumClifford.ECC ~/Documents/Repos/QuantumClifford.jl/src/ecc/decoder_pipeline.jl:297
 [5] PyBeliefPropOSDecoder(args::CircuitCode)
   @ QuantumClifford.ECC ~/Documents/Repos/QuantumClifford.jl/src/ecc/decoder_pipeline.jl:292
 [6] top-level scope
   @ ./In[27]:14
```

src/ecc/codes/random_circuit.jl
Krastanov: Sounds good, no need to change anything here, I like your approach. A bit more documentation would be helpful.

src/ecc/codes/random_circuit.jl
```diff
@@ -0,0 +1,74 @@
...
+- `arrange`: the spatial arrangement of qubits
+- `connect`: the gate connectivity, can be `alltoall` or `brickwork`
+- `circ`: the encoding circuit
+- `encode_qubits`: the qubits to be encoded
+
+Qubit arrangement for different connectivity types:
+
+- For `all-to-all` connectivity, `arrange` is a one-element tuple containing the number of qubits.
+- For `brickwork` connectivity, the qubits are arranged as a lattice, and `arrange` contains side length in each dimension.
+  For example, a 5×5 lattice will have `arrange = (5, 5)`.
```
Krastanov, line 16: Does this imply that the `connect` and `arrange` arguments are "correlated"? If that is the case, it would be better if it is a single argument. We can just stick to `arrange` maybe call it something like `lattice_dims`.

src/ecc/codes/random_circuit.jl
```diff
@@ -0,0 +1,74 @@
...
+
+TODO it would be nicer if we can use CartesianIndex for `encode_qubits` in brickworks,
+but its conversion to LinearIndex is limited, not supporting non-one step.
```
Krastanov, line 19: TODOs should not be visible in the documentation.

Also, this TODO would be a breaking change if adopted, right? We should not merge things that already have the expectation that there will be a breaking change to them in the future.

src/randoms.jl
```diff
@@ -228,3 +228,43 @@ function fill_tril(rng, matrix, n; symmetric::Bool=false)
     end
     matrix
 end
+
+##############################
+# Random circuit
+##############################
+
+function random_brickwork_clifford_circuit(rng::AbstractRNG, arrange::NTuple{N,Int} where {N}, nlayers::Int)
+    circ = QuantumClifford.AbstractOperation[]
```
Krastanov, line 237: Can this list be more concretely typed?

test/test_ecc_base.jl
```diff
@@ -5,12 +5,23 @@ using InteractiveUtils
...
-    Toric => [(3,3), (4,4), (3,6), (4,3), (5,5)],
-    Surface => [(3,3), (4,4), (3,6), (4,3), (5,5)],
+    Toric => [(3, 3), (4, 4), (3, 6), (4, 3), (5, 5)],
+    Surface => [(3, 3), (4, 4), (3, 6), (4, 3), (5, 5)],
     Gottesman => [3, 4, 5],
-    CSS => (c -> (parity_checks_x(c), parity_checks_z(c))).([Shor9(), Steane7(), Toric(4,4)]),
-    Concat => [(Perfect5(), Perfect5()), (Perfect5(), Steane7()), (Steane7(), Cleve8()), (Toric(2,2), Shor9())],
+    CSS => (c -> (parity_checks_x(c), parity_checks_z(c))).([Shor9(), Steane7(), Toric(4, 4)]),
+    Concat => [(Perfect5(), Perfect5()), (Perfect5(), Steane7()), (Steane7(), Cleve8()), (Toric(2, 2), Shor9())],
+    CircuitCode => random_circuit_code_args
```
Krastanov, line 24: The formatting change here is not bad, but I would prefer if it is not done -- it makes reading `git blame` more difficult

src/ecc/codes/random_circuit.jl
Krastanov: nope, forget this, I misunderstood what you had written.

However, I am still not a very big fan of `arrange` because it sounds to me like the verb, not a contraction of `arrangement`. I see you use the word `dim` for something else. How about `lattice_size`?

Krastanov: But now I have also a completely separate problem. Don't `circ` and `encode_qubits` completely define the code? No need for `arrange` and for `connect`? The documentation of `random_circuit_code` can instead be more detailed and there should be a `See also: ` ref here.

Krastanov (changes requested):
a few more minor stylistic changes

src/ecc/codes/random_circuit.jl
royess: > However, I am still not a very big fan of `arrange` because it sounds to me like the verb, not a contraction of `arrangement`. I see you use the word `dim` for something else. How about `lattice_size`?

I agree with these new names, which are more clear.

> Don't `circ` and `encode_qubits` completely define the code? No need for `arrange` and for `connect`?

That is true.

I kept them thinking that `arrange` and `connect` may help in some potential future features. For example, we may want to accept `arrange` in CartesianIndex, as mentioned in the TODO below. But currently, they look redundant and should indeed better be moved.

src/randoms.jl
royess: Do you mean by we should have `arrange::NTuple{N,Int} where {N<:Integer}`?

Krastanov: no, the list just does not need to be an abstract operation list. You are creating only one type of gates here.

Krastanov:
What is the current status?

royess:
I did not fully comprehend some of your previous review comments and was waiting for your further replies on them.

I will get this PR updated quickly.

src/ecc/codes/random_circuit.jl
```diff
@@ -0,0 +1,74 @@
...
+
+# TODO it would be nicer if we can use CartesianIndex for `encode_qubits` in brickworks,
+# but its conversion to LinearIndex is limited, not supporting non-one step.
+function random_circuit_code(rng::AbstractRNG, lattice_size::NTuple{N,Int} where {N}, connect::Val{:brickwork}, nlayers::Int, encode_qubits::AbstractArray)
+    CircuitCode(prod(lattice_size), random_brickwork_clifford_circuit(rng, lattice_size, nlayers), encode_qubits)
+end
+
+function random_circuit_code(lattice_size::NTuple{N,Int} where {N}, connect::Val{:brickwork}, nlayers::Int, encode_qubits::AbstractArray)
+    CircuitCode(prod(lattice_size), random_brickwork_clifford_circuit(lattice_size, nlayers), encode_qubits)
+end
```
Krastanov, line 74: I am not very comfortable with all these different methods. The brickwork and the alltoall methods barely have any overlapping signatures, at which point it does not really make sense to keep them as methods of the same function. I think it would be cleaner and less confusing to simply have two separate functions, the way you have two separate functions for generating random circuits.

src/randoms.jl
```diff
@@ -228,3 +228,43 @@ function fill_tril(rng, matrix, n; symmetric::Bool=false)
     end
     matrix
 end
+
+##############################
+# Random circuit
+##############################
+
+function random_brickwork_clifford_circuit(rng::AbstractRNG, arrange::NTuple{N,Int} where {N}, nlayers::Int)
```
Krastanov, line 236: needs a docstring if it is exported

src/randoms.jl
```diff
@@ -228,3 +228,43 @@ function fill_tril(rng, matrix, n; symmetric::Bool=false)
...
+                push!(circ, SparseGate(random_clifford(rng, 2), [j, k]))
+            end
+        end
+    end
+    circ
+end
+
+random_brickwork_clifford_circuit(arrange::NTuple{N,Int} where {N}, nlayers::Int) = random_brickwork_clifford_circuit(GLOBAL_RNG, arrange, nlayers)
+
+function random_all_to_all_clifford_circuit(rng::AbstractRNG, nqubits::Int, ngates::Int)
```
Krastanov, line 260: needs a docstring if it is exported

src/ecc/codes/random_circuit.jl
```diff
@@ -0,0 +1,74 @@
...
+The connectivity of the random circuit can be either all-to-all [brown2013short](@cite) or brickwork in some dimensions [gullans2021quantum](@cite).
+Each gate in the random circuit is a random 2-qubit Clifford gate.
+
+For `brickwork` connectivity, the qubits are arranged as a lattice, and `lattice_size` contains side length in each dimension.
+For example, a chain of length five will have `lattice_size = (5,)`, and a 5×5 lattice will have `lattice_size = (5, 5)`.
+
+In multi-dimensional cases, gate layers act alternatively along each direction.
+The nearest two layers along the same direction are offset by one qubit, forming a so-called brickwork.
+The boundary condition is chosen as open.
+"""
```
Krastanov, line 47: probably should have a `see also` section to the random circuit functions and to the struct

src/ecc/codes/random_circuit.jl
```diff
@@ -0,0 +1,74 @@
...
+struct CircuitCode <: AbstractECC
+    n::Int
+    circ::Vector{QuantumClifford.AbstractOperation}
+    encode_qubits::AbstractArray
+end
+
+iscss(::Type{CircuitCode}) = nothing
+
+code_n(c::CircuitCode) = prod(c.n)
```
Krastanov, line 21: I do not think the `prod` is necessary

Krastanov (changes requested):
looks good, but there are a few stylistic changes that I would like, to make it a bit more in the style of the rest of the library

Krastanov:
Wonderful! Thanks for the addition

PR #312, opened 2024-07-10
Lifted and lifted product codes

royess:
Add lifted codes, lifted product (LP) codes, and group ring functionality, which is essential for these codes.

**Note:** The added LP codes trigger errors in syndrome circuits, as reported in #306. So, decoding tests only include commutativity checks at the moment. For the same reason, I also skip LP codes in `all_testablable_code_instances`.


royess:
- The failure of the downgrade test should be related to Nemo's degrading from 0.4.5 to 0.4.2.

  In Nemo 0.4.5, `FqFieldElem` is a subtype of `RingElement`, which matches the following function call in my codes.
  
  ```julia
  base_ring_type(::Type{PermGroupRing{T}}) where {T<:RingElement}
  ```
  
  When tested with Nemo 0.4.2, this function call breaks. Possibly it is because `FqFieldElem` was not a subtype of `RingElement` back then.

- The nighly tests fail because of a bug in the external `PyBeliefPropOSDecoder`, which looks irrelevant to the added implementation.

    ```
    belief prop decoders, good for sparse codes: Error During Test at /home/runner/work/QuantumClifford.jl/QuantumClifford.jl/test/test_ecc_decoder_all_setups.jl:89
      Got exception outside of a @test
      Python: Julia: MethodError: no method matching pyjlarray_array__array(::Nothing)
      The function `pyjlarray_array__array` exists, but no method is defined for this combination of argument types.
      
      Closest candidates are:
        pyjlarray_array__array(::AbstractArray)
         @ PythonCall ~/.julia/packages/PythonCall/S5MOg/src/JlWrap/array.jl:269
      
      Stacktrace:
        [1] _pyjl_callmethod(f::Any, self_::Ptr{PythonCall.C.PyObject}, args_::Ptr{PythonCall.C.PyObject}, nargs::Int64)
          @ PythonCall.JlWrap ~/.julia/packages/PythonCall/S5MOg/src/JlWrap/base.jl:61
        [2] _pyjl_callmethod(o::Ptr{PythonCall.C.PyObject}, args::Ptr{PythonCall.C.PyObject})
          @ PythonCall.JlWrap.Cjl ~/.julia/packages/PythonCall/S5MOg/src/JlWrap/C.jl:63
        [3] PyObject_CallObject(x1::Ptr{PythonCall.C.PyObject}, x2::Ptr{PythonCall.C.PyObject})
          @ PythonCall.C ~/.julia/packages/PythonCall/S5MOg/src/C/pointers.jl:297
        [4] macro expansion
          @ ~/.julia/packages/PythonCall/S5MOg/src/Core/Py.jl:132 [inlined]
        [5] pycallargs(f::PythonCall.Core.Py, args::PythonCall.Core.Py)
          @ PythonCall.Core ~/.julia/packages/PythonCall/S5MOg/src/Core/builtins.jl:212
        [6] pycall(f::PythonCall.Core.Py, args::Vector{Bool}; kwargs::@Kwargs{})
          @ PythonCall.Core ~/.julia/packages/PythonCall/S5MOg/src/Core/builtins.jl:230
        [7] pycall
          @ ~/.julia/packages/PythonCall/S5MOg/src/Core/builtins.jl:220 [inlined]
        [8] (::PythonCall.Core.Py)(args::Vector{Bool})
          @ PythonCall.Core ~/.julia/packages/PythonCall/S5MOg/src/Core/Py.jl:339
        [9] decode(d::QuantumCliffordPyQDecodersExt.PyBeliefPropOSDecoder, syndrome_sample::SubArray{Bool, 1, Matrix{Bool}, Tuple{Int64, Base.Slice{Base.OneTo{Int64}}}, true})
          @ QuantumCliffordPyQDecodersExt ~/work/QuantumClifford.jl/QuantumClifford.jl/ext/QuantumCliffordPyQDecodersExt/QuantumCliffordPyQDecodersExt.jl:74
       [10] batchdecode(d::QuantumCliffordPyQDecodersExt.PyBeliefPropOSDecoder, syndrome_samples::Matrix{Bool})
          @ QuantumClifford.ECC ~/work/QuantumClifford.jl/QuantumClifford.jl/src/ecc/decoder_pipeline.jl:20
    ```

royess:
@Krastanov please help review this PR, thanks!

src/ecc/codes/classical/lifted.jl
```diff
@@ -0,0 +1,53 @@
...
+
+code_n(c::LiftedCode) = size(c.A, 2) * size(c.repr(parent(c.A[1,1])(0)), 2)
+
+function mod2rank(h::Matrix{<:Integer})
+    Z2, _ = residue_ring(ZZ, 2)
+    S = matrix_space(Z2, size(h)...)
+    rank(S(h))
+end
+
+code_s(c::LiftedCode) = mod2rank(parity_checks(c)) # note that redundant rows exist in general
```
Krastanov, line 51: As you noted, the redundancy of rows seems to be causing problems with other routines. Either we should have the redundant rows removed or we should fix the other routines to work in the presence of redundant rows (and raise a warning as already done in a couple of places).

src/ecc/codes/lifted_product.jl
```diff
@@ -0,0 +1,58 @@
...
+ default to be [`permutation_repr`](@ref) for GF(2)-algebra.
+
+A lifted product code is constructed by hypergraph product of the two lifted codes `c₁` and `c₂`.
+Here, the hypergraph product is taken over a group ring, which serves as the base ring for both lifted codes.
+After the hypergraph product, the parity-check matrices are lifted by `repr`.
+
+Children of code families:
+- Hypergraph product code: use trivial group ring `PermutationGroupRing(GF(2), 1)` in lifted codes.
+- Two-block group-algebra (2GBA) code: choose 1×1 base matrices for `c₁` and `c₂`.
+
```
Krastanov, line 19: Could you give more details in all of this, so it is unambiguously clear to the reader what is happening.

"Children of code families" is a bit nonstandard of a phrasing. May we change it to something like "Examples of code subfamilies" -- specifically, I want to distinguish this from "instance of a code from this family" which is what "children" sounds like to me.

src/ecc/group_ring.jl
```diff
@@ -0,0 +1,293 @@
+import Base: *, +, -, ==, deepcopy_internal, show, isone, iszero, one, zero, adjoint
+
+import Nemo: Ring, RingElem, RingElement, NCRing, NCRingElem, NCRingElement, Perm, CacheDictType,
+    @attributes, base_ring, base_ring_type, elem_type, get_cached!,
+    is_domain_type, is_exact_type, parent, parent_type
```
Krastanov, line 5: The `CacheDictType` and `get_cached!` are not public and I am very reluctant to depend on them.

Is there a public interface that is equivalent in Nemo?

What is the reason the Nemo folks do not have this permutation object not included in their library?

Is there an alternative that is more idiomatic for them?

Have you spoken with the Nemo folk? Would the accept a contribution like this in their library?

What would your code look like if you do not use this custom type and rather use only what is available in Nemo?

src/ecc/group_ring.jl
```diff
@@ -0,0 +1,293 @@
...
+    end
+
+    function PermGroupRingElem{T}() where {T<:RingElement}
+        return new{T}(Dict{Perm,T}())
+    end
+end
+
+# Data type and parent object methods
+
+parent_type(::Type{PermGroupRingElem{T}}) where {T<:RingElement} = PermGroupRing{T}
```
Krastanov, line 90: A lot of functionality in this file is not tested and I am not a Nemo expert so this makes me pretty worried about the long-term health of this part of the library. Does Nemo have some form of interface test suite we can use? Or can we remove functionality we are not using?

Another way to think of this: is there a simpler, more direct way to do this, even if it does not provide for a very elegant abstraction?

src/ecc/group_ring.jl
```diff
@@ -0,0 +1,293 @@
...
+    parent(f) == R || error("Unable to coerce a group ring element")
+    return f
+end
+
+"""
+Permutation group ring constructor.
+
+See also: [`PermutationGroupRing`](@ref).
+"""
+function PermutationGroupRing(R::Ring, l::Int, cached::Bool=true)
```
Krastanov, line 275: this docstring does not seem to be necessary (it does not really say convey new knowledge)

test/test_ecc_decoder_all_setups.jl
```diff
@@ -33,29 +36,78 @@ include("test_ecc_base.jl")
...
+    codes = vcat(LP04, LP118, other_lifted_product_codes)
 
     noise = 0.001
 
     setups = [
         CommutationCheckECCSetup(noise),
-        NaiveSyndromeECCSetup(noise, 0),
-        ShorSyndromeECCSetup(noise, 0),
+        # NaiveSyndromeECCSetup(noise, 0),
+        # ShorSyndromeECCSetup(noise, 0),
```
Krastanov, line 97: these should be re-enabled before we can merge this

test/test_ecc_decoder_all_setups.jl
```diff
@@ -33,29 +36,78 @@ include("test_ecc_base.jl")
...
     ]
+    # lifted product codes currently trigger errors in syndrome circuits
```
Krastanov, line 99: it is something related to the redundant rows. Just check whether the number of rows and the size of the matrix that go in `naive_encoder` are correct. If not, just canonicalize the matrix and raise a warning, similarly to what is done in other parts of the codebase

test/test_ecc_decoder_all_setups.jl
```diff
@@ -33,29 +36,78 @@ include("test_ecc_base.jl")
...
+    push!(LP04, LPCode(LiftedCode(A), LiftedCode(A)))
+end
+
+LP118 = []
+for l in keys(B118)
+    A = map(B118[l]) do x
+        (PermutationGroupRing(GF(2), l))(cyclic_permutation(x, l))
+    end
+    push!(LP118, LPCode(LiftedCode(A), LiftedCode(A)))
+end
```
Krastanov, line 68: A user would not know how to do this given the current API and documentation.

I am worried that the current level of abstraction does not make it obvious how to actually create a useful code with this library. That should be documented (and potentially simplified).

It is perfectly valid to say "in this or that paper they used some proprietary search method to find a good matrix" or "picking a sparse matrix at random will probably be a good choice" or "picking a good base matrix is a difficult problem of its own", but without something like that it would be very difficult for a user to figure out how to do something productive.

test/test_ecc_decoder_all_setups.jl
```diff
@@ -33,29 +36,78 @@ include("test_ecc_base.jl")
...
+
+other_lifted_product_codes = []
+
+# from https://arxiv.org/abs/2202.01702v3
```
Krastanov, line 72: Actually, these two paper examples you have should probably just become part of the docstring

src/ecc/codes/classical/lifted.jl
```diff
@@ -0,0 +1,53 @@
...
+    # such as residue_ring(ZZ, 2), which is mathematically equivalent, and residue_ring(ZZ, n) for n > 2
+    !(characteristic(base_ring(A[1,1])) == 2) && error("The default permutation representation applies only to GF(2) group algebra")
+    LiftedCode(A, permutation_repr)
+end
+
+"""
+Represent a permutation group ring element by mapping permutations to circulant matrices.
+"""
+function permutation_repr(x::PermGroupRingElem{FqFieldElem})
+    return sum([Int(lift(ZZ, x.coeffs[k])) .* Array(matrix_repr(k)) for k in keys(x.coeffs)], init=zeros(Bool, parent(x).l, parent(x).l))
```
Krastanov, line 34: I know this type of requests are annoying, but could you write this in a simpler form or provide inline comments. This is a bit too smart.

src/ecc/group_ring.jl
```diff
@@ -0,0 +1,293 @@
...
+        end::PermGroupRing{T}
+    end
+end
+
+const PermGroupRingElemID = CacheDictType{Tuple{Ring, Int}, NCRing}()
+
+"""
+Element of a [`PermGroupRing`](@ref).
+
+- `coeffs`: A dictionary of permutations and their coefficients. Empty dictionary represents zero.
```
Krastanov, line 69: define in more details what this is

src/ecc/group_ring.jl
```diff
@@ -0,0 +1,293 @@
...
+"""
+Permutation [group ring](https://en.wikipedia.org/wiki/Group_ring) over a base ring, also known as group algebra.
+
+- `base_ring`: The base ring, whose elements are used as coefficients of permutations.
+- `l`: The length of permutations.
+
+Basic usage:
+
+- To construct a permutation group ring, use [`PermutationGroupRing`](@ref).
+- To construct an element of a permutation group ring, use the call syntax of the parent object.
```
Krastanov, line 17: Clarify this line.

src/ecc/group_ring.jl
```diff
@@ -0,0 +1,293 @@
...
+julia> R = PermutationGroupRing(GF(2), 3)
+Permutation group ring over Prime field of characteristic 2
+
+julia> f0 = R(0)
+Dict{Perm, Nemo.FqFieldElem}()
+
+julia> f1 = R(1)
+Dict{Perm{Int64}, Nemo.FqFieldElem}(() => 1)
+
+julia> f2 = R(Perm([1,3,2]))
```
Krastanov, line 32: clarify what `Perm` is

src/ecc/group_ring.jl
```diff
@@ -0,0 +1,293 @@
...
+- To perform arithmetic operations, use the standard arithmetic operators.
+
+```jldoctest
+julia> ENV["NEMO_PRINT_BANNER"] = "false"; using Nemo: GF, Perm
+
+julia> R = PermutationGroupRing(GF(2), 3)
+Permutation group ring over Prime field of characteristic 2
+
+julia> f0 = R(0)
+Dict{Perm, Nemo.FqFieldElem}()
```
Krastanov, line 27: Why is this printing as a dictionary?

src/ecc/group_ring.jl
```diff
@@ -0,0 +1,293 @@
...
+
+julia> f1 = R(1)
+Dict{Perm{Int64}, Nemo.FqFieldElem}(() => 1)
+
+julia> f2 = R(Perm([1,3,2]))
+Dict{Perm{Int64}, Nemo.FqFieldElem}((2,3) => 1)
+
+julia> f3 = R(Dict(Perm(3) => 1, Perm([3,1,2]) => 1))
```
Krastanov, line 35: what it means to call this on a dictionary

src/ecc/codes/lifted_product.jl
```diff
@@ -0,0 +1,58 @@
+import Nemo: matrix_space
+
+
+"""
+Lifted product codes [panteleev2021degenerate](@cite) [panteleev2022asymptotically](@cite)
```
Krastanov, line 5: I think a lot of my concerns expressed elsewhere can be addressed by being a bit more detailed with what these are, something like an extra paragraph expounding upon: these are the lifted product codes (cite) which are a generalization of the general bicycle codes (cite) which are a generalization of the bicycle code (cite).

Then we can have something like: the bicycle code is defined by such and such set and if you have that set you can convert it to the necessary representation and use it here by doing so and so.

The generalized bicycle code is defined by a circulant matrix and you can use a circulant matrix to create a code here by doing yada yada

Finally, the most general form is yada yada where you need to provide so and so.

One of my worries is that currently I do not know how to use what you have implemented to set up the generalized bicycle codes from https://quantum-journal.org/papers/q-2021-11-22-585/pdf/

See footnote 2 on page 2 of https://quantum-journal.org/papers/q-2021-11-22-585/pdf/

src/ecc/codes/classical/lifted.jl
```diff
@@ -0,0 +1,53 @@
+import Nemo: characteristic, lift, matrix_repr
+
+"""
+Classical codes lifted over a permutation group ring [panteleev2021degenerate](@cite) [panteleev2022asymptotically](@cite).
+
+- `A::Matrix`: the base matrix of the code, whose elements are in a permutation group ring.
```
Krastanov, line 6: Looking at the end of this list of changes, it seems like you are imagining the user doing a couple of different things to generate codes, but it does not seem like the starting point is really this matrix.

Can we structure it in a way that is simpler, so that the user just gives what would be the natural object for them to work with. Especially looking at the test file, it does not seem like you expect most users to start with such a matrix. Here are a few suggestions, probably some of them are bad, maybe all of them are bad, but some discussion of making this API easier to use is needed:

- maybe a constructor that just takes a dictionary of permutation and does all the conversions behind the scene
- a constructor that just takes the circulant matrices or maybe a separate struct for simpler instances of these codes (bicycle codes and generalized bicycle codes)

src/ecc/codes/classical/lifted.jl
```diff
@@ -0,0 +1,53 @@
...
+- `repr::Function`: a function that converts a permutation group ring element to a matrix;
+  default to be [`permutation_repr`](@ref) for GF(2)-algebra.
+
+The parity-check matrix is constructed by applying `repr` to elements of `A`, known as a lift.
+
+See also: [`LPCode`](@ref), [`PermGroupRing`](@ref).
+"""
+struct LiftedCode <: ClassicalCode
+    A::Matrix{PermGroupRingElem}
+    repr::Function
```
Krastanov, line 16: You are already hardcoding the use of your `PermGroupRingElem`, so we can probably be a bit neater about `repr`. Instead of asking for a repr here, you can ask for a new method of permutation_repr that specializes on the group type.

Krastanov (changes requested):
This is shaping up neatly, thank you!

I left in a lot of comments but the overarching concerns are:

- most importantly, better documentation, and potentially reworking the user-facing API, as currently a user will have a lot of difficulty creating a code
- a lot of custom code that might break if Nemo has a breaking release (they do that often)
- better test coverage

For documenting fields of structs, I would suggest using the docstringhelpers package. See for instance here: https://github.com/QuantumSavory/QuantumSavory.jl/blob/master/src/ProtocolZoo/switches.jl#L139-L157

src/ecc/codes/classical/lifted.jl
royess: We cannot "have the redundant rows removed." Because in lifted product construction, redundancy in each lifted classical codes is needed in many cases. Otherwise, the dimension of the resulting LP code will be zero or negative.

I think we should raising a warning instead.

royess: We cannot "have the redundant rows removed." Because in lifted product construction, redundancy in each lifted classical codes is needed in many cases. Otherwise, the dimension of the resulting LP code will be zero or negative.

I think we should raising a warning instead.

src/ecc/codes/classical/lifted.jl
```diff
@@ -0,0 +1,53 @@
+import Nemo: characteristic, lift, matrix_repr
+
+"""
+Classical codes lifted over a permutation group ring [panteleev2021degenerate](@cite) [panteleev2022asymptotically](@cite).
+
+- `A::Matrix`: the base matrix of the code, whose elements are in a permutation group ring.
+- `repr::Function`: a function that converts a permutation group ring element to a matrix;
```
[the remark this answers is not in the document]
Krastanov: Could you give more specific examples? Permitting an arbitrary function in a situation like this is generally a sign of missing abstraction, especially in a language like Julia where you have access to multimethods.

src/ecc/codes/classical/lifted.jl
Krastanov: This does not make sense to me. The redundancy might be needed for the construction or conceivably for a decoding algorithm even, but the row redundancy would certainly not matter when actually using the parity check matrix -- after all, if you have already measured the non-redundand syndromes, you can just compute the redundant ones classically.

royess: I agree that lifted codes themselves do not need redundant rows or syndromes. I should not say we cannot remove them.

The point I am trying to make is that lifted product construction needs these redundant rows to make the resulting quantum LP codes have a positive dimension (in many cases). Therefore, I think it is beneficial to keep the redundancy in classical lifted codes. Otherwise, they cannot be used in lifted product construction.

So, for the classical lifted codes, I think the better solution is to throw warnings but keep the redundant rows (or at least make it an option).

> after all, if you have already measured the non-redundand syndromes, you can just compute the redundant ones classically.

By this part of your comment, I think you are talking about quantum LP codes. For quantum LP codes, I agree that it makes sense to remove the redundant syndromes.

However, I am unsure whether the removal can be efficiently done. My naive thought is to check one row by another whether it is linearly independent from other rows. When the number of rows and redundant rows is large, this procedure of enumerating rows can be very costly, perhaps more costly than the construction of the parity check matrix.

So, my question is whether it is worthwhile to have this potentially costly removal or whether there are more efficient ways.

Krastanov: These are all good points. Indeed, nothing should be removed from the classical codes. Nothing has to be removed from the quantum codes either, on construction. However, when these quantum codes are used for things like encoding, it makes sense to remove the unnecessary rows inside of the various circuit-creating functions (e.g. the ones that are currently raising errors). For those routines, it does not matter whether the rows are low-weight, so the removal is pretty efficient: just run canonicalize! on a copy of the tableau, and remove the lines beyond the rank.

royess: I see your points. Does it make sense that we should modify the encoding and syndrome routines? Before the routines, we can check the code and do `canonicalize! ` if there is redundancy.

src/ecc/codes/classical/lifted.jl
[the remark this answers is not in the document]
royess: Let's consider a simple case: D4 group. [The wiki.](https://en.wikipedia.org/wiki/Dihedral_group)

In our default way, the representation will be 6x6 matrices, where 6 is the group's order.

But we can find a 4x4 matrix representation for the group. We can start with its representation on integer matrix space:
image

What we need is binary matrices; therefore, we can replace "1" with the 2x2 identity matrix, "0" with the 2x2 zero matrices, and "-1" with the Pauli X matrix. After replacement, we will get 2x2 matrix representation.

src/ecc/codes/classical/lifted.jl
royess: (This seems to be related to your comments on the `repr` in the docstring. There, I left a comment about why I think an other-than-default representation, customized by users, will be useful.)

Regarding the suggestion here, I think it assumes that users would define their own group ring type in analogy to `PermGroupRing`, which can be done in principle but require too much work. I do not expect most users would do that. And if they still use our `PermGroupRing`, then they have to re-define the method and cover the default one, which looks more confusing to me. Also, there is possibility that one wants to use different representations for a same group type in different codes. In that case, the methods will conflict.

Instead, the current implementation assume that users need to customize matrix representation for their own group and still want to use our `PermGroupRing`. The logic behind is that one can always use permutation group to represent any other group, which is the ways `OScar.jl` does, for example. In this way, `PermGroupRing` can repsent the specialized group and we ask a customized function that may better deal the specialized group.

The general function here also looks bad for me. But with these considerations, I do not have better ideas. I would appreciate your input on this.

src/ecc/codes/classical/lifted.jl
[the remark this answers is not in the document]
Krastanov: Examples like this should probably be part of the documentation. Otherwise the user would be pretty confused how to approach these extra capabilities.

src/ecc/codes/lifted_product.jl
```diff
@@ -0,0 +1,58 @@
...
+Children of code families:
+- Hypergraph product code: use trivial group ring `PermutationGroupRing(GF(2), 1)` in lifted codes.
+- Two-block group-algebra (2GBA) code: choose 1×1 base matrices for `c₁` and `c₂`.
+
+See also: [`LiftedCode`](@ref), [`PermGroupRing`](@ref).
+"""
+struct LPCode <: AbstractECC
+    c₁::LiftedCode
+    c₂::LiftedCode
+    repr::Function
```
Krastanov, line 25: I just noticed this, but why does LPCode have a repr function if the LiftedCode structures inside of it have one already as well?

src/ecc/codes/classical/lifted.jl
Krastanov: My reasoning starts from the following point: you are asking the user of this structure to provide an "unlifted" matrix and a lifting function, but the only thing you do with those two entities is to calculate the "lifted" matrix. That is something the user could have already done themselves and then they just provide us with the lifted matrix directly. We do not do anything sophisticated with the abstraction created here, so that abstraction probably should not exist (because the abstraction does not simplify anything for the user -- they still need to know what the unlifted matrix is and what the lifting function is, and now they also need to learn this library-specific abstraction).

Obviously, my description is a bit willfully negative, but this is a design principle on which I believe we should be leaning hard.

Then there is a secondary issue that by just asking for a function without specifying what interface that function should have, we are making this too unconstrained and error-prone.

Lastly, I am confused when you are saying that the user will define their own type in analogy to the one you have created. If they do so, they would not be able to use this structure, because the matrix element type is hardcoded.

Defining a `parity_checks(c::LiftedCode; repr)` would be a bit of a "pun" in Julia parlance. It creates a method for a function that does not follow the structure expected from all other methods, making composability a bit more difficult.

I am still leaning towards avoiding using the new type you have created (I have a lot of questions related to it elsewhere through this PR -- answering them would help me understand why such a custom type is necessary, instead of just using Nemo/Oscar as is). Independently of whether we are relying on the new custom type, setting up a `lift`  function and asking the user to define methods for it seems cleaner to me (with a default method provided by us).

But I might be completely wrong about this. I would not know that until I have more information about how the user is expected to use this code. I have a few more questions in this PR that when answered I will have a better idea. Look at the questions about "how would the user use this to represent common types of codes". I will mark them up again now.

src/ecc/codes/classical/lifted.jl
Krastanov: related to my post from a minute ago: I think this is an important question

test/test_ecc_decoder_all_setups.jl
Krastanov: related to my post from a minute ago: I think this is an important question

royess: I would start with this question. After some consideration, I think the following approaches are user-friendly, which we should provide with clear documentation.

1. The case, where `A`'s elements are permutations (not the general form---the linear combination of group elements). This would require a constructor, `LiftedCode(::Matrix{Perm})`.

2.  The more special case, where `A`'s elements are cyclic permutations. This would require a constructor, `LiftedCode(::Matrix{Int})`, where the integers are the "shift" of the cyclic permutation. The `LP04` and `LP118` belong to this category; the `B04` and `B118` are integer matrices that are supposed to be the input of this new constructor.

3. The general cases, where `A`'s elements are linear combinations of group ring elements. One way is to use dictionaries of `Perm` keys for construction, as I did for `other_lifted_product_codes` in the later part of this file.
 
    Alternatively, we can also first let `x = R(cyclic_permutation(1, l))` as the unit. The meaning of `x` is a 1-shift cyclic permutation. Then, we construct the `A`'s as something like `x^27`, `x^54`, and `1 + x + x^6`. The advantage of this approach is that a lot of papers, such as Panteleev's, express lifted codes exactly in this manner. Therefore, the user can construct the codes directly from these papers without much translation.

The last case, where `1 + x + x^6` is needed, is the most practical reason for the abstraction of the group ring. Without the abstraction, it seems that we cannot have such a clear expression unless we define operators like `+` in dictionaries.

src/ecc/codes/classical/lifted.jl
royess: In https://github.com/QuantumSavory/QuantumClifford.jl/pull/312#discussion_r1684644902, I proposed two new constructors: one from `Matrix{Perm}`, the other from `Matrix{Int}`.

As for the suggestion here,
> maybe a constructor that just takes a dictionary of permutation and does all the conversions behind the scene

I agree it will be good to have this.

> a constructor that just takes the circulant matrices or maybe a separate struct for simpler instances of these codes (bicycle codes and generalized bicycle codes)

I think taking circulant matrices is mathematically the same as taking `Matrix{Int}`, the meaning of which is explained in https://github.com/QuantumSavory/QuantumClifford.jl/pull/312#discussion_r1684644902. And the latter would be more efficient.

As for 2GBA codes, including bicycle codes and generalized bicycle codes, the only difference is the matrix will be 1x1. It looks to me that having something like `LiftedCode(::Perm)` might be slightly confusing, although more convenient.

If we make a separate struct, then seemingly everything other than new constructors will be the same as `LiftedCode`. We can choose to re-implement and simplify the production construction. But it is not very appealing to me.

src/ecc/codes/lifted_product.jl
royess: Regarding the concern you have on generalized bicycle codes, I think we can construct them using 1x1 matrices. For example, we consider the first example in Appendix B (page 20) of https://quantum-journal.org/papers/q-2021-11-22-585/pdf/,

image

using my implementation,

```julia
l = 127
R = PermutationGroupRing(GF(2), l)
x = R(cyclic_permutation(1, l))
A = [1 + x^15 + x^20 + x^28 + x^66]
B = [1 + x^58 + x^59 + x^100 + x^121]
LPCode(LiftedCode(A), LiftedCode(B))
```

The current documentation is indeed so bad for these special cases. I will improve the document by adding these cases.

src/ecc/codes/classical/lifted.jl
royess: Thanks for these comments! I replied to the questions that you marked.

For a short answer to "why such a custom type is necessary" is that we need something with multiplication and addition, where the multiplication is defined by permutation group.

Before writing my type, I checked the types in Nemo and Oscar. But I did not find any of them suitable. There, they have rings that do not have group multiplication and the permutation group that does not have addition. What I have implemented is basically a composition of both.

src/ecc/group_ring.jl
royess: > The CacheDictType and get_cached! are not public and I am very reluctant to depend on them.
> Is there a public interface that is equivalent in Nemo?

Not as far as I know.

> What is the reason the Nemo folks do not have this permutation object not included in their library?
> Is there an alternative that is more idiomatic for them?

They have `Perm` that does not have an addition. 

> Have you spoken with the Nemo folk? Would the accept a contribution like this in their library?

I think this belongs to Nemo logically, but I have not yet.

The main reason I did not talk to Nemo folks in the first place is the absence of the exact division function for group ring elements. Because it is not very useful in our code construction. Also because I do not know how to implement it, although it is mathematically doable. However, this function is required by Nemo's `NCRing` interface, and Nemo folks may want to maintain this consistency.

I will post an issue in Nemo to see their attitude. However, for the above reason, I am not very optimistic about contributing this to Nemo.

> What would your code look like if you do not use this custom type and rather use only what is available in Nemo?

Then, we would use dictionaries of `Perm`-type keys instead, which is the underlying data structure for my group ring element type.

In https://github.com/QuantumSavory/QuantumClifford.jl/pull/312#discussion_r1684644902, I give examples that this abstraction is helpful.

Another reason is that without this abstraction, the implementation of the hypergraph product will be less "smart." With this new type, we can directly reuse `kron` because we already provide methods for operators `*` and `+`. Otherwise, we will reimplement a new version of `kron` based on dictionaries.

I understand sometimes such "smartness" is not a favourable thing. But here, I think this might provide the most clear and consistent way of production construction.

src/ecc/group_ring.jl
royess: > Does Nemo have some form of interface test suite we can use? 

They indeed have a test suite that we need. Adding them will be good.

> Or can we remove functionality we are not using?

When I was writing these, I tried to cover Nemo's interface requirements as much as possible. For users familiar with AbstractAlgebra.jl or Nemo (including myself), this would enable them to manipulate group ring elements as they wish for a ring element. If adding tests can ease the concern about stability, I would still favor such a Nemo-style type.

Krastanov:
Thanks for all the answers you have provided -- they help in wrapping my head around your work. In terms of TODOs, besides the things you are already taking care of, I think we need the following:

- [ ] Make a draft pull request to the Nemo folks with the code you have created for the permutation representations here, to get their opinion on your approach. No expectation to have such a PR merged in the near term, but it would be valuable to engage more properly with that community, given how heavily we are relying on their functionality.
- [ ] Add to the docstring of the permutation object you have created a section along the lines of "why do we need this and why is it not upstreamed to Nemo". I think I am mostly convinced by the reasons you have given, but keeping a record of these reasons for future developers would be very useful. This is related to the first point.
- [ ] Make simple intuitive ways to construct the historical precursors of the general lifted product codes (bicycle and generalized bicycle) with documentation (and references to papers) that explain how the generalization works
- [ ] Make all the constructors you enumerated in https://github.com/QuantumSavory/QuantumClifford.jl/pull/312#discussion_r1684644902 (this is related to the previous point)
- [ ] Document all of these constructors (potentially including examples from papers)
- [ ] Re-examine the choice of internal fields for the code types you have created -- the hardcoded matrix type and the general "representation" function. I personally do not have better ideas right now, but after having the previous 3 points completed, I imagine it would be easier to figure out whether there is a better option here.

Krastanov:
You mentioned that you have some pros and cons in mind about whether to make helper functions, or constructors, or new types for all these "simpler" versions of LP codes -- please proceed with what you believe is most reasonable and do not worry if I happen to use the word "constructor" when you were thinking about a "helper function"

royess:
Thanks for your detailed suggestions!

I think I am clear about these TODOs and plan to complete them in the next couple of days.

src/ecc/codes/classical/lifted.jl
Krastanov: Measuring extra syndromes should not be a problem (if anything, it is required by some fault-tolerant approaches). And the code for generating the measurement circuits should care only about the size of the tableau, not about the rank. I would suggest still raising a warning if the size and rank do not match though.

Krastanov:
closing now that #356 is merged

PR #326, opened 2024-07-30
introducing `doctests` for `tensor` methods and `inv`

Fe-r-oz:
This PR aims to introduce `doctests` for `tensor` methods and `inv` in the `linalg.jl`. This will help users in using these methods which might go unnoticed.  I am looking for your feedback and comments! Thanks for your help reviewing the PR!

### DIFF
--- a/src/linalg.jl
+++ b/src/linalg.jl
@@ -2,6 +2,24 @@
 $TYPEDSIGNATURES
 
 Inverse of a `CliffordOperator`
+
+```jldoctest
+julia> inv(CliffordOperator(sCNOT))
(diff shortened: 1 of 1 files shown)


Krastanov (commented):
remove a few of the non-documenting tests

Krastanov:
Thanks, these are pretty useful! I deleted a couple of them as I am trying to have a separation between "doctests" which exist for the purpose of documentation (and are just incidentally tested, for consistency of the examples when changes are made in the future) and proper tests which exit for testing and do not need to be neat / short / clean as expected of documentation examples.

PR #331, opened 2024-08-03
Make doctest inside @testitem work

thofma:


thofma:
Hm, I guess it is necessary for the Documentation build.

Krastanov:
On the bright side, it seems it is necessary only in `ECC_evaluations.md` :D

Anyway, it seems to be a recognized issue with the TestItems folks https://github.com/julia-vscode/TestItemRunner.jl/issues/49

PR #333, opened 2024-08-03
adding new single-qubit gates

Fe-r-oz:
This PR aims to add the missing single qubit gate w.r.t stim. Added and Tested all the missiing single qubit gates that stim has: https://github.com/quantumlib/Stim/blob/main/doc/gates.md

In this book(https://threeplusone.com/pubs/on_gates.pdf), `SQRTX` is called `V` gate, so `V` gate is used as its name. 
```
julia> sHadamardXY(1)
sHadamardXY on qubit 1
X₁ ⟼ + Y
Z₁ ⟼ - Z

julia> sHadamardYZ(1)
sHadamardYZ on qubit 1
X₁ ⟼ - X
Z₁ ⟼ + Y

julia> sV(1)
sV on qubit 1
X₁ ⟼ + X
Z₁ ⟼ - Y

julia> sInvV(1)
sInvV on qubit 1
X₁ ⟼ + X
Z₁ ⟼ + Y

julia> sSQRTY(1)
sSQRTY on qubit 1
X₁ ⟼ - Z
Z₁ ⟼ + X

julia> sInvSQRTY(1)
sInvSQRTY on qubit 1
X₁ ⟼ + Z
Z₁ ⟼ - X

```


Fe-r-oz:
The gates fail this test (hope this is not too bad), The gate pass all the other tests for single-qubit gates. 

```
s = random_destabilizer(1)
Expression: sop * s == sqsop * s == cop * s == csqop * s == tcop * s == stcop * s
```

Krastanov:
I would expect there to be a but in the gates if they fail this test.

Fe-r-oz:
Indeed, some part of implementation is wrong. My bad!

Fe-r-oz:
Indeed, the implementation was so bad. In particular, The macro qubit kernel descriptions were wrong that's why the failing tests occured in this testset "Convert between small ops". 

The [build ](https://buildkite.com/quantumsavory/quantumclifford/builds/752#01911d82-5f3d-4363-bc93-c545f123c4d4)failed because of more occasional (comparison to MC under perturbative expansion purification example testset) error not related to this PR.  

The [windows latest CI](https://github.com/QuantumSavory/QuantumClifford.jl/actions/runs/10236496786/job/28318612830?pr=333) fails because of this error:

```
Starting tests with 1 threads out of `Sys.CPU_THREADS = 4`...
ERROR: LoadError: InitError: Python: ImportError: DLL load failed while importing _path: The specified procedure could not be found.
```

src/symbolic_cliffords.jl
```diff
@@ -85,12 +85,20 @@ macro qubitop1(name, kernel)
...
+@qubitop1 Y            (x   ,z   , (x⊻z)!=0) 
+@qubitop1 Z            (x   ,z   , x!=0)
+@qubitop1 HadamardXY   (x   ,x⊻z , x==0 && z!=0)
+@qubitop1 HadamardYZ   (x⊻z ,z   , x!=0 && z==0) 
+@qubitop1 V            (x⊻z ,z   , x==0 && z!=0)
+@qubitop1 InvV         (x⊻z ,z   , x!=0 && z!=0)
+@qubitop1 CXYZ         (x⊻z ,x   , z==0 && x==0)
+@qubitop1 CZYX         (z   ,x⊻z , z==0 && x==0)
+@qubitop1 SQRTY        (z   ,x   , z==0)
+@qubitop1 InvSQRTY     (z   ,x   , z!=0 && x==0)
```
Krastanov, line 101: Could you point to a reference for the chosen names? Maybe stim uses the same names?

test/test_symcliff.jl
```diff
@@ -66,7 +66,7 @@
     end
 
     @testset "SingleQubitOperator inv methods" begin
-        for gate_type in [sHadamard, sX, sY, sZ, sId1 , sPhase, sInvPhase]
+        for gate_type in [sHadamard, sX, sY, sZ, sId1, sPhase, sInvPhase, sHadamardXY, sHadamardYZ, sV, sInvV, sSQRTY, sInvSQRTY]
```
Krastanov, line 69: given that this is adding a ton of new types, it will probably be safer to use a `subtypes(AbstractSingleQubit...)` and then filter out the odd one out (`SingleQubitOperator`)

Fe-r-oz: Thank you for this suggestion!

src/symbolic_cliffords.jl
Fe-r-oz: Thank you for particularly focusing on qubit kernel implementations. I hope this is okay.

Fe-r-oz:
BTW, A small comment that is not related to this PR: 

https://github.com/QuantumSavory/QuantumClifford.jl/blob/263195ca43a51b5f67d0496e3c3183558ec112fa/src/QuantumClifford.jl#L78

There is no `delete_columns` function in the `group_tableaux` file so when testing locally, there is error due to this. I checked this file, there is no `delete_columns` function.

src/symbolic_cliffords.jl
Fe-r-oz: No problem!

Krastanov:
> There is no `delete_columns` function in the `group_tableaux`

should be here https://github.com/QuantumSavory/QuantumClifford.jl/blob/263195ca43a51b5f67d0496e3c3183558ec112fa/src/project_trace_reset.jl#L887

Fe-r-oz:
Thanks, It seems I don't have this in the` test_project_trace_reset!`, prolly because I didn't sync the PR.

test/test_symcliff.jl
Fe-r-oz: Thanks once again, this really helped, especially for larger values of `n`, say `100!` There was a minor error that was inverse of `CXYZ` is `CZYX` and vice versa was skipped the earlier test. The` LinearAlgebra.inv ` shows this as well but I had set the wrong inverses for these gates. Now, it's fixed.

Krastanov (changes requested):
This looks pretty good. Thank you for spending time on it! I think the only thing left is:

- [ ] thinking through whether there is some more idiomatic naming convention for us compared to the `DAG` suffix. Probably we will just stick to that suffix.
- [ ] Having a set of tests that specifically check that gates match the definitions that stim uses. E.g. something like `CliffordOperator(gate) == C"Z X"` or whatever the appropriate C string is, taken from the markdown file you shared.

Fe-r-oz:
For us, `Inv` is the naming convention that has been used. Primarily, it can be seen in the `LinearAlgebra.inv` method which we use a lot for `inv(CliffordOperator(...))`. [Searching for](https://github.com/search?q=repo%3AQuantumSavory%2FQuantumClifford.jl+inv&type=code&p=1) `inv`, we see quite a lot of use of this in ~20 files, making it default convention to compared to `DAG`. Also, `sInvPhase` and `sInvZCrY` are existing examples of `inv` convention. Given that, should we revert back to using `Inv`? 

Please share your thoughts on this. Thank you!

Fe-r-oz:
Thanks again for the very hepful `Task TODO list` and feedback! Hopefully, the tasks have been addressed.

Krastanov:
Thanks, this is very useful!

PR #338, opened 2024-08-09
Mutual Information I(A,B) and improving doctests/documentation in  `entanglement.jl`

Fe-r-oz:
This PR aims to introduce Mutual Information `I(A,B)` along with improving the doctests and documentation in the `entanglement.jl` .
 
In the future, if you like, it would be better to have a entanglement introduced in the manual of either "algebra" or upcoming "circuit simulation" manual.   

In the papers, there were many graphs, I think based on entanglement entropy that could be reproduced in a tutorial fashion. Hopefully, this sounds interesting.
....

### DIFF
--- a/docs/src/references.bib
+++ b/docs/src/references.bib
@@ -111,13 +111,13 @@
   publisher={APS}
 }
 
-@article{hein2006entanglement,
+@incollection{hein2006entanglement,
   title={Entanglement in graph states and its applications},
-  author={Hein, Marc and D{\"u}r, Wolfgang and Eisert, Jens and Raussendorf, Robert and Nest, M and Briegel, H-J},
-  journal={arXiv preprint quant-ph/0602096},
-  url={https://arxiv.org/abs/quant-ph/0602096},
-  doi={10.48550/arXiv.quant-ph/0602096},
-  year={2006}
+  author={Hein, Marc and D{\"u}r, Wolfgang and Eisert, Jens and Raussendorf, Robert and Van den Nest, Maarten and Briegel, H-J},
+  booktitle={Quantum computers, algorithms and chaos},
+  pages={115--218},
+  year={2006},
+  publisher={IOS Press}
 }
 
 @article{wilde2009logical,
--- a/src/QuantumClifford.jl
+++ b/src/QuantumClifford.jl
@@ -77,7 +77,7 @@ export
     # Group theory tools
     groupify, minimal_generating_set, pauligroup, normalizer, centralizer, contractor, delete_columns,
     # Clipped Gauge
-    canonicalize_clip!, bigram, entanglement_entropy,
+    canonicalize_clip!, bigram, entanglement_entropy, mutual_information,
     # mctrajectories
     CircuitStatus, continue_stat, true_success_stat, false_success_stat, failure_stat,
     mctrajectory!, mctrajectories, applywstatus!,
--- a/src/entanglement.jl
+++ b/src/entanglement.jl
@@ -132,14 +132,39 @@ end
 """
 $TYPEDSIGNATURES
 
-Get the bigram of a tableau.
+The Bigram `B` of stabilizer endpoints represents the "span" of each stabilizer within a set of Pauli operators `𝒢 = {g₁,…,gₙ}`.
 
-It is the list of endpoints of a tableau in the clipped gauge.
+For each stabilizer `g`, the left endpoint `𝓁(g)` is defined as the minimum site `x` where `g` acts non-trivially, while the
+right endpoint `𝓇(g)` is the maximum site where `g` acts non-trivially. 
+
+The site `x` represent the position within the system, taking values from `{1,2,…,n}` where `n` is the number of qubits.
+
+The bigram set `B(𝒢)` encodes these endpoints as pairs:
+
+`B(𝒢) ≡ {(𝓁(g₁),𝓇(g₁)),…,(𝓁(gₙ),𝓇(gₙ))}`
+
+The clipped gauge `𝒢` is a specific choice of stabilizer state where exactly two stabilizer endpoints exist at each site,
+ensuring `ρ𝓁(x) + ρ𝓇(x) = 2` for all sites `x`.
+
+In the clipped gauge, entanglement entropy is determined only by the stabilizers' endpoints, regardless of their internal structure.
 
 If `clip=true` (the default) the tableau is converted to the clipped gauge in-place before calculating the bigram.
 Otherwise, the clip gauge conversion is skipped (for cases where the input is already known to be in the correct gauge).
 
 Introduced in [nahum2017quantum](@cite), with a more detailed explanation of the algorithm in [li2019measurement](@cite) and [gullans2021quantum](@cite).
+
+```jldoctest
+julia> s = ghz(3)
++ XXX
++ ZZ_
++ _ZZ
+
+julia> bigram(s)
+3×2 Matrix{Int64}:
+ 1  3
+ 1  2
+ 2  3
+```
 
 See also: [`canonicalize_clip!`](@ref)
 """
@@ -162,23 +187,51 @@ end
 """
 $TYPEDSIGNATURES
 
-Get bipartite entanglement entropy of a subsystem
+Get bipartite entanglement entropy of a subsystem `𝒶`. 
 
-Defined as entropy of the reduced density matrix.
+In a system of `n`-qubits divided into regions `𝒶` and `𝒷`, the Rényi (or von Neumann)
+entropy `S(ρ𝒶)` where `ρ𝒶` is reduced density matrix, quantifies the quantum information
+in region `𝒶` after tracing out region `𝒷`.
 
-It can be calculated with multiple different algorithms,
-the most performant one depending on the particular case.
+The entropy is given by: `S(ρ𝒶) = |𝒶| - log₂ |𝒢𝒶|`, where `𝒢𝒶` is a subgroup of stabilizers
+acting non-trivially on `𝒶`. This entropy measures the number of independent stabilizers 
+on `𝒶`,  reflecting the quantum  correlations between regions. 
+
+The stabilizer group `𝒢`, viewed as a binary vector space `V`, can be decomposed such that:
+`S𝒶 = |𝒷| - dim V𝒷 = dim (Π𝒶 V) - |𝒶|` where `Π𝒶` and `Π𝒷` are truncation maps for regions
+`𝒶` and `𝒷`, respectively. 
+
+Initially, each site has one stabilizer and zero entanglement, but as stabilizers enter 
+region `𝒶` over time, entanglement `S𝒶` increases by one bit for each new independent
+operator in `𝒶`.
+
+It can be calculated with multiple different algorithms, the most performant one depending
+on the particular case.
 
 Currently implemented are the `:clip` (clipped gauge), `:graph` (graph state), and `:rref` (Gaussian elimination) algorithms.
 Benchmark your particular case to choose the best one.
+
+See Appendix C of [nahum2017quantum](@cite).
 """
 function entanglement_entropy end
 
 
 """
+$TYPEDSIGNATURES
+
 Get bipartite entanglement entropy of a contiguous subsystem by passing through the clipped gauge.
 
 If `clip=false` is set the canonicalization step is skipped, useful if the input state is already in the clipped gauge.
+
+```jldoctest
+julia> s = ghz(3)
++ XXX
++ ZZ_
++ _ZZ
+
+julia> entanglement_entropy(s, 1:3, Val(:clip))
+0
+```
 
 See also: [`bigram`](@ref), [`canonicalize_clip!`](@ref)
 """
@@ -193,10 +246,25 @@ end
 
 
 """
+$TYPEDSIGNATURES
+
 Get bipartite entanglement entropy by first converting the state to a graph and computing the rank of the adjacency matrix.
 
-Based on "Entanglement in graph states and its applications".
-""" # TODO you should use [hein2006entanglement](@cite) instead of "Entanglement in graph states and its applications", but Documenter is giving the weirdest error if you do so...
+```jldoctest
+julia> using Graphs # hide
+
+julia> s = Stabilizer(Graph(ghz(4)))
++ XZZZ
++ ZX__
++ Z_X_
++ Z__X
+
+julia> entanglement_entropy(s, [1,4], Val(:graph))
+1
+```
+
+Based on [hein2006entanglement](@ref). 
+"""
 function entanglement_entropy(state::AbstractStabilizer, subsystem::AbstractVector, algorithm::Val{:graph})
     graph = Graphs.Graph(state)
     adjmat = Graphs.adjacency_matrix(graph)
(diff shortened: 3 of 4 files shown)


docs/src/references.bib
```diff
@@ -111,13 +111,13 @@ @article{gullans2020dynamical
...
 }
 
-@article{hein2006entanglement,
+@incollection{hein2006entanglement,
   title={Entanglement in graph states and its applications},
-  author={Hein, Marc and D{\"u}r, Wolfgang and Eisert, Jens and Raussendorf, Robert and Nest, M and Briegel, H-J},
-  journal={arXiv preprint quant-ph/0602096},
-  url={https://arxiv.org/abs/quant-ph/0602096},
-  doi={10.48550/arXiv.quant-ph/0602096},
-  year={2006}
```
Krastanov, line 120: could you keep the url and doi (so people can find this resource more easily)

src/entanglement.jl
```diff
@@ -132,15 +132,40 @@ end
...
+right endpoint `𝓇(g)` is the maximum site where `g` acts non-trivially. 
+
+The site `x` represent the position within the system, taking values from `{1,2,…,n}` where `n` is the number of qubits.
+
+The bigram set `B(𝒢)` encodes these endpoints as pairs:
+
+`B(𝒢) ≡ {(𝓁(g₁),𝓇(g₁)),…,(𝓁(gₙ),𝓇(gₙ))}`
+
+The clipped gauge `𝒢` is a specific choice of stabilizer state where exactly two stabilizer endpoints exist at each site,
+ensuring `ρ𝓁(x) + ρ𝓇(x) = 2` for all sites `x`.
```
Krastanov, line 147: I do not think `ρ𝓁(x)` is defined.

src/entanglement.jl
```diff
@@ -162,24 +187,52 @@ end
 """
 $TYPEDSIGNATURES
 
-Get bipartite entanglement entropy of a subsystem
+Get bipartite entanglement entropy of a subsystem `𝒶`. 
 
-Defined as entropy of the reduced density matrix.
+In a system of `n`-qubits divided into regions `𝒶` and `𝒷`, the Rényi (or von Neumann)
```
Krastanov, line 192: We do not use Rényi entropy anywhere here, let's not mention it as it can lead to confusion.

src/entanglement.jl
```diff
@@ -162,24 +187,52 @@ end
...
+entropy `S(ρ𝒶)` where `ρ𝒶` is reduced density matrix, quantifies the quantum information
+in region `𝒶` after tracing out region `𝒷`.
```
Krastanov, line 194: "quantifies the quantum information in region `𝒶` after tracing out region `𝒷`." is not wrong, but it does not really say anything either. You can remove this paragraph and just have a mathematical definition.

src/entanglement.jl
```diff
@@ -162,24 +187,52 @@ end
...
 
-It can be calculated with multiple different algorithms,
-the most performant one depending on the particular case.
+The entropy is given by: `S(ρ𝒶) = |𝒶| - log₂ |𝒢𝒶|`, where `𝒢𝒶` is a subgroup of stabilizers
+acting non-trivially on `𝒶`. This entropy measures the number of independent stabilizers 
+on `𝒶`,  reflecting the quantum  correlations between regions. 
+
+The stabilizer group `𝒢`, viewed as a binary vector space `V`, can be decomposed such that:
+`S𝒶 = |𝒷| - dim V𝒷 = dim (Π𝒶 V) - |𝒶|` where `Π𝒶` and `Π𝒷` are truncation maps for regions
+`𝒶` and `𝒷`, respectively.
```
Krastanov, line 202: This would be more appropriate in a tutorial or a math lecture. The docstring should probably not include all of this, rather just a quick definition. This is a difficult balance to strike, but my main reason to prefer this to not be included is because it is too difficult to follow (much more difficult than just giving the formula and saying it is also related to the rank of the tableau)

src/entanglement.jl
```diff
@@ -162,24 +187,52 @@ end
...
+acting non-trivially on `𝒶`. This entropy measures the number of independent stabilizers 
+on `𝒶`,  reflecting the quantum  correlations between regions. 
+
+The stabilizer group `𝒢`, viewed as a binary vector space `V`, can be decomposed such that:
+`S𝒶 = |𝒷| - dim V𝒷 = dim (Π𝒶 V) - |𝒶|` where `Π𝒶` and `Π𝒷` are truncation maps for regions
+`𝒶` and `𝒷`, respectively. 
+
+Initially, each site has one stabilizer and zero entanglement, but as stabilizers enter 
+region `𝒶` over time, entanglement `S𝒶` increases by one bit for each new independent
+operator in `𝒶`.
```
Krastanov, line 206: similarly getting a bit off topic

src/entanglement.jl
```diff
@@ -227,4 +310,61 @@ function entanglement_entropy(state::AbstractStabilizer, subsystem::AbstractVect
     return nb_of_qubits - rank_after_deletion - nb_of_deletions
 end
 
+"""
+$TYPEDSIGNATURES
```
Krastanov, line 314: Can you post a screenshot of the docs with these additions? I am a bit worried that adding this to every method makes things repetitive (but I am not sure how it renders).

src/entanglement.jl
```diff
@@ -227,4 +310,61 @@ function entanglement_entropy(state::AbstractStabilizer, subsystem::AbstractVect
...
+julia> mutual_information(ghz(3), 1:2, 3:4, Val(:clip))
+2
+```
+
+See Eq. E6 of [li2019measurement](@cite).
+"""
+function mutual_information(state::AbstractStabilizer, A::UnitRange, B::UnitRange, algorithm::Val{:clip}; clip::Bool=true)
+    S𝒶 = entanglement_entropy(state, A, algorithm; clip=clip)
+    S𝒷 = entanglement_entropy(state, B, algorithm; clip=clip) 
+    S𝒶𝒷 = entanglement_entropy(state, _to_unitrange(union(A, B)), algorithm; clip=clip)
```
Krastanov, line 337: Why is `_to_unitrange` necessary?

src/entanglement.jl
```diff
@@ -227,4 +310,61 @@ function entanglement_entropy(state::AbstractStabilizer, subsystem::AbstractVect
...
+Clifford circuits, all Renyi entropies are equal due to the flat entanglement spectrum.
+
+```jldoctest
+julia> mutual_information(ghz(3), 1:2, 3:4, Val(:clip))
+2
+```
+
+See Eq. E6 of [li2019measurement](@cite).
+"""
+function mutual_information(state::AbstractStabilizer, A::UnitRange, B::UnitRange, algorithm::Val{:clip}; clip::Bool=true)
```
Krastanov, line 334: should probably check that A and B do not overlap

src/entanglement.jl
```diff
@@ -227,4 +310,61 @@ function entanglement_entropy(state::AbstractStabilizer, subsystem::AbstractVect
...
++ XZZZ
++ ZX__
++ Z_X_
++ Z__X
+
+julia> mutual_information(s, [1,2], [3, 4], Val(:graph))
+2
+```
+
+"""
```
Krastanov, line 357: Unless the docstrings are specifically to a method, there should probably just be one docstring for the function as a whole.

src/entanglement.jl
```diff
@@ -227,4 +310,61 @@ function entanglement_entropy(state::AbstractStabilizer, subsystem::AbstractVect
...
+function mutual_information(state::AbstractStabilizer, A::AbstractVector, B::AbstractVector, algorithm::Union{Val{:rref}, Val{:graph}}; pure::Bool=false)
+    if algorithm == Val(:graph)
```
Krastanov, line 359: I think it will probably be cleaner to have separate methods and use dispatch, not if branches

Krastanov (changes requested):
thanks, this is an interesting addition. I left a few comments in.

Fe-r-oz:
Thanks for your comments, I will polish it as per your comments.

src/entanglement.jl
Fe-r-oz: `union `of two `UnitRange` returns a `vector`, while `clip` requires `UnitRange.`

```
julia> typeof(union(1:2, 3:5))
Vector{Int64} (alias for Array{Int64, 1})

```

src/entanglement.jl
Fe-r-oz: This is pretty important point. Thanks for letting me know.

src/entanglement.jl
Fe-r-oz: Sorry for that. I will remove all the repetitve `$TYPEDSIGNATURES`. I haven't explored dev docs rendering yet!

src/entanglement.jl
Fe-r-oz: Indeed.  Thanks!

`ρ` represents the reduced density matrix for the subsystem under consideration.

src/entanglement.jl
Fe-r-oz: `_to_unitrange` as helper function is not necessary and it is removed as it was causing one `JET` error when `mutual_information` with `clip` algorithm was called,  which caused two `CI-nightly` tests to fail.

Probably, the function was not a cleaner approach which `JET` pointed out.   Now, it's fixed.

Fe-r-oz:
All of your suggestions have been incorporated. Thanks for your helpful suggestions.

src/QuantumClifford.jl
```diff
@@ -77,7 +77,7 @@ export
     # Group theory tools
     groupify, minimal_generating_set, pauligroup, normalizer, centralizer, contractor, delete_columns,
     # Clipped Gauge
-    canonicalize_clip!, bigram, entanglement_entropy,
```
Krastanov, line 80: let's not export it as this is something we would want to synchronize with QuantumInterface

Could you post an issue at https://github.com/qojulia/QuantumInterface.jl/ describing the need for a new generic function and referencing this PR. The goal would be to pick a name that can be reused by other packages.

In the meantime, keeping this unexported will let us merge this PR without waiting for that coordination.

src/entanglement.jl
```diff
@@ -132,15 +132,41 @@ end
...
+right endpoint `𝓇(g)` is the maximum site where `g` acts non-trivially. 
+
+The site `x` represent the position within the system, taking values from `{1,2,…,n}` where `n` is the number of qubits.
+
+The bigram set `B(𝒢)` encodes these endpoints as pairs:
+
+`B(𝒢) ≡ {(𝓁(g₁),𝓇(g₁)),…,(𝓁(gₙ),𝓇(gₙ))}`
+
+The clipped gauge `𝒢` is a specific choice of stabilizer state where exactly two stabilizer endpoints exist at each site,
+ensuring `ρₗ(x) + ρᵣ(x) = 2` for all sites `x` where `ρ` represents the reduced density matrix for the subsystem under
```
Krastanov, line 147: That sentence is unclear to me. How can two density matrices sum up to be the scalar `2`?

src/entanglement.jl
```diff
@@ -228,3 +283,56 @@ function entanglement_entropy(state::AbstractStabilizer, subsystem::AbstractVect
...
++ ZX__
++ Z_X_
++ Z__X
+
+julia> mutual_information(s, [1,2], [3, 4], Val(:graph))
+2
+```
+
+See Eq. E6 of [li2019measurement](@cite). See also: [`entanglement_entropy`](@ref)
+"""
```
Krastanov, line 309: this should probably be a docstring for a function declaration `function mutual_information end`, and not be attached to a specific method. Or probably even better if there is a method with some default choice of algorithm (that way it will not cause issues when this gets declared in QuantumInterface).

test/test_entanglement.jl
```diff
@@ -50,4 +50,24 @@
         @test entanglement_entropy(copy(s), subsystem, Val(:graph))==2
         @test entanglement_entropy(copy(s), subsystem, Val(:rref))==2
     end
+
+    @testset "Mutual information for Clifford circuits" begin
```
Krastanov, line 54: additional tests that use QuantumOptics to double check the results independently would be very valuable

Krastanov (changes requested):
This looks pretty good.

Apologies for not prioritizing this style of PRs -- you have a lot and I have tried to address the higher value ones first, whenever I have time.

I left a few comments in, mainly motivated by the needs of some other parts of the ecosystem that we try to coordinate with.

src/entanglement.jl
Fe-r-oz: Thank you for helpful comments!

src/entanglement.jl
Fe-r-oz: Thank you for helpful comment!  I will correct this documentation mistake.

src/QuantumClifford.jl
Fe-r-oz: Thank you! Done!

test/test_entanglement.jl
Fe-r-oz: Thank you for this very cool suggestion! I will use stuff from `QuantumOpticsBase` to complete this suggestion!

src/entanglement.jl
Fe-r-oz: Yup, a method with a default choice would be better, so let's use `:clip` as the default algorithm

Fe-r-oz:
Please help review this PR, thank you!  

My apologies for the delay as in adding code review suggestions. I had initially planned to work on the generalized code paper for some time, but I got caught up in them as they seemed quite exciting to read.

Thank you so much for your helpful suggestions! I especially found the suggestion to add `QuantumOpticsBase` tests in  https://github.com/QuantumSavory/QuantumClifford.jl/pull/338#discussion_r1894183567 very helpful and wonderful! I have reduced the code repetition that I did before (my bad) and also added the `QuantumOpticsBase` tests. I hope that the PR is now polished thanks to your helpful codereview suggestions!

There is now warning that `using QOpticsBase.entanglement_entropy` conflicts with`QC.entanglement_entropy ` 😅  in the tests!

Krastanov (changes requested):
- the addition to docs are great
- the new capability is definitely very useful
- the tests for it are great

Couple of notes to double check:
- the suggested commit simplifies things a bit, but now we need to see what type of error it raises if we try to use `:clip`
- making a default does not need kwargs... You can just do `mutual_information(state::..., rangea::..., rangeb::...) = mutual_information(state, rangea, rangeb, Val{:rref}())`
- why did the test fail to catch the mistake with union of ranges with a large gap between them? Is something missing from the tests


Last thing, let's add `mutual_information` to QuantumInterface.jl and import it here (just make a very small PR to QuantumInterface, bump the version number, add changelog, and it will get merged pretty quickly)

Fe-r-oz:
> why did the test fail to catch the mistake with union of ranges with a large gap between them? Is something missing from the tests

Thank you so much for your insights! I’ve fixed an issue where the tests weren’t properly checking for large gaps in the system. The clipped gauge now ensures stabilizers have well-defined left/right endpoints, and the entanglement entropy calculation now correctly errors when non-contiguous regions (e.g., subsystem_rangeA = 1:2, subsystem_rangeB = 5:6) are provided. Additionally, I addressed the out-of-memory errors by using more moderate system sizes, as ket{} becomes computationally expensive for larger systems. The previous implementation was faulty.  My bad 🙏🏼

PR #346, opened 2024-08-22
noncliff: improvements to `apply!` and `dictvaltype`

Fe-r-oz:
Improving `dictvaltype` and `apply!`. 

### DIFF
--- a/src/nonclifford.jl
+++ b/src/nonclifford.jl
@@ -223,11 +223,10 @@ end
 
 nqubits(pc::PauliChannel) = nqubits(pc.paulis[1][1])
 
-"""
-Applies a Pauli channel to the generalized stabilizer `τ = (χ, B(S, D))`, where `χ`
-represents the density matrix and `B(S, D)` denotes the stabilizer basis in which
-`χ` is expressed. The channel is defined as `ρ ↦ ∑ ϕᵢⱼ Pᵢ ρ Pⱼ†`, where `ϕᵢⱼ` are
-complex coefficients, and `Pᵢ` and `Pⱼ` are Pauli operators."""
+"""Applies a Pauli channel to the generalized stabilizer `τ = (χ, B(S, D))`, where
+`χ` represents the density matrix and `B(S, D)` denotes the stabilizer basis in
+which `χ` is expressed. The channel is defined as `ρ ↦ ∑ ϕᵢⱼ Pᵢ ρ Pⱼ†`, where `ϕᵢⱼ`
+are complex coefficients, and `Pᵢ` and `Pⱼ` are Pauli operators."""
 function apply!(state::GeneralizedStabilizer, gate::AbstractPauliChannel; prune_threshold=1e-10)
     dict = state.destabweights
     stab = state.stab
--- a/src/nonclifford.jl
+++ b/src/nonclifford.jl
@@ -243,7 +243,7 @@ function apply!(state::GeneralizedStabilizer, gate::AbstractPauliChannel; prune_
             dᵢ′ = dₗ .⊻ dᵢ
             dⱼ′ = dᵣ .⊻ dⱼ
             χ′ = χ * w * (-tone)^c * (im)^(-phaseₗ+phaseᵣ+4)
-             newdict[(dᵢ′,dⱼ′)] += χ′
+            newdict[(dᵢ′,dⱼ′)] += χ′
         end
     end
     filter!(x -> abs(x[2]) > prune_threshold, newdict)
--- a/src/nonclifford.jl
+++ b/src/nonclifford.jl
@@ -89,6 +89,31 @@ function _stabmixdestab(mixeddestab, d)
     p
 end
 
+"""
+Updates the generalized stabilizer `τ = (χ, B(S, D))` by applying a Clifford gate `G`,
+where `χ` represents the density matrix and `B(S, D)` denotes the stabilizer basis in
+which `χ` is expressed.
+
+```jldoctest
+julia> sm = GeneralizedStabilizer(S"-X")
+A mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is
+𝒟ℯ𝓈𝓉𝒶𝒷
++ Z
+𝒮𝓉𝒶𝒷
+- X
+with ϕᵢⱼ | Pᵢ | Pⱼ:
+ 1.0+0.0im | + _ | + _
+
+julia> apply!(sm, CliffordOperator(sHadamard))
+A mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is
+𝒟ℯ𝓈𝓉𝒶𝒷
++ X
+𝒮𝓉𝒶𝒷
+- Z
+with ϕᵢⱼ | Pᵢ | Pⱼ:
+ 1.0+0.0im | + _ | + _
+```
+"""
 function apply!(state::GeneralizedStabilizer, gate::AbstractCliffordOperator) # TODO conjugate also the destabs
     apply!(state.stab, gate)
     state
--- a/src/nonclifford.jl
+++ b/src/nonclifford.jl
@@ -116,7 +116,7 @@ function _allthreesumtozero(a,b,c) # TODO consider using bitpacking and SIMD xor
 end
 
 function _dictvaltype(dict)
-    return eltype(dict).parameters[2] # TODO there must be a cleaner way to do this
+    return valtype(dict)
 end
 
 function project!(sm::GeneralizedStabilizer, p::PauliOperator)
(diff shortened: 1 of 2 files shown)


Fe-r-oz:
This pr implements a safe version of `apply!`

P.S. There is no `ref[]` due to which there was test error in non-cliff nonclifford quantum optics test branch. commented that out for a time being.

Krastanov:
Ah, this is great, much easier to work with! Thanks for setting it up this way. Now it is not even necessary to worry about squashing commits (mostly), because it is on top of another branch. Request a review whenever you feel it is ready.

Indeed, do not worry about the changelog check at all.

Fe-r-oz:
Please review!

Krastanov:
Hi, Feroz! Please be more careful with review requests -- here for instance you made a review request but then you continued making changes. It is better to wait a couple of days before making a review request to make sure you are done with your changes. Of course, this is not a big deal, especially if it happens just occasionally, however with your pull requests this is pretty common. Your contributions are appreciated, but reviewing is a significant investment of time, so I need to be selective when I start reviewing things.

Fe-r-oz:
Apologies for that. I really your time and your patience. Sorry for these annoying commits towards the very end.

test/test_nonclifford_quantumoptics.jl
```diff
@@ -44,17 +44,17 @@ for _ in 1:10
...
-            ket = qo_bigtgate*ket
-            apply!(genstab, bigtgate)
-            @test dm(ket) ≈ Operator(genstab)
-            if isapprox(expect(qo_pauli, ket), expect(pauli, genstab); atol=1e-5)
-            else
-                @show step
-                ref[] = (stab,pauli)
-                break
-            end
-        end
```
Krastanov, line 58: These tests should not be skipped. Do they run?

`ref[]=...` and `@show` can be deleted

src/nonclifford.jl
```diff
@@ -173,29 +198,30 @@ end
...
     dict = state.destabweights
     stab = state.stab
     dtype = _dictvaltype(dict)
     tzero = zero(dtype)
     tone = one(dtype)
-    newdict = typeof(dict)(tzero) # TODO jeez, this is ugly
+    newdict = typeof(dict)(tzero)
     for ((dᵢ,dⱼ), χ) in dict # the state
-        for ((Pₗ,Pᵣ), w) in zip(gate.paulis,gate.weights) # the channel
+        for ((Pₖ,Pₗ), w) in zip(gate.paulis, gate.weights) # the channel
```
Krastanov, line 212: Where do the names `k` and `l` come from? Are they just arbitrary indices? Previously they meant `left` and `right`. Either one is fine, just want to make sure I understand the naming convention

src/nonclifford.jl
```diff
@@ -173,29 +198,30 @@ end
...
-    for (k,v) in newdict # TODO is it safe to modify a dict while iterating over it?
-        if abs(v) < 1e-14 # TODO parameterize this pruning parameter
-            delete!(newdict, k)
+            cₖₗ = (dot(dₖˢᵗᵃᵇ,dᵢ) + dot(dₗˢᵗᵃᵇ,dⱼ))*2
+            dᵢ′ = dₖ .⊻ dᵢ
+            dⱼ′ = dₗ .⊻ dⱼ
+            χ′ = χ * w * (-tone)^cₖₗ * (im)^(-phaseₖ+phaseₗ+4)
+            if abs(χ′) >= prune_threshold
+                newdict[(dᵢ′,dⱼ′)] = get!(newdict,(dᵢ′,dⱼ′),0)+χ′
+            end
```
Krastanov, line 221: What was actually changed here? It looks like just renaming without changing any of the formulas? I prefer to avoid such renaming for the sake of renaming as it makes looking through the git history more difficult (and it makes reviewing more difficult as I can not figure out whether there was a real change or just renaming)

src/nonclifford.jl
```diff
@@ -116,7 +141,7 @@ function _allthreesumtozero(a,b,c) # TODO consider using bitpacking and SIMD xor
 end
 
 function _dictvaltype(dict)
-    return eltype(dict).parameters[2] # TODO there must be a cleaner way to do this
+    return valtype(dict)
 end
```
Krastanov, line 145: this can altogether be deleted and `valtype` can be used directly in its place

src/nonclifford.jl
```diff
@@ -89,6 +89,31 @@ function _stabmixdestab(mixeddestab, d)
...
+julia> sm = GeneralizedStabilizer(S"-X")
+A mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is
+𝒟ℯ𝓈𝓉𝒶𝒷
++ Z
+𝒮𝓉𝒶𝒷
+- X
+with ϕᵢⱼ | Pᵢ | Pⱼ:
+ 1.0+0.0im | + _ | + _
+
+julia> apply!(sm, CliffordOperator(sHadamard))
```
Krastanov, line 107: just use `tHadamard` instead of converting `sHadamard`

src/nonclifford.jl
```diff
@@ -173,29 +198,30 @@ end
 
 nqubits(pc::PauliChannel) = nqubits(pc.paulis[1][1])
 
-function apply!(state::GeneralizedStabilizer, gate::PauliChannel)
+function apply!(state::GeneralizedStabilizer, gate::AbstractPauliChannel; prune_threshold::Union{Nothing, Float64}=nothing)
+    if prune_threshold === nothing
+        prune_threshold = 1e-14  # Default value
+    end
```
Krastanov, line 204: just set that as the default value, no need to have `nothing` as an option

Krastanov: i.e. `prune_threshold::Float64=nothing`

src/nonclifford.jl
```diff
@@ -173,29 +198,30 @@ end
...
-        end
-    end
-    for (k,v) in newdict # TODO is it safe to modify a dict while iterating over it?
-        if abs(v) < 1e-14 # TODO parameterize this pruning parameter
-            delete!(newdict, k)
+            cₖₗ = (dot(dₖˢᵗᵃᵇ,dᵢ) + dot(dₗˢᵗᵃᵇ,dⱼ))*2
+            dᵢ′ = dₖ .⊻ dᵢ
+            dⱼ′ = dₗ .⊻ dⱼ
+            χ′ = χ * w * (-tone)^cₖₗ * (im)^(-phaseₖ+phaseₗ+4)
+            if abs(χ′) >= prune_threshold
```
Krastanov, line 219: no need to do this twice. Remove this one and keep only the original pruning done outside of the loop

Krastanov (changes requested):
looks like a good start. I left in some initial comments

test/test_nonclifford_quantumoptics.jl
Fe-r-oz: Yup, all of these tests pass now that `expect` was was fixed. `expect `was after this, so at this moment in time I skipped it. But, now they all should pass!

src/nonclifford.jl
Fe-r-oz: Oh...,  I thought I had reverted back to the old convention. I will undo this.

Fe-r-oz:
Thank you, addressed all of your comments. I think it's ready for review.

src/nonclifford.jl
```diff
@@ -188,14 +238,10 @@ function apply!(state::GeneralizedStabilizer, gate::PauliChannel)
             dᵢ′ = dₗ .⊻ dᵢ
             dⱼ′ = dᵣ .⊻ dⱼ
             χ′ = χ * w * (-tone)^c * (im)^(-phaseₗ+phaseᵣ+4)
-            newdict[(dᵢ′,dⱼ′)] += χ′
-        end
-    end
-    for (k,v) in newdict # TODO is it safe to modify a dict while iterating over it?
-        if abs(v) < 1e-14 # TODO parameterize this pruning parameter
-            delete!(newdict, k)
+            newdict[(dᵢ′,dⱼ′)] = get!(newdict,(dᵢ′,dⱼ′),0)+χ′
```
Krastanov, line 241: Why is the `get!` necessary? I thought `DefaultDict` would just provide a `0` by "default".

src/nonclifford.jl
```diff
@@ -173,13 +223,13 @@ end
 
 nqubits(pc::PauliChannel) = nqubits(pc.paulis[1][1])
 
-function apply!(state::GeneralizedStabilizer, gate::PauliChannel)
+function apply!(state::GeneralizedStabilizer, gate::AbstractPauliChannel; prune_threshold::Float64=1e-14)
```
Krastanov, line 226: We should not hardcode `Float64` here. The storage format might be different.

Also, let's me a bit more liberal and change the default to `1e-10`

src/nonclifford.jl
```diff
@@ -302,7 +348,7 @@ end
 
 nqubits(pc::UnitaryPauliChannel) = nqubits(pc.paulis[1])
 
-apply!(state::GeneralizedStabilizer, gate::UnitaryPauliChannel) = apply!(state, gate.paulichannel)
+apply!(state::GeneralizedStabilizer, gate::UnitaryPauliChannel; prune_threshold::Float64=1e-14) = apply!(state, gate.paulichannel; prune_threshold=prune_threshold)
```
Krastanov, line 351:
```suggestion
apply!(state::GeneralizedStabilizer, gate::UnitaryPauliChannel; prune_threshold::Float64=1e-14) = apply!(state, gate.paulichannel; prune_threshold)
```

keyword arguments permit this short notation if the variable names match

Krastanov (changes requested):
Almost there! Some minor cosmetic things left.

Fe-r-oz:
Thank you for your suggestions. I think this is ready for review.

Krastanov:
Thanks!

Issue #309, opened 2024-07-09
Completing the non-Clifford capabilities [$800]


Krastanov:
<details>
<summary><strong>Bug bounty logistic details</strong> (click to expand)</summary>

To claim exclusive time to work on this bounty either post a comment here or message [skrastanov@umass.edu](mailto:skrastanov@umass.edu) with:

- your name
- github username
- **(optional)** a brief list of previous pertinent projects you have engaged in

If you want to, you can work on this project without making a claim, however claims are encouraged to give you and other contributors peace of mind. Whoever has made a claim takes precedence when solutions are considered.

You can always propose your own funded project, if you would like to contribute something of value that is not yet covered by an official bounty.
</details>

The project is claimed by @Fe-r-oz  until Sep 10th 2024.

# Project: Completing the non-Clifford capabilities [$800]

This library already has some nascent capabilities to represent non-Clifford states as a weighted sum of tableaux states. Such a representation is useful in settings where there is a small number of non-Clifford gates in a mostly Clifford circuit. The cost is still exponential in the number of non-Clifford gates, but manageable. We have simple state representation and gate application already done, `expect` is [nearly implemented in this PR](https://github.com/QuantumSavory/QuantumClifford.jl/pull/259), and `project!` still needs to be implemented. To obtain this bounty, the aforementioned features need to be completed, well tested, and well documented.

**Required skills**: Understanding of the Stabilizer formalism and decompositions of arbitrary states in terms of Pauli channels

**Reviewer**: Stefan Krastanov

**Duration**: 2 months

**Publication**: In the next 2 years we plan to release a paper in a selective journal about this software. Contributing to this issue would deserve a co-authorship status on such a paper (if the contributor so desires)

#### Payout procedure:

The Funding for these bounties comes from the National Science Foundation and from the NSF Center for Quantum Networks. The payouts are managed by the NumFOCUS foundation and processed in bulk once every two months. If you live in a country in which NumFOCUS can make payments, you can participate in this bounty program.

[Click here for more details about the bug bounty program.](https://github.com/QuantumSavory/.github/blob/main/BUG_BOUNTIES.md)

Fe-r-oz:
I checked out Ted's paper some time ago, which seems very exciting. Your initial work on non-Clifford is also quite cool about generalized stabilizer. 

Therefore, I will be delighted to work on this as that will teach me more about non-Clifford functionalities. 

[...]

PR #355, opened 2024-09-08
noncliff: define `_proj` subroutine for `projectrand!` and test multi-qubit projection for stabilizer and non-stabilizer states

Fe-r-oz:
PR implements `project!` for `noncliff.` Basic tests are added as well. 


Fe-r-oz:
I think the PR is ready for review. I hope it's not awkward. Thank you!

src/nonclifford.jl
```diff
@@ -168,16 +168,78 @@ function _allthreesumtozero(a,b,c) # TODO consider using bitpacking and SIMD xor
...
 function project!(sm::GeneralizedStabilizer, p::PauliOperator)
     eval = expect(p, sm)
-    prob₁ = (real(eval)+1)/2
-    error("This functionality is not implemented yet")
+    prob = (real(eval)+1)/2
+    if prob ≈ 0
+        return _proj₋(sm, p)
+    elseif prob ≈ 1
+        return _proj₊(sm, p)
+    else
```
Krastanov, line 225: you can just remove these, they are not needed given that you already have the rand() statement below. If anything, they will probably cause some small incorrect bias in your result.

src/nonclifford.jl
```diff
@@ -168,16 +168,78 @@ function _allthreesumtozero(a,b,c) # TODO consider using bitpacking and SIMD xor
...
+- Y
+with ϕᵢⱼ | Pᵢ | Pⱼ:
+ 0.0+0.353553im | + _ | - X
+ 0.0-0.353553im | - X | + _
+ 0.853553+0.0im | + _ | + _
+ 0.146447+0.0im | - X | - X
+```
+
+"""
 function project!(sm::GeneralizedStabilizer, p::PauliOperator)
```
Krastanov, line 218: I think what you have implemented is more properly the `projectrand!` function. Check the docstrings of the two functions. One of them is deterministic and mostly makes sense only for tableaux. The one you have implemented is the one that performs sampling. That is fine though, if anything this might be more approrpiate (just change the name).

test/test_nonclifford_quantumoptics.jl
```diff
@@ -50,3 +50,18 @@ qo_tgate.data[2,2] = exp(im*pi/4)
...
+
+@testset "project!" begin
+    for s in [S"X", S"Y", S"Z", S"-X", S"-Y", S"-Z"]
+        for p in [P"X", P"Y", P"Z", P"-X", P"-Y", P"-Z"]
+            gs = GeneralizedStabilizer(s)
+            apply!(gs, pcT)
+            ρ = Operator(gs)
+            gs_proj, _, _ = project!(gs, p)
+            ρ_proj = Operator(gs_proj)
+            @test isapprox(Operator(gs_proj), ρ_proj; atol=1e-5)
```
Krastanov, line 62: this tests looks a bit strange. You just made `ρ_proj = Operator(gs_proj)` and then checked `ρ_proj == Operator(gs_proj)`. This check will always be (tautologically) true

test/test_nonclifford_quantumoptics.jl
```diff
@@ -50,3 +50,18 @@ qo_tgate.data[2,2] = exp(im*pi/4)
         end
     end
 end
+
+@testset "project!" begin
```
Krastanov, line 54: These tests do not really check the projection operation results. They just run some projection operation and make sure that errors are not raised, but do not do verification.

We also should try something that is not exclusively single-qubit.

Something like this maybe:
```
for n in 1:4
    for repetition in 1:5
        s = random_stabilizer(n)
        p = random_pauli(n)
        gs = GeneralizedStabilizer(s)
        for i in 1:rand(1:3)
            apply!(gs, appropriately_padded_pcT) # this might need an `embed` or `tensor`
        end
        qo_state = Operator(gs)
        projectrand!(gs)
        qo_state_after_proj = Operator(gs)
        # finally, compute the projected result with QuantumOptics and compare
        qo_pauli = Operator(p)
        qo_proj1 = (Identity - qo_pauli)/2
        qo_proj2 = (Identity - qo_pauli)/2
        result1 = qo_proj1*qo_state*qo_proj1'
        result2 = qo_proj2*qo_state*qo_proj2'
        @test qo_state_after_proj ≈ result2 || qo_state_after_proj ≈ result1
```

Krastanov (changes requested):
Thanks for starting this, it is incredibly helpful to have someone tackle it.

I left a few comments in, but I think there is a bigger thing that is currently unclear to me: in `_proj₊` you access only the base tableau out of which all terms in the weighted sum are generated, but you do not access any of those terms themselves, nor are their relative weights taken into account. And it seems that you are returning a pure state, not a `GeneralizedStabilizer`. But one single measurement is not enough to turn a mixed state into a pure state, so there must be an issue with the current code.

src/nonclifford.jl
Fe-r-oz: Thanks a lot for the insights!

test/test_nonclifford_quantumoptics.jl
Fe-r-oz: Indeed!

test/test_nonclifford_quantumoptics.jl
Fe-r-oz: Thanks a lot for this! That is really appreciated.

Fe-r-oz:
Thanks for your pertinent observation. Indeed, something was wrong as I was only considering the tableau, not the catering `X`. What I overlooked was that in the base paper, Ted tells us to update the `GeneralizedStabilizer` state `τ` with the `reduced χ-matrix` namely `χ'` after performing the measurement (`expect`). That I did not... 

This is why Ted says that 'trace Tr [τ′] = Tr [χ′] is the probability of measuring 0' when talking about the case of measuring 0.  My bad...

Now, we have deterministic `project!` and random` projectrand!`.

Fe-r-oz:
The PR is ready for review, Thank you!

src/nonclifford.jl
```diff
@@ -168,17 +168,81 @@ function _allthreesumtozero(a,b,c) # TODO consider using bitpacking and SIMD xor
...
+    return newsm, res
 end
 
+function _proj₋(sm::GeneralizedStabilizer, p::PauliOperator)
+    newstab, res = projectrand!(sm.stab, -p)
+    χ′ = expect(p, sm)
+    n = nqubits(newstab)
+    newsm = GeneralizedStabilizer(newstab, DefaultDict(0.0im, (falses(n),falses(n))=>χ′))
+    return newsm, res
+end
```
Krastanov, line 241: there is a lot of code repetition here. It seems it would be simpler if on line `224` we just change the sign of `p` -- then we do not need separate `proj+` and `proj-` functions.

src/nonclifford.jl
```diff
@@ -168,17 +168,81 @@ function _allthreesumtozero(a,b,c) # TODO consider using bitpacking and SIMD xor
...
+    newstab, res = projectrand!(sm.stab, p)
+    χ′ = expect(p, sm)
+    n = nqubits(newstab)
+    newsm = GeneralizedStabilizer(newstab, DefaultDict(0.0im, (falses(n),falses(n))=>χ′))
+    return newsm, res
 end
 
+function _proj₋(sm::GeneralizedStabilizer, p::PauliOperator)
+    newstab, res = projectrand!(sm.stab, -p)
+    χ′ = expect(p, sm)
```
Krastanov, line 237: It seems you are modifying an internal detail of `sm` before you are calling `expect` on it. Is `sm` still a valid object after this operation?

src/nonclifford.jl
```diff
@@ -168,17 +168,81 @@ function _allthreesumtozero(a,b,c) # TODO consider using bitpacking and SIMD xor
...
+    n = nqubits(newstab)
+    newsm = GeneralizedStabilizer(newstab, DefaultDict(0.0im, (falses(n),falses(n))=>χ′))
```
Krastanov, line 239: For this to be a valid object we need `χ′==1`, right? Is that always the case (e.g. you can use this to check for bugs)? If this is always the case (i.e. no bugs detected), then we can presumably just skip the computation of `χ′`.

test/test_nonclifford_quantumoptics.jl
```diff
@@ -50,3 +50,47 @@ qo_tgate.data[2,2] = exp(im*pi/4)
...
+    for n in 1:4
+        for repetition in 1:5
+            s = random_stabilizer(n)
+            p = random_pauli(n)
+            gs = GeneralizedStabilizer(s)
+            for i in 1:rand(1:3)
+                apply!(copy(gs), embed(n, i, pcT))
+            end
```
Krastanov, line 62: I do not think these 3 lines of code do anything. In line 61 you are making copies of `gs`, so `gs` is never actually modified. This means you are only testing the correctness of your code on pure states, which can easily lead to undetected bugs.

test/test_nonclifford_quantumoptics.jl
```diff
@@ -50,3 +50,47 @@ qo_tgate.data[2,2] = exp(im*pi/4)
...
+            qo_state = Operator(gs)
+            project!(copy(gs), copy(p))[1]
+            qo_state_after_proj = Operator(gs)
```
Krastanov, line 65: this `project!` does not modify your `gs` so `qo_state==qo_state_after_proj` (in other words, `qo_state_after_proj` is definitely not "after proj")

test/test_nonclifford_quantumoptics.jl
```diff
@@ -50,3 +50,47 @@ qo_tgate.data[2,2] = exp(im*pi/4)
...
+    for n in 1:4
+        for repetition in 1:5
+            s = random_stabilizer(n)
+            p = random_pauli(n)
+            gs = GeneralizedStabilizer(s)
+            for i in 1:rand(1:3)
+                apply!(copy(gs), embed(n, i, pcT))
+            end
+            qo_state = Operator(gs)
+            projectrand!(copy(gs), copy(p))[1]
```
Krastanov, line 86: same issues with `copy`

Krastanov (changes requested):
Thanks for continuing with this. It seems there are a few mistakes in the tests. Could you look into them?

Fe-r-oz:
Thank you for your feedback.

I've removed `projectrand!` from this PR, as I don’t have clarity on whether `χ'` is needed in `projectrand!`, based on your comment here: [discussion link](https://github.com/QuantumSavory/QuantumClifford.jl/pull/355#discussion_r1815904266).

This PR focuses solely on implementing `project!`, which corresponds to Algorithm 2 in Ted's paper. Regarding `project!`, Ted states that we should update the `GeneralizedStabilizer` state `τ` with the reduced `χ`-matrix, specifically `χ'`, after performing the measurement (expectation). That is why we take the trace (i.e., $\text{Tr} [\tau'] = \text{Tr} [χ']$) to calculate the probability of obtaining `0` or `1`. I hope this sounds reasonable.

I’ll address `projectrand!` in a separate PR after giving it more thought in light of your feedback.

I attempted to use the suggested test (shown below) from this comment: [discussion link](https://github.com/QuantumSavory/QuantumClifford.jl/pull/355#discussion_r1778945533), but I encountered an error when testing with `Stabilizer/MixedDestabilizer`. Consequently, I added a similar test in the test section that passes for both `GeneralizedStabilizer` and `Stabilizer/MixedDestabilizer`.

While the test may be basic, I would appreciate your feedback on its conceptual correctness. I hope it doesn't fall into the trap of being tautologically true.

Thank you again for your guidance, and I look forward to your thoughts.
```
# gives error for Stabilizer/MixedDestabilizer
julia> @testset "project! for Stabilizer/MixedStabilizer" begin
           for n in 1:4
               for repetition in 1:5
                   s = random_stabilizer(n) # or md = MixedDestabilizer(random_stabilizer(n))
                   p = random_pauli(n)
                   apply!(s, p)
                   qo_state = Operator(s)
                   project!(s, p)[1]
                   qo_state_after_proj = Operator(s)
                   qo_pauli = Operator(p)
                   qo_proj1 = (identityoperator(qo_pauli) - qo_pauli)/2
                   qo_proj2 = (identityoperator(qo_pauli) + qo_pauli)/2
                   result1  = qo_proj1*qo_state*qo_proj1'
                   result2  = qo_proj2*qo_state*qo_proj2'
                   @test qo_state_after_proj ≈ result2 || qo_state_after_proj ≈ result1
               end
           end
       end
```

test/test_nonclifford_quantumoptics.jl
```diff
@@ -50,3 +50,20 @@ qo_tgate.data[2,2] = exp(im*pi/4)
         end
     end
 end
+
+@testset "project!" begin
+    checks = [(S"X",P"X"),(S"Y",P"Y"),(S"Z",P"Z"),(S"-X",P"-X"),(S"-Y",P"-Y"),(S"-Z",P"-Z")]
```
Fe-r-oz, line 55: The tests pass for `gs` when changed to `checks = [(random_stabilizer(1), random_pauli(1))]` as well.

It appears that using a repetition loop outside or employing more than one qubit results in errors, even when utilizing the `s = Stabilizer(...)` instead of `gs`

Krastanov, line 55: This tests only very specific simple states. Right now I do not believe this is actually correct for more complicated states like:

- multi-qubit states
- non-stabilizer states

Using `random_stabilizer` and applying a few non-Clifford gates at random qubits would make this much more believable. It can be in a separate testset to help readability, so no need to necessarily modify what is already here.

test/test_nonclifford_quantumoptics.jl
```diff
@@ -50,3 +50,20 @@ qo_tgate.data[2,2] = exp(im*pi/4)
...
+        apply!(gs, p)
+        qo_state = Operator(gs)
+        project!(gs, p)[1]
+        qo_state_after_proj = Operator(gs)
+        qo_pauli = Operator(p)
+        qo_proj1 = (identityoperator(qo_pauli) - qo_pauli)/2
+        qo_proj2 = (identityoperator(qo_pauli) + qo_pauli)/2
+        result1 = qo_proj1*qo_state*qo_proj1'
+        result2 = qo_proj2*qo_state*qo_proj2'
+        @test qo_state_after_proj ≈ result2 || qo_state_after_proj ≈ result1
```
Krastanov, line 67: it would also be nice to check that one of result1 or result2 is the zero matrix given that you are working stabilizer states only here

test/test_nonclifford_quantumoptics.jl
Krastanov: and multiqubit pauli operators

src/nonclifford.jl
```diff
@@ -172,17 +172,60 @@ function _allthreesumtozero(a,b,c)
...
 
 function _proj₋(sm::GeneralizedStabilizer, p::PauliOperator)
 end
 function _proj₊(sm::GeneralizedStabilizer, p::PauliOperator)
 end
 
+Base.copy(sm::GeneralizedStabilizer) = GeneralizedStabilizer(copy(sm.stab),copy(sm.destabweights))
+
+Base.:(==)(sm₁::GeneralizedStabilizer, sm₂::GeneralizedStabilizer) = sm₁.stab==sm₂.stab && sm₁.destabweights==sm₂.destabweights
+
```
Krastanov, line 228: these would be nice to add (maybe closer to where `struct GeneralizedStabilizer` is defined), so if this PR gets protracted feel free to add them in a separate PR (which probably will get merged faster). Just make sure to add some consistency tests for them.

src/nonclifford.jl
```diff
@@ -172,17 +172,60 @@ function _allthreesumtozero(a,b,c)
...
+𝒟ℯ𝓈𝓉𝒶𝒷
++ Z
+𝒮𝓉𝒶𝒷
+- X
+with ϕᵢⱼ | Pᵢ | Pⱼ:
+ 0.707107+0.0im | + _ | + _
+```
+
+"""
 function project!(sm::GeneralizedStabilizer, p::PauliOperator)
```
Krastanov, line 212: I do not think I understand what this `project!` function is meant to mean.

- It is not a simulation of a (potentially nondeterministic) projective measurement (that would be `projectrand!` and it would involve some randomness in the output)
- It is not what `project!` usually means in this library (a deterministic operation that tells you whether your measurement operator commutes with your stabilizers, and depending on whether it commutes it performs a variety of extra steps to tell you what the state would be after projection). I actually do not believe one can even define `project!` consistenly for `GeneralizedStabilizer`. After all, what would be the meaning of `anticom` and `res` in that case.

In short: I think this needs `projectrand!` and not `project!`. I think earlier in a related PR there was already a discussion of that semantic issue. Maybe we should even define `project!(::GeneralizedStabilizer, ::Pauli) = error(...)` with an explanation why `projectrand!` needs to be used.

Krastanov (changes requested):
Thanks for working on this. There are some issues on semantics in terms of what `project!` and `projectrand!` are supposed to mean and the tests seem to be missing a lot of cases.

test/test_nonclifford_quantumoptics.jl
Fe-r-oz: This has been now implemented with `projectrand!`

src/nonclifford.jl
Fe-r-oz: Thank you! I’ve consolidated `_proj+` and` proj-` into a single `_proj,` and added a detailed explanation in #416 with definition of  `project!(::GeneralizedStabilizer, ::Pauli) = error(...)`.

src/nonclifford.jl
Fe-r-oz: χ′ is the updated density matrix which is not necessarily be `1` as this is observed when we apply a non-clifford op to `sm`, and them find expectation value of a given pauli operator using  `expect ` as done in the doctest. In that particular case,` χ′  = 0.7071067811865475` as seen in doctest whose probability is `~0.85. `

src/nonclifford.jl
Fe-r-oz: Addressed with the latest commit https://github.com/QuantumSavory/QuantumClifford.jl/pull/355/commits/a6c5feda7e6853da4aff28b513ab11d94f336c74

src/nonclifford.jl
Fe-r-oz: Thanks for the suggestion, implemented in #419

test/test_nonclifford_quantumoptics.jl
Fe-r-oz: The condition where result1 or result2 is probabilistic, particularly when the test is repeated multiple times. This holds true even for `Stabilizer` and `MixedDestabilizer` projections, particularly when using multi-qubit `random_stabilizer` states and multi-qubit `random_pauli` operators

Fe-r-oz:
Thank you for your valuable feedback!

This PR builds upon the scaffolding introduced in #420, addresses #418, and includes a test for `GeneralizedStabilizer` from #424. It demonstrates multi-qubit projection using `random_stabilizer` states and the `random_pauli` operator for `GeneralizedStabilizer`, which also holds for `Stabilizer` and `MixedDestabilizer`. The remaining task is to add tests for non-stabilizer case, which pass probabilistically but require additional work. Overall, this PR demonstrates that `GeneralizedStabilizer` encompasses all the capabilities outlined in the Gottesman-Knill theorem.


```mermaid
graph TD

subgraph NittyGritty [ ]
        direction TB
        D--> D1[define_proj]
    end

subgraph Tests [ ]
    direction TB
     E[Consistency Tests] --> E1[random_stabilizer, *done* <br>random_pauli, *done* <br> random_clifford, *done* <br> non-stabilizer **remaining**]
 end

    subgraph ScaffoldingBranch [ ]
        direction TB
        B[Scaffolding] --> B1[completed]
    end
    subgraph ErrorBranch [Unrelated Errors, <br> Issue #418]
        direction TB
        C[Unrelated Errors, <br> Issue #418] --> C1[Resolve projection inconsistencies for multi-qubit Stabilizer/MixedDestabilizer]
    end
    A[projectrand!] --> C
    A[projectrand!] --> B
  A[projectrand!] --> E
    A --> D[Nitty Gritty]
   
```

Fe-r-oz:

test/test_nonclifford_quantumoptics.jl
Fe-r-oz: Thank you!  Incorporated both of your suggestions in the test section.

src/nonclifford.jl
```diff
@@ -233,7 +233,12 @@ function projectrand!(sm::GeneralizedStabilizer, p::PauliOperator)
 end
 
 function _proj(sm::GeneralizedStabilizer, p::PauliOperator)
-    error("This functionality is not implemented yet")
+    # Returns the updated `GeneralizedStabilizer` state sm' = (χ', B(S', D')),
+    # where (S', D') is derived from (S, D) through the traditional stabilizer update.
```
Fe-r-oz, line 237: Informal Review: These comments could benefit from additional detail. While it’s brief, it may lack certain technical details that some readers would find valuable. Consider adding relevant technical information from Ted’s work to enhance clarity. PFA:

```
    # As detailed in https://www.scottaaronson.com/showcase2/report/ted-yoder.pdf, "A generalized
    # stabilizer (χ, B(S, D)) separates the “classical” part of the quantum state from the quantum
    # state." In this framework, the quasi-classical tableau T = (S, D) is updated through Clifford
    # gates and measurements, while the χ-matrix is updated solely by non-Clifford operations.

    # In this divided simulation approach:
    # 1) When encountering a Clifford gate or measurement, we update the tableau as per usual.
    # 2) For a non-Clifford gate or channel, we store the Kraus operator decompositions and the
    #   operator-sum coefficients. These stored operators can be applied with
    #   apply!(state::GeneralizedStabilizer, gate::AbstractPauliChannel) at any time.

    # This function returns an updated `GeneralizedStabilizer` state sm' = (χ', B(S', D')), where
    # (S', D') is derived from (S, D) through standard stabilizer updates. The χ' matrix is
    # updated whenever a non-Clifford gate is applied via apply!(sm, appropriately_padded_pcT).
```

test/test_nonclifford_quantumoptics.jl
Fe-r-oz: Here’s a reproducer showing non-stabilizer projection with the 5-qubit `pcT^3` gate. The goal of this example is to highlight the limitations of `GeneralizedStabilizer` as stated by Ted:

- "No magic state will be simulatable through every stabilizer circuit"
- "What set of quantum states can be simulated efficiently through every stabilizer circuit? What conditions are both necessary and sufficient is unresolved" remains an open question in Ted's work.

In this case, states generated with `embed(5,i,pcT)` for `i = 1, 3, 5` were simulatable, while those with `i = 2, 4` were found to be non-simulatable through the given stabilizer circuit.

Further details on the limitations can be found here: https://github.com/QuantumSavory/QuantumClifford.jl/issues/309#issuecomment-2469207018

```
julia> s = S"+YXZXY
             +Z_YYZ
             -ZZXXZ
             -XX_XZ
             +XZXXX";
julia> p = P"XZYZY";
julia> τ = GeneralizedStabilizer(s);
julia> apply!(τ, embed(5, 1, pcT));
julia> apply!(τ, embed(5, 3, pcT));
julia> apply!(τ, embed(5, 5, pcT));
julia> qo_state = Operator(τ);
julia> projectrand!(τ, p)[1];
julia> qo_state_after_proj = Operator(τ);
julia> qo_pauli = Operator(p);
julia> qo_proj1 = (identityoperator(qo_pauli) - qo_pauli)/2;
julia> qo_proj2 = (identityoperator(qo_pauli) + qo_pauli)/2;
julia> result1 = qo_proj1*qo_state*qo_proj1';
julia> result2 = qo_proj2*qo_state*qo_proj2';
julia> norm_qo_state_after_proj = qo_state_after_proj/tr(qo_state_after_proj);
julia> norm_result1 = result1/tr(result1);
julia> norm_result2 = result2/tr(result2);
julia> @test norm_qo_state_after_proj ≈ norm_result2 || norm_qo_state_after_proj ≈ norm_result1
Test Passed
```

src/nonclifford.jl
```diff
@@ -233,7 +233,23 @@ function projectrand!(sm::GeneralizedStabilizer, p::PauliOperator)
...
+    # 2) For a non-Clifford gate or channel, the Kraus operator decompositions and operator-sum
+    #    coefficients are stored for use with apply!(::GeneralizedStabilizer, ::AbstractPauliChannel).
+
+    # This function returns an updated `GeneralizedStabilizer` state sm' = (χ', B(S', D')), where
+    # (S', D') is derived from (S, D) through traditional stabilizer update. The χ' matrix is
+    # updated whenever a non-Clifford gate is applied via apply!(sm, appropriately_padded_pcT).
+    updated_state, res = projectrand!(sm.stab, p)
+    # sm'.stab' is derived from sm.stab through the traditional stabilizer update.
+    sm.stab = updated_state # in-place
+    return sm, res
```
Krastanov, line 252: There are a lot of comments here, but the one line of code looks definitely wrong:

```
julia> projectrand!(apply!(GeneralizedStabilizer(S"X"), pcT), P"X")
(A mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is
MixedDestablizer 1×1
with ϕᵢⱼ | Pᵢ | Pⱼ:
 0.0+0.353553im | + _ | + Z
 0.0-0.353553im | + Z | + _
 0.853553+0.0im | + _ | + _
 0.146447+0.0im | + Z | + Z, 0x00)

```

Measuring a single-qubit pure state in the X basis should give me an eigenstate of the X operator.

Krastanov (changes requested):
I am not sure whether this is actually finished, but there was a review request. There seems to be some pretty severe issues with this implementation. See the comments.

Could you run me through why this was done in this way? I am a bit confused what the goal or reasoning was.

src/nonclifford.jl
Fe-r-oz: ### Reasoning

The transition from a "pure" state to a "non-stabilizer" state occurs after the application of `apply!(sm, pcT) `(Algorithm 3), as discussed in Ted's framework. This is an essential step because the state being measured is no longer a "pure state." When doing `projectrand!(apply!(GeneralizedStabilizer(S"X"), pcT), P"X")`, the measurement is no longer acting on a pure state because it involves non-Clifford operations, which modify the density matrix and transform the state into a non-stabilizer one.

As Ted explains, the density matrix χ is updated solely by non-Clifford operations. Therefore, the state being measured is not a pure state, but rather a non-stabilizer state that has been evolved through non-Clifford gates.

 At this point, the system is in a pure state:

```
julia> sm = GeneralizedStabilizer(S"X") 
A mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is
𝒟ℯ𝓈𝓉𝒶𝒷
+ Z
𝒮𝓉𝒶𝒷
+ X
with ϕᵢⱼ | Pᵢ | Pⱼ:
 1.0+0.0im | + _ | + _
```

Next, when applying `apply!(sm, pcT)`, the system evolves into a non-stabilizer state:
```
julia> apply!(sm, pcT) 
A mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is
𝒟ℯ𝓈𝓉𝒶𝒷
+ Z
𝒮𝓉𝒶𝒷
+ X
with ϕᵢⱼ | Pᵢ | Pⱼ:
 0.0+0.353553im | + _ | + Z
 0.0-0.353553im | + Z | + _
 0.853553+0.0im | + _ | + _
 0.146447+0.0im | + Z | + Z
```
At this point, the system is in a non-stabilizer state, and the measurement `projectrand!(apply!(GeneralizedStabilizer(S"X"), pcT), P"X")` is performed on this non-stabilizer state, rather than on a pure state.

### Tests for "pure" state under Ted's framework

It’s sad that the tests haven’t been reviewed yet. Here are the tests for "pure" states within this framework:

```
function multiqubit_projrand(τ,p)
    qo_state = Operator(τ)
    projectrand!(τ, p)[1]
    qo_state_after_proj = Operator(τ)
    qo_pauli = Operator(p)
    qo_proj1 = (identityoperator(qo_pauli) - qo_pauli)/2
    qo_proj2 = (identityoperator(qo_pauli) + qo_pauli)/2
    result1 = qo_proj1*qo_state*qo_proj1'
    result2 = qo_proj2*qo_state*qo_proj2'
    return qo_state_after_proj, result1, result2
end

@testset "Multi-qubit projections using GeneralizedStabilizer for stabilizer states" begin
    for n in 1:10
        for repetition in 1:5
            for state in [Stabilizer, MixedDestabilizer, GeneralizedStabilizer]
                s = random_stabilizer(n)
                p = random_pauli(n)
                τ = state(s)
                apply!(τ, random_clifford(n))
                qo_state_after_proj, result1, result2 = multiqubit_projrand(τ,p)
                # Normalize to ensure consistent comparison of the projected state, independent of scaling factors
                norm_qo_state_after_proj = qo_state_after_proj/tr(qo_state_after_proj)
                norm_result1 = result1/tr(result1)
                norm_result2 = result2/tr(result2)
                @test norm_qo_state_after_proj ≈ norm_result2 || norm_qo_state_after_proj ≈ norm_result1
            end
        end
    end
end
```

### Ted's Simulation Framework

The generalized stabilizer (χ, B(T)), separates the "classical" part of the quantum state from the quantum. The quasi-classical tableau T updates through Clifford gates and measurements, while the χ-matrix is updated by non-Clifford operations.

In this divided simulation approach:

- **Clifford gates and measurements**: When we encounter Clifford gates or measurements, we perform the appropriate update to the tableau.
- **Non-Clifford gates and channels**: Likewise, upon encountering a non-Clifford gate or channel, we store the Kraus operator decompositions {(αₕ, bₕ, cₕ)} and the coefficients from the operator-sum, to be used at any time with Algorithm 3.

Thus, simulating the classically difficult part of the circuit, the evolution of the χ-matrix, can be separated from the classically easy part of the circuit, updating the tableau.




Since the density matrix is updated by non-Clifford operations, the evolved state is no longer purely defined. It is a non-stabilizer state that results from the non-Clifford operations applied to the original state. Therefore, under Ted's framework, it would not be accurate to call the evolved state a "pure state," because non-Clifford operations inherently involve transformations that lead to non-stabilizer states, not pure ones.

Reference:
Screenshot_select-area_20241116022734

Krastanov: This is a pretty long post but it does not really address the point I am making. The result of the following is wrong:

```
projectrand!(apply!(GeneralizedStabilizer(S"X"), pcT), P"X")
```

Also, `pcT` is a unitary operation, just not a Clifford one. The state after applying `pcT` is pure.

Krastanov: Not that purity matters in this case. If you measure a single-qubit state, you are getting a pure state after the measurement.

Fe-r-oz: The error appears to be specific to single-qubit states because the unitary channel also updates the density matrix. Typically, the density matrix is updated only by non-Clifford operations. However, in this case, the density matrix is updated by both unitary channels and non-Clifford operations.  The approach outlined in the paragraph above (in the snapshot) does not account for this subtle distinction. 

Thanks! I will incorporate your comments very soon.  I need to specifically address the single-qubit case within this divided simulation approach. 

The good news is the divided simulation approach works for multi-qubit stabilizer and non-stabilizer states, for  example:

```
julia> s = S"+YXZXY
             +Z_YYZ
             -ZZXXZ
             -XX_XZ
             +XZXXX";
julia> p = P"XZYZY";
julia> τ = GeneralizedStabilizer(s);
julia> apply!(τ, embed(5, 1, pcT));
julia> apply!(τ, embed(5, 3, pcT));
julia> apply!(τ, embed(5, 5, pcT));
julia> qo_state = Operator(τ);
julia> projectrand!(τ, p)[1];
julia> qo_state_after_proj = Operator(τ);
julia> qo_pauli = Operator(p);
julia> qo_proj1 = (identityoperator(qo_pauli) - qo_pauli)/2;
julia> qo_proj2 = (identityoperator(qo_pauli) + qo_pauli)/2;
julia> result1 = qo_proj1*qo_state*qo_proj1';
julia> result2 = qo_proj2*qo_state*qo_proj2';
julia> norm_qo_state_after_proj = qo_state_after_proj/tr(qo_state_after_proj);
julia> norm_result1 = result1/tr(result1);
julia> norm_result2 = result2/tr(result2);
julia> @test norm_qo_state_after_proj ≈ norm_result2 || norm_qo_state_after_proj ≈ norm_result1
Test Passed
 
```

Fe-r-oz: Thanks so much for your feedback. I will make sure to add the more tests in the test section so nothing is skipped.

Krastanov: > The error appears to be specific to single-qubit states

Such "special case" reasoning is almost never true. 

This conversation brings up a recurring issue that we should address. This issue was also very common in your initial GSoC application. When there is an error in your code, you frequently search for ways in which you are actually correct and the error is actually minimized. This is an understandable natural habit that every human has, but this habit is extremely detrimental to scientific rigor. If there is a mistake, your first thought should be "oh, this probably affects all kinds of things, I should check as wide as possible for **other** things that might also be broken in the same way". If there is a small mistake, that is a reason to search for bigger mistakes, not a reason to try to paper over the small mistake.

More generally, you should think of yourself as the person that you should trust the least. There is no person that you can confuse more easily than yourself. In science and especially in experimental science it is too easy to rationalize the wrong result to yourself. You have to be your own most aggressive critic.

> Typically, the density matrix is updated only by non-Clifford operations

You are misunderstanding a bit the terminology. A non-Clifford operation can still be unitary and map a pure state to a pure state.

- pure: there is no classical uncertainty; it is purely a quantum state; it can be represented as a ket; it can be represented as a density matrix but that density matrix will be of rank one (just a projector on a single ket);
- mixed state: there is classical uncertainty; it can not be represented as a ket, but it can be represented as a density matrix
- stabiliser state: a very particular pure state; it can be represented as a tableau, as a ket, or a density matrix
- the type of state representation introduced here: a non-stabilizer state represented as a density matrix in a weird stabilizer basis can be mixed or pure
- unitary: an operation that maps a pure state to a pure state
- non-Clifford unitary: an operation that maps a pure state to a pure state but is not a Clifford operation
- a non-unitary channel: an operation that maps a pure state to a mixed state

The non-Clifford formalism discussed here is capable of representing both non-Clifford operations and non-unitary operations. There does not seem to be an error in the paper, only an error in your translation of that paper to code.

Here is an example of it being broken for two-qubit states as well:
```
julia> state = GeneralizedStabilizer(S"XX ZZ");

julia> op = embed(2, 1, pcT);

julia> apply!(state, op);

julia> projectrand!(state, P"XX");

julia> projectrand!(state, P"ZZ");

julia> state
A mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is
𝒟ℯ𝓈𝓉𝒶𝒷
+ Z_
+ _X
𝒮𝓉𝒶𝒷
+ XX
+ ZZ
with ϕᵢⱼ | Pᵢ | Pⱼ:
 0.853553+0.0im | + __ | + __
 0.0+0.353553im | + __ | + Z_
 0.0-0.353553im | + Z_ | + __
 0.146447+0.0im | + Z_ | + Z_
```

This should have been a stabilizer state. Admittedly in the check above and in the first check I gave you I did not carefully verify whether the state can still be represented as a stabilizer state by some coincidence of the summation that is involved in the representation so let's do that do:

```
julia> Operator(state)
Operator(dim=4x4)
  basis: [Spin(1/2) ⊗ Spin(1/2)]
      0.5+0.0im       0.0+0.0im  0.0+0.0im  0.353553-0.353553im
      0.0+0.0im       0.0+0.0im  0.0+0.0im       0.0+0.0im
      0.0+0.0im       0.0+0.0im  0.0+0.0im       0.0+0.0im
 0.353553+0.353553im  0.0+0.0im  0.0+0.0im       0.5+0.0im
```

That seems wrong. And it is not only wrong, but it also shows that `projectrand!` did not even modify the state at all:

```
julia> Operator(apply!(GeneralizedStabilizer(S"XX ZZ"), embed(2,1,pcT)))
Operator(dim=4x4)
  basis: [Spin(1/2) ⊗ Spin(1/2)]
      0.5+0.0im       0.0+0.0im  0.0+0.0im  0.353553-0.353553im
      0.0+0.0im       0.0+0.0im  0.0+0.0im       0.0+0.0im
      0.0+0.0im       0.0+0.0im  0.0+0.0im       0.0+0.0im
 0.353553+0.353553im  0.0+0.0im  0.0+0.0im       0.5+0.0im
```

Krastanov: these examples above are before your last commit 854e5c5

Fe-r-oz: Thank you for all these explanations. This helps a lot. I am going through it.

Yup, the paper says to update the density matrix after measurement which I was not doing before the  https://github.com/QuantumSavory/QuantumClifford.jl/commit/854e5c5faf301ff0034d5dd6b0959e9d95eb224f

> This should have been a stabilizer state.

In the meantime, Let me know your comments with the latest commit: https://github.com/QuantumSavory/QuantumClifford.jl/commit/854e5c5faf301ff0034d5dd6b0959e9d95eb224f .
```
julia> state = GeneralizedStabilizer(S"XX ZZ");

julia> op = embed(2, 1, pcT);

julia> apply!(state, op);

julia> state1 = projectrand!(state, P"XX")[1]
A mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is
𝒟ℯ𝓈𝓉𝒶𝒷
+ Z_
+ _X
𝒮𝓉𝒶𝒷
+ XX
+ ZZ
with ϕᵢⱼ | Pᵢ | Pⱼ:
 0.707107+0.0im | + __ | + __

julia> state2 = projectrand!(state, P"ZZ")[1]
A mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is
𝒟ℯ𝓈𝓉𝒶𝒷
+ Z_
+ _X
𝒮𝓉𝒶𝒷
+ XX
+ ZZ
with ϕᵢⱼ | Pᵢ | Pⱼ:
 1.0+0.0im | + __ | + __

julia> Operator(state1)
Operator(dim=4x4)
  basis: [Spin(1/2) ⊗ Spin(1/2)]
 0.353553+0.0im  0.0+0.0im  0.0+0.0im  0.353553+0.0im
      0.0+0.0im  0.0+0.0im  0.0+0.0im       0.0+0.0im
      0.0+0.0im  0.0+0.0im  0.0+0.0im       0.0+0.0im
 0.353553+0.0im  0.0+0.0im  0.0+0.0im  0.353553+0.0im

julia> Operator(state2)
Operator(dim=4x4)
  basis: [Spin(1/2) ⊗ Spin(1/2)]
 0.5+0.0im  0.0+0.0im  0.0+0.0im  0.5+0.0im
 0.0+0.0im  0.0+0.0im  0.0+0.0im  0.0+0.0im
 0.0+0.0im  0.0+0.0im  0.0+0.0im  0.0+0.0im
 0.5+0.0im  0.0+0.0im  0.0+0.0im  0.5+0.0im
```

Krastanov: One of these does not seem to be normalized. The sum of probabilities should be one. I guess that can be another check that can be added to the tests.

Fe-r-oz:
> There seems to be some pretty severe issues with this implementation

I must acknowledge that I overlooked pages 6 to 8 and equations 14 to 26, which provide the constructive proof of the equations. Although I was aware of this section, I neglected to incorporate it properly. Furthermore, the proper implementation of Algorithm 2, which embodies the constructive proof, was absent from my initial submission. This was a significant oversight on my part, and I apologize for the mistake.

It is also evident that my earlier attempt did not effectively translate the paper into code but rather resulted in a somewhat confused interpretation. For example, the author dedicated two pages to explaining Pauli measurements before presenting Algorithm 2 using the specialized stabilizer basis, emphasizing that a proper implementation requires a more thorough approach than the simplistic one I initially attempted. This misstep is, admittedly, quite embarrassing.

I greatly appreciate your feedback and patience throughout this process. Based on your comments, I have revisited the paper's methodology and reworked the implementation to align with the original approach outlined in Algorithm 2.  As a result, I am closing this PR in favor of the new work, which is currently in progress: #427.

PR #356, opened 2024-09-09
Lifted and lifted product codes via Hecke's GroupAlgebra

royess:
This PR aims to provide lifted and lifted product code construction.

Compared with the previous approach #312, the group algebra (or, group ring) functionality is replaced with the one provided by Hecke, which would benefit the maintenance. I have made the Hecke-dependent codes as an extension, named `QuantumCliffordHeckeExt`.

During implementation, there are the following adjustments concerning other parts of the repository:
1. I have made changes to handle redundancy in parity checks. Previously, it was assumed in many places that `code_n == code_k + code_s`. This holds if the parity-check matrix does not have redundant rows. But with redundancy, as usually the cases for lifted product codes, this does not hold and will cause problems.
2. I have relaxed some code property checks, also related to the redundancy.

Besides, when building circuits from subcircuits, I replace list unpacking with `vcat` because the former will trigger promotion errors when the number of gates gets large.


Fe-r-oz:
Hi, please have a look at the classical Lifted Product codes [here](https://esabo.github.io/CodingTheory/dev/Tutorials/Weight%20Reduction/#Lifted-Products): They are also constructed using Oscar Framework. Maybe adding a test of canonical equivalence of parity check matrices generated by these two constructors can serve as a useful correctness check. There are `Generealized bicycle`, `Hyperbicycle` etc. codes [here](https://esabo.github.io/CodingTheory/dev/Quantum/product_codes/) in `CodingTheory.jl` as well.  Maybe they can also serve as useful correctness tests. I have also referred (above) to a tested implementation of BBQLDPC codes which can also serve as a correctness check as well. Hope you will find these comments useful.

royess:
@Krastanov I made some small changes to make the tests pass. I think it is now ready for review. Thank you!

Project.toml
```diff
@@ -33,6 +35,7 @@ QuantumOpticsBase = "4f57444f-1401-5e15-980d-4471b28d5678"
 
 [extensions]
 QuantumCliffordGPUExt = "CUDA"
+QuantumCliffordHeckeExt = ["AbstractAlgebra", "Hecke"]
```
Krastanov, line 38: This means that the user needs to import both `AbstractAlgebra` and `Hecke` for things to work. Given that `AbstractAlgebra` is already a dependency due to Nemo, I thing it is best to just add it as an explicit dependency (so that users can access the codes just by typing `using Hecke`)

src/ecc/codes/lifted_product.jl
```diff
@@ -0,0 +1,31 @@
+function LPCode(args...; kwargs...)
+    ext = Base.get_extension(QuantumClifford, :QuantumCliffordHeckeExt)
+    if isnothing(ext)
+        throw("The `LPCode` depends on the package `AbstractAlgebra` and `Hecke` but you have not installed or imported them yet. Immediately after you import `AbstractAlgebra` and `Hecke`, the `LPCode` will be available.")
+    end
+    return ext.LPCode(args...; kwargs...)
+end
+
+function two_block_group_algebra_codes(args...; kwargs...)
+    ext = Base.get_extension(QuantumClifford, :QuantumCliffordHeckeExt)
```
Krastanov, line 10: You do not need to make such a check here. Just make sure that your extension adds a new method to `QuantumClifford.ECC.two_block_group_algebra_codes`.

The trick done here is necessary only for structures and their constructors, not for functions.

Krastanov (changes requested):
I have not had time to make an actual review, but here are a couple of quick comments.

royess:
Thanks for your comments! I have made them fixed.

Krastanov:
Congrats on the merge! This is pretty awesome work.

Issue #360, opened 2024-09-15
Color codes [$200]


Krastanov:
## Color codes support for Quantum Clifford's ECC module

After Surface codes, color codes seem to be the most popular topological error correcting codes, and it would be nice to have them available for easy use in Quantum Clifford. Specifically, it would be nice to have built in functions to generate stabilizer tableaux for different color codes, as well as test different syndrome extraction techniques and decoders.

- As there are a few varieties of color codes, a first step to this functionality would be to implement some relevant hardcoded structures, so that a user could easily obtain color code stabilizer tableaux, which can then be combined with pre-existing generic syndrome extraction circuit functionality within QuantumClifford.jl.
-  A second step would be to extend the above functionality to be able to generate color codes of arbitrary sizes based on user-inputed parameters.
- Outside of the scope of the bounty: A third step would then be to add color code specific syndrome extracting circuits as discussed in (for example Fig 2): https://arxiv.org/pdf/2312.08813
- Outside of the scope of the bounty: A fourth step would be to add decoders specific to the color codes, also discussed in the above paper.

**Required skills**: Knowledge of QuantumClifford, color codes, the quantum error correction pipeline, and error correction decoders (for step 4)

**Reviewer**: Stefan Krastanov

**Duration**: a month per step

**Publication**: In the next 2 years we plan to release a paper in a selective journal about this software. Contributing to this issue would deserve a co-authorship status on such a paper (if the contributor so desires)

#### Payout procedure:

The Funding for these bounties comes from the National Science Foundation and from the NSF Center for Quantum Networks. The payouts are managed by the NumFOCUS foundation and processed in bulk once every two months. If you live in a country in which NumFOCUS can make payments, you can participate in this bounty program.

[Click here for more details about the bug bounty program.](https://github.com/QuantumSavory/.github/blob/main/BUG_BOUNTIES.md)

<details>
<summary><strong>Bug bounty logistic details</strong> (click to expand)</summary>

To claim exclusive time to work on this bounty either post a comment here or message [skrastanov@umass.edu](mailto:skrastanov@umass.edu) with:

- your name
- github username
- **(optional)** a brief list of previous pertinent projects you have engaged in

Currently the project is claimed by `no one` until `...`.

If you want to, you can work on this project without making a claim, however claims are encouraged to give you and other contributors peace of mind. Whoever has made a claim takes precedence when solutions are considered.

[...]

PR #361, opened 2024-09-15
Color codes support 1

amicciche:
This PR implements two infinite families of triangular planar color codes out of the three possible on triangular color codes, using https://arxiv.org/abs/1108.5738 as a main reference. I implemented the 4.8.8 and 6.6.6 tilings, for any odd distance the user may provide. A third possible tiling exists, however, to my knowledge, no relevant work has used the 4.6.12 tiling (perhaps weight 12 stabilizer checks seem a bit too difficult on a topological code?). Despite this, I'd be happy to add that functionality as well, if asked. I simply thought it would probably be a better use of time and resources to focus on other functionality before that code. Here is the error correction zoo page for the 4.6.12 code, and notice that according to its references, there hasn't been much mention of it in almost 15 years: https://errorcorrectionzoo.org/c/4612_color

Notably this PR addresses Step 1 and Step 2 of issue Color codes [$200], assuming that these two tilings of triangular color codes are sufficiently many. If not, I'd happily do some more reading and add more!


Krastanov:
related to #360

src/ecc/codes/color_codes.jl
```diff
@@ -0,0 +1,208 @@
+abstract type ColorCode <: AbstractECC end
+
+"""Planar color codes that encode a single logical qubit."""
+abstract type TriangularCode <: ColorCode end
+
+"""Triangular code following the 4.8.8 tiling. Constructor take a distance d as input."""
+struct Triangular4_8_8 <: TriangularCode
```
Krastanov, line 7: darn, these underscores are a bit jarring... How about switching to `Triangular488` - I know it will be weird if we ever implement patterns with double digits, but by then we might just have a more general-purpose constructor.

Same for the other code implemented here.

Krastanov (changes requested):
Thank you for submitting this, it is a very valuable addition!

There are few things to fix up:

- the CI failures (the spell checker and the changelog (you can just post the changelog update in a comment to avoid a merge conflict))
- a few style comments
- more substantially, this needs unit tests
    - confirmations that `throw` happens in case in the future someone modifies the constructor ( in https://github.com/QuantumSavory/QuantumClifford.jl/blob/master/test/test_ecc_throws.jl )
    - testing the properties of these codes by adding them in here: https://github.com/QuantumSavory/QuantumClifford.jl/blob/master/test/test_ecc_base.jl#L66 -- this file gets imported by other test files
    - testing specifically their decoding with perfect matching decoders here https://github.com/QuantumSavory/QuantumClifford.jl/blob/master/test/test_ecc_decoder_all_setups.jl#L77 (if applicable)

src/ecc/ECC.jl
```diff
@@ -366,6 +366,7 @@ include("codes/gottesman.jl")
 include("codes/surface.jl")
 include("codes/concat.jl")
 include("codes/random_circuit.jl")
+include("codes/color_codes.jl")
```
Krastanov, line 369: oh, I forgot. The newly implemented codes should also be exported so that they are part of the public API of the library (see at the start of this file)

src/ecc/codes/color_codes.jl
```diff
@@ -0,0 +1,208 @@
+abstract type ColorCode <: AbstractECC end
+
+"""Planar color codes that encode a single logical qubit."""
+abstract type TriangularCode <: ColorCode end
+
+"""Triangular code following the 4.8.8 tiling. Constructor take a distance d as input."""
```
Krastanov, line 6: Yet one more thing I forgot. This should have a `[](@cite)` reference, that reference should be added to the bibtex file and to the listing of references in the "suggested readings" doc page.

Krastanov:
@amicciche , there are a few folks expressing interest in finishing the bounty. Could you let me know if you would like to finish this PR or to let someone else rebase it and finish it?

amicciche:
> @amicciche , there are a few folks expressing interest in finishing the bounty. Could you let me know if you would like to finish this PR or to let someone else rebase it and finish it?

As all the main functionality is done, I'll try to wrap this up before the start of May. If it can't do that, than yes someone else can rebase and finish it.

Krastanov:
Thanks! Much appreciated!

src/ecc/codes/color_codes.jl
```diff
@@ -0,0 +1,208 @@
...
+end
+
+Triangular4_8_8() = Triangular4_8_8(3) # smallest d
+Triangular6_6_6() = Triangular6_6_6(3) # smallest d
+
+function parity_checks(c::TriangularCode)
+    matrix = get_check_matrix(c)
+
+    num_checks, qubits = size(matrix)
+    return Stabilizer(vcat(matrix,zeros(Bool,num_checks,qubits)), vcat(zeros(Bool,num_checks,qubits), matrix))
```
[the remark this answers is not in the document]
amicciche: The get_check_matrix will return a different matrix based on the subtype of the color code. There are different functions written for the different subtypes. So I don't think the content of that function should be pasted here

I am changing the name to _colorcode_get_check matrix however. I'll also use the CSS constructor

src/ecc/codes/color_codes.jl
```diff
@@ -0,0 +1,208 @@
...
+        return new(d)
+    end
+end
+
+"""Triangular code following the 6.6.6 tiling. Constructor take a distance d as input."""
+struct Triangular6_6_6 <: TriangularCode
+    d::Int
+    function Triangular6_6_6(d)
+        if d%2!=1
+            throw(DomainError("only odd distance trianglular color codes are allowed.\nRefer to https://arxiv.org/abs/1108.5738"))
```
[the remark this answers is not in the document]
amicciche: This is what the error looks like when thrown:
image

amicciche:
For the Changelog update, how about this?

- Added two new abstract types. `ColorCode` which subtypes `AbstractECC`, and `TriangularCode` which subtypes `ColorCode`. 
- Added two new parametrized code families which subtype `TriangularCode`, `Triangular488` and `Triangular666`. These correspond to the 4.8.8 and 6.6.6 tilings found in the paper: https://arxiv.org/abs/1108.5738. Each of these constructors take a single parameter, which corresponds to the desired distance of the code.

PR #364, opened 2024-09-19
reduce JET errors

Fe-r-oz:
This PR aims to reduce JET errors a bit further.

### DIFF
--- a/src/fastmemlayout.jl
+++ b/src/fastmemlayout.jl
@@ -16,8 +16,8 @@ See also: [`fastrow`](@ref)"""
 function fastcolumn end
 
 fastrow(t::Tableau{Tₚᵥ,Tₘ}) where {Tₚᵥ, Tₘ} = t
-fastrow(t::Tableau{Tₚᵥ,Tₘ}) where {Tₚᵥ, Tₘ<:Adjoint} = Tableau(t.phases, t.nqubits, collect(t.xzs))
-fastcolumn(t::Tableau{Tₚᵥ,Tₘ}) where {Tₚᵥ, Tₘ} = Tableau(t.phases, t.nqubits, collect(t.xzs')')
+fastrow(t::Tableau{Tₚᵥ,Tₘ}) where {Tₚᵥ, Tₘ<:Adjoint} = Tableau(t.phases, t.nqubits, Matrix(collect(t.xzs)))
+fastcolumn(t::Tableau{Tₚᵥ,Tₘ}) where {Tₚᵥ, Tₘ} = Tableau(t.phases, t.nqubits, Matrix(collect(t.xzs')'))
 fastcolumn(t::Tableau{Tₚᵥ,Tₘ}) where {Tₚᵥ, Tₘ<:Adjoint} = t
 
 fastrow(s::Stabilizer) = Stabilizer(fastrow(s.tab))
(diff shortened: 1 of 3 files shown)


Fe-r-oz:
There were [two errors](https://github.com/QuantumSavory/QuantumClifford.jl/actions/runs/10859673439/job/30139321989?pr=359#step:6:2085) that were not appearing in the local JET: related to `fast_rows` and `quantum_mallows`. They were appearing in the nightly tests that was conducting the JET tests. I have fixed the former as well. I have tested locally as well. 

src/petrajectory.jl
```diff
@@ -34,7 +34,7 @@ function petrajectory(state, circuit; branch_weight=1.0, current_order=0, max_or
         if status==continue_stat # TODO is the copy below necessary?
             out_probs = petrajectory(copy(newstate), rest_of_circuit,
                 branch_weight=branch_weight*prob, current_order=current_order+order, max_order=max_order)
-            status_probs .+= out_probs
```
Krastanov, line 37: I am not a big fan of this change -- it is great that it fixes a JET issue, but it makes the code much harder to read. Without a better understanding of why this is an issue, I prefer to keep this as it is.

src/fastmemlayout.jl
```diff
@@ -16,8 +16,8 @@ See also: [`fastrow`](@ref)"""
 function fastcolumn end
 
 fastrow(t::Tableau{Tₚᵥ,Tₘ}) where {Tₚᵥ, Tₘ} = t
-fastrow(t::Tableau{Tₚᵥ,Tₘ}) where {Tₚᵥ, Tₘ<:Adjoint} = Tableau(t.phases, t.nqubits, collect(t.xzs))
-fastcolumn(t::Tableau{Tₚᵥ,Tₘ}) where {Tₚᵥ, Tₘ} = Tableau(t.phases, t.nqubits, collect(t.xzs')')
+fastrow(t::Tableau{Tₚᵥ,Tₘ}) where {Tₚᵥ, Tₘ<:Adjoint} = Tableau(t.phases, t.nqubits, Matrix(collect(t.xzs)))
+fastcolumn(t::Tableau{Tₚᵥ,Tₘ}) where {Tₚᵥ, Tₘ} = Tableau(t.phases, t.nqubits, Matrix(collect(t.xzs')'))
```
Krastanov, line 20: I think this introduces a bug:

```
julia> typeof(a)
Matrix{Float64} (alias for Array{Float64, 2})

julia> typeof(a')
LinearAlgebra.Adjoint{Float64, Matrix{Float64}}

julia> typeof(Matrix(a'))
Matrix{Float64} # we want it to be an Adjoint view
```

It undoes the purposeful change to the memory layout. Sure, the results you get by indexing into it are the same, but we specifically want to preserve the column-major or row-major order

Krastanov (commented):
Thank you for spending time on these. Reducing the JET warnings is very valuable, but I am not very comfortable with these changes. One of them introduces a bug and the other one makes the code much less readable. I would say that these are issues that should be fixed in JET itself or a more lower-level fix is necessary (e.g. in Base). Because of this, I will close this PR for now.

It is pretty frustrating and timeconsuming to search for these warnings, I know -- this close to Base it is difficult to make heads or tails of JET reports. I appreciate you attempted these fixes.

Krastanov:
if you are still interested in trying to pursue these tasks (which I should warn you, it will probably be very time consuming work with limited gains), I would suggest looking into what in Base causes the issues.

src/fastmemlayout.jl
Fe-r-oz: Thanks for all these insightful comments.

PR #371, opened 2024-09-27
enable Lifted Product Code tests and reintroduce piracy

Krastanov:
@royess , it seems the tests in your PR were never actually running. Part of them were missing because the dictionary of codes was not set properly (you entered the public function as a key instead of the private type). Part of them were not enabled because they were pasted outside of the testitem definition.

These two issues are now fixed, but that led to actually having errors in the PR, partially due to the fix to the piracy issue that the PR had introduced.

Could you look into this sometime soon? In particular, (1) given how you are using the lift function, maybe we can completely drop the `repr::Function` and just rely on Nemo/Hecke; (2) we need a fix for the adjoint piracy (potentially having this upstreamed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,11 @@
 
 # News
 
-## v0.9.11-dev
+## v0.9.11
 
 - `hcat` of Tableaux objects
 - `QuantumReedMuller` codes added to the ECC module
+- **(breaking)** change the convention for how to provide a representation function in the constructor of `LPCode` -- strictly speaking a breaking change, but this is not an API that is publicly used in practice
 
 ## v0.9.10 - 2024-09-26
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "QuantumClifford"
 uuid = "0525e862-1e90-11e9-3e4d-1b39d7109de1"
 authors = ["Stefan Krastanov <stefan@krastanov.org> and QuantumSavory community members"]
-version = "0.9.11-dev"
+version = "0.9.11"
 
 [deps]
 Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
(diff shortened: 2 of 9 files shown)


royess:
Thanks for letting me know this! I think these tests were missed at some late point when I made mistakes in a merge.

For (1), I want to first note that the `lift` that I defined and `Nemo.lift` are different and unrelated. My `lift` maps a matrix of Hecke group algebra elements to a binary matrix, while `Nemo.lift` works on Nemo elements for type conversion. I think my use of the same name causes this confusion, and we may want to change the name.

Whether to drop the `repr::Function` is a separate issue. It is now defaulted to be `representation_matrix` from `Hecke`. I introduced this freedom because the representation can be more efficient than the default one, as detailed in the docstring of the struct `LiftedCode`.

For (2), I think the problem is that the "adjoint" or "conjugate" of a group algebra element does not seem to be a standard math notation. But here, it would be convenient to define the "conjugate" of a group algebra element by inversing all of its group elements and keeping the coefficients. This convention is also seen in the literature on lifted product codes.

To me, it also looks like a naming problem. To resolve the piracy issue, we may just change the name "adjoint" and make it an internal function. Doing so would also require adjustment of  `hgp`, i.e.,  the hypergraph product.

Krastanov:
Sounds good! I removed the lift piracy (what you did is frequently referred to as a "pun" in the julia ecosystem as you used a function that has a good name to implement a method that does not match how the function is already used).

Check the last commit I added -- it simplified the representation code a bit in order to make it possible to directly provide a Hecke function as `default_repr`. In terms of an API, I prefer to stick to this as anyone seriously working on these codes will be mostly using Hecke anyway and this library will be used only after the code has been converted to a tableau.

Thus, I think the only thing left to do is fix the adjoint piracy. I like the suggestion you made for a fix, but when I tried to implement it I got something wrong (I do not have a very good understanding of Nemo and Hecke). Please submit it when you have a chance (it is not urgent though, no need to rush)

Krastanov:
a side note -- I changed one of your nested vcat/hcat statements to `hvcat(size(x,2), transpose(x)...)` -- could you please double check I did not get this wrong? It is shorter and I think a bit neater, but I am not sure whether you meant the same operation.

royess:
@Krastanov  I have just locally fixed the piracy issue and will soon make a PR for it with some documentation enhancement. (I was previously not aware of the "pun" problem. Thanks for letting me know!)

The `hvcat` is correctly used there.

PR #378, opened 2024-10-04
Lifted Product construction of CoPrime Bivariate Bicycle codes

Fe-r-oz:
This PR aims to introduce an easy construction methods for coprime bivaraite bicycle codes using lifted product `LPCode`, and Oscar `SubPcGroup`.  The code_n and code_k parameters match exactly from the table 2 of https://arxiv.org/pdf/2408.10001v1.


Fe-r-oz:

Fe-r-oz:
Since #394 is fixed, I can now add the tests that reproduce all the results from Table 2 for these codes. I waited because otherwise, it would have caused CI failures.

ext/QuantumCliffordHeckeExt/lifted_product.jl
```diff
@@ -70,6 +70,35 @@ julia> code_n(c2), code_k(c2)
...
+
+julia> A = reshape([𝜋^2 + 𝜋^5 + 𝜋^44], (1,1));
+
+julia> B = reshape([𝜋^8 + 𝜋^14 + 𝜋^47], (1,1));
+
+julia> c = LPCode(A, B);
+
+julia> code_n(c), code_k(c)
+(108, 12)
+```
```
Krastanov, line 100: This seems like an example of a `two_block_group_algebra_codes` so it should probably go in its docstring, not here, next to the other example you have already given (that I merged earlier today, so presumably this also needs a merge with master to avoid a merge conflict).

test/test_ecc_coprime_bivaraite_bicycle.jl
```diff
@@ -0,0 +1,59 @@
+@testitem "ECC coprime Bivaraite Bicycle" begin
```
Krastanov, line 1: very useful set of tests! It should probably directly use `two_block_group_algebra_code` instead of the more round-about construction with LPCode

Krastanov (changes requested):
This is wonderful, thank you! I think it would make sense to put the docstring in a different location though (see comments below)

Fe-r-oz:
Thank you for your suggestions! Incorporated all of your comments and added coprimeBB codes in `test_ecc_base `to test their properties as well!

PR #381, opened 2024-10-05
Multivariate Bicycle code via Hecke's Group Algebra

Fe-r-oz:
This PR implements all the Multivariate Bicycle (MB) using Lifted Product construction and use of Oscar Product of Groups from the paper [Multivariate Bicycle Codes](https://arxiv.org/pdf/2406.19151). All the MB code constructions from Appendix A Table 2 can be verified using the construction method presented. 

Many thanks to Sim J. Xian for clarifying something explicitly, it was an important detail about construction which seemed to be missing from their paper since they only mentioned order `l` and `m`, for two cyclic groups. That third generator is `z=xy `.  I was wondering for some time the order of the third cyclic group for the codes to be MB code: `𝐺ᵣ = ℤ/l₁ × ℤ/l₂ ×  ℤ/l₃` as only `l ` and `m`, order of first two cyclic groups,  were mentioned in their paper explicitly.

 With `z=xy`, the codes can be constructed with correct `code_n` and `code_k` parameters.

I have added doctests for Weight-7 and Weight-6 MB codes, which give correct results for `code_n` and `code_k` values.


Fe-r-oz:

Fe-r-oz:
Since https://github.com/QuantumSavory/QuantumClifford.jl/issues/394 is fixed, added the tests that reproduce all the results from Table 1 for these codes.

Krastanov (changes requested):
This is great as well, but I think the exact same comments as in https://github.com/QuantumSavory/QuantumClifford.jl/pull/381/files apply

PR #382, opened 2024-10-08
Semi Direct Product of cyclic groups via Oscar for 2GBA given structure `ℤₗ⋉ℤₘ` and presentation

Fe-r-oz:
Given a group presentation, and cyclic groups of order `l`, `m`, etc, this PR introduces Semi Direct Product of Groups via `Oscar` for the group structure required by the code construction. This PR aims to define all the 2BGA codes that are currently not supported by the HeckeExt. Currently, the GroupAlgebra only supports one cyclic group, meaning `GroupAlgebra(GF(2), abelian_group(l))` but not group structure for the following cases which are covered in this PR:

- [x] Direct Group Product between two cyclic groups
- [x] Direct Group Product for more than two cyclic groups 
- [x] SemiDirect Product of Group for two cyclic groups
- [x] SemiDirect Product of Group for more than two cyclic groups
- [x] SemiDirect and Direct Products (both) of Group for more than two cyclic groups

These cases can be seen from  Table 1 of https://arxiv.org/pdf/2306.16400.
Screenshot_select-area_20241008061432

I have utilized Direct Product of Groups in previous PRs which can be similarly applied here as well.  In addition, these codes rely on the use of Semi Direct Product of Groups, which is found in `Oscar`. As seen for example aforementioned instances, most of the codes uses structure that is given by Semi Direct Product of Groups with some Direct Product of Groups. The previous PRs introduced  the use of Direct Product of Groups which can accurately use the code instance that uses Direct Product. In this PR, Semi Direct Product of Groups is covered.

I think Oscar will be needed as a weak dependency for HeckeExt for all the cases described above to be used for circuit simulation. The authors has also considered some non-abelian groups, such as Dihedral Group with this [presentation](https://en.wikipedia.org/wiki/Dihedral_group#Other_definitions) for defining direct products of these groups as well. Other groups in consideration are: Alternating Group, Symmetric Group for Direct Product of groups.

They used `GAP` for these calculations and defining group structures in the table, so I will have to check the methods in `GAP` as well.  Specifically,` StructureDescription(SmallGroup(ℓ,#))` from `GAP` was used by them. The  documentation for [StructureDescription ](https://docs.gap-system.org/doc/ref/chap39.html#X87BF1B887C91CA2E)  from `GAP`. 

Huge thanks to Stevell from the Oscar team for valuable feedback about this generalization of direct product of groups.


Krastanov:
this was merged elsewhere

PR #383, opened 2024-10-08
Add `Oscar` as for Free Group Functionality 

Fe-r-oz:
This aims to add `Oscar` as a weak dependency as `Hecke,` `AbstractAlgebra`, or `Nemo` does not support the following when constructing `GroupAlgebra`:

1) Semi Direct Product of two Groups
2) Semi Direct Product of more than Groups
3) Dihedral Groups
4) Alternating Groups
5) Direct Products of the Dihedral Group, Alternating Group, and Symmetric Group

The GroupAlgebra for the 2BGA codes require several of the aforementioned features as mentioned in https://github.com/QuantumSavory/QuantumClifford.jl/pull/382. Even Hecke's `abelian_group` uses group types from `Oscar`, namely `PcGroup`, `SubPcGroup,` for Direct Product of Groups.

Since Oscar also provides GAP integration, this also provide the toolkit to do the following. In #382, the authors use GAP function call to determine the group structure.  This can be done as follows:

```
julia> Oscar.GAP.Globals.StructureDescription(Oscar.GAP.Globals.SmallGroup(36,1))
GAP: "C9 : C4"
# : refers to Semi Direct Product

julia> Oscar.GAP.Globals.StructureDescription(Oscar.GAP.Globals.SmallGroup(42,3))
GAP: "C7 x S3"

julia> Oscar.GAP.Globals.StructureDescription(Oscar.GAP.Globals.SmallGroup(36,2))
GAP: "C36"
```

Furthermore, many paper use the `GAP` `QDistRnd` package for minimum distance calculation for QLDPC codes. Using `Oscar`, we can have access to this package as well. Panteleev used `QDistRnd` and `GNU`'s  http://www.gnu.org/software/glpk/glpk.html for minimum distance calculation. 

```
julia> Oscar.GAP.Packages.install("https://github.com/QEC-pages/QDistRnd.git");
julia> Oscar.GAP.Packages.load("QDistRnd")
true

julia> Oscar.GAP.Globals.DistRandCSS
GAP: function( GX, GZ, num, mindist, opt... ) ... end
```


Fe-r-oz:
This is related to https://github.com/oscar-system/GAP.jl/issues/960
```
Starting tests with 5 threads out of `Sys.CPU_THREADS = 4`...
ERROR: LoadError: InitError: GAP.jl currently does not support multithreaded garbage collection. Please run julia with `--gcthreads=1` for now.
```

Fe-r-oz:
Hi, @lgoettgens, could you help with resolving the `OSCAR` related compatibility issues and weird CI errors?  That would be greatly appreciated.  Pardon for the ping. 

You are closely integrated into OSCAR  ecosystem and have better insights about these CI errors: 

```
Starting tests with 5 threads out of `Sys.CPU_THREADS = 4`...
ERROR: LoadError: InitError: GAP.jl currently does not support multithreaded garbage collection. Please run julia with `--gcthreads=1` for now.
ERROR: LoadError: Windows is not supported, please try the Windows Subsystem for Linux.
For details please check: https://www.oscar-system.org/install/
```

Your insights about this would be greatly appreciated: https://github.com/QuantumSavory/QuantumClifford.jl/pull/383#issue-2572768072. Thank you for your time.

lgoettgens:
Hi @Fe-r-oz,

some comments to your above questions:
- `ERROR: LoadError: Windows is not supported, please try the Windows Subsystem for Linux.`: Oscar, in particular GAP.jl does not work on Windows. You need to check in this QuantumClifford's CI setup that Oscar is not used in any Windows job.
- `ERROR: LoadError: InitError: GAP.jl currently does not support multithreaded garbage collection. Please run julia with `--gcthreads=1` for now.`: As the message already tells you, GAP.jl v0.10.x does not support multithreaded garbage collection. The message contains a setting to circumvent this on the julia side. GAP.jl v0.11.x (and v0.12.x) fixed this restriction, but you will need to wait for Oscar v1.2.0 release to get access to it.
- If you don't want to use Oscar and GAP, but first only Hecke: there is `Hecke.small_group(i, j)` which gives you the `j`-th group of order `i`. If you have some few fixed group constructions, you could just lookup which ID in this database correspond to the groups you are interested in. Of course, this is a hard restriction compared to using Oscar/GAP groups.
- One part of Oscar philosophy is no need to use `GAP.Globals.xxx` calls. Instead, Oscar provides `snake_case` functions for a lot of things already. E.g. `Oscar.GAP.Globals.StructureDescription(Oscar.GAP.Globals.SmallGroup(36,1))
` can be written as `describe(small_group(36,1))` with the additional benefit that this returns a julia string (instead of a GAP string).
- Similar to the previous point, only the Oscar objects are properly typed (`GAP.Globals.SymmetricGroup(3)` returns a `GapObj` while `symmetric_group(3)` returns a `PermGroup`), and `group_algebra` only takes the properly typed ones.

I wanted to write another bullet point but forgot what... ah damn...

Fe-r-oz:
Indeed. The GAP support was the side thing that was not really required.

I think the last point contained the answer to 'What about Direct Products of Groups, Semi Direct Products of Groups, as they are essential to the construction of Group Algebra shown in #382 and in many other places for our requirements? For example,  Using `abelian_group(PcGroup, [l,m])` is quite helpful for direct products.

Fe-r-oz:
Closing in favor of #400

Issue #307, opened 2024-07-09
Improve static analysis tests with JET [$200]


Krastanov:
<details>
<summary><strong>Bug bounty logistic details</strong> (click to expand)</summary>

To claim exclusive time to work on this bounty either post a comment here or message [skrastanov@umass.edu](mailto:skrastanov@umass.edu) with:

- your name
- github username
- **(optional)** a brief list of previous pertinent projects you have engaged in

Currently the project is claimed by `no one` until `...`.

If you want to, you can work on this project without making a claim, however claims are encouraged to give you and other contributors peace of mind. Whoever has made a claim takes precedence when solutions are considered.

You can always propose your own funded project, if you would like to contribute something of value that is not yet covered by an official bounty.
</details>

# Project: Improve static analysis tests with JET [$200]

We use JET for static analysis of the codebase. **To claim this bounty, fix all of the reported warnings**. Some of them are false positives that can be removed with extra type annotation. Some of them are due to dependencies, in which case you can simply blacklist the dependency (as already done). See the `test_jet.jl` test file and the JET CI run for the current state of affairs.

**Required skills**: General Julia and JET skills

**Reviewer**: Stefan Krastanov

**Duration**: 1 month

#### Payout procedure:

The Funding for these bounties comes from the National Science Foundation and from the NSF Center for Quantum Networks. The payouts are managed by the NumFOCUS foundation and processed in bulk once every two months. If you live in a country in which NumFOCUS can make payments, you can participate in this bounty program.

[Click here for more details about the bug bounty program.](https://github.com/QuantumSavory/.github/blob/main/BUG_BOUNTIES.md)

JoanArrow:
I would like to claim this one!

Krastanov:
Hi, @JoanArrow ! Sounds good, happy to reserve it. Let me know if you would like to proceed (we can talk more about it at the hackathon tonight).

lgoettgens:
Hi @Krastanov, after already succeeding https://github.com/QuantumSavory/QuantumSavory.jl/issues/130 this issue seems like the logical next step to me. Is this bounty currently open or already reserved?

Krastanov:
[...]

PR #385, opened 2024-10-09
Fix issues reported by JET

lgoettgens:
Resolves https://github.com/QuantumSavory/QuantumClifford.jl/issues/307.


src/mul_leftright.jl
```diff
@@ -56,11 +56,14 @@ function mul_ordered_lv!(r::AbstractVector{T}, l::AbstractVector{T}; phases::Val
...
 =#
 
+# This method is already covered by the identical but less specific one right below. (And JET does not like it)
+#=
 function mul_ordered!(r::SubArray{T,1,P,I1,L1}, l::SubArray{T,1,P,I2,L2}; phases::Val{B}=Val(true)) where {T<:Unsigned, B, I1, I2, L1, L2, P<:Adjoint}
     # This method exists because SIMD.jl does not play well with Adjoint
     # Delete it and try `QuantumClifford.mul_left!(fastcolumn(random_stabilizer(194)), 2, 1)` # works fine for 192
     _mul_ordered_nonvec!(r,l; phases=B)
 end
+=#
```
lgoettgens, line 66: @Krastanov should I keep this method commented out or straight up remove it?

lgoettgens:
This is good to go from my POV @Krastanov

Krastanov:
This looks great, thank you! It will probably take me a couple of days to get around to completely reviewing it. I am curious about one particular point:

I see that in a lot of places you have changed old historic code like `x.rank` to much neater and idiomatic `rank(x)`. That is great, but I am curious why did that have an effect on the amount of JET warnings. Do you have a brief explanation?

lgoettgens:
> I see that in a lot of places you have changed old historic code like `x.rank` to much neater and idiomatic `rank(x)`. That is great, but I am curious why did that have an effect on the amount of JET warnings. Do you have a brief explanation?

In an in-between version while fiddling around with some reports, I thought that one JET report was due to badly inferred `.tab` (or `.rank`, don't remember it exactly). My "solution" was to add a type assertion to a `tab(::Dunno)` return value, and update the call-sites to use this getter instead of direct field access. It turned out that the JET issue was resolved by some other change on its own, but I thought I'll just keep this improvement in the PR.

src/mul_leftright.jl
Krastanov: yup, indeed. I just removed it while reviewing

Krastanov:
Thank you, this is very much appreciated! This should be merged soon, just waiting on a final round of tests. I submitted the final list of bounties to NumFOCUS this week, and I will try to add this one as well, but there is a chance they will have to process it in the next round.

Krastanov:
I merged master in here because there was a bug in LinearAlgebra 1.11.1 that we were hitting. The master branch has the workaround.

lgoettgens:
> Thank you, this is very much appreciated! This should be merged soon, just waiting on a final round of tests. I submitted the final list of bounties to NumFOCUS this week, and I will try to add this one as well, but there is a chance they will have to process it in the next round.

Thanks!

PR #352, opened 2024-09-02
Generalized Circulant Bivariate Bicycle code 

Fe-r-oz:
Dear @Krastanov, 

In this 2024 paper [High-threshold and low-overhead fault-tolerant quantum memory](https://arxiv.org/pdf/2308.07915), Bravyi et. al. has introduced a simple `BBQLDPC` code.  In the `Table 3`, as seen below, these are the code constructions given in the paper.

All these constructions have been tested (especially at the moment, tests have been added to verify the number of logical qubits, `k` which is calculated on formula: ` k = n - rank(Hx) - rank(Hz) `.  Please checkout the test section:
Screenshot_select-area_20240902134341

 The interface is user-friendly:
```
julia> using QuantumClifford.ECC: BBQLDPC, parity_checks

julia> parity_checks(BBQLDPC(12, 6, [3, 1, 2], [3, 1, 2]))
# the stabilizer
```

The documentation is simple so it's easier to follow as well.  Python implementation of this code is available at [here](https://github.com/sbravyi/BivariateBicycleCodes/tree/main). 

Cross verification between python implementation and this implementation has been done. The matrices are huge, but we can export results from python, import them in julia and compare the results.  

Would you like to have a dataset prepared for these given constructions, so we can import them and test for correctness comparison in the test section?

```
julia> using MAT
julia> using QuantumClifford
julia> using QuantumClifford.ECC: BBQLDPC, parity_checks, naive_encoding_circuit
julia> mat_file_path = "/home/Feroz/Desktop/matrices96.mat";
julia> mat_data = matopen(mat_file_path);
julia> AT = read(mat_data, "AT");
julia> BT = read(mat_data, "BT");
julia> BTT, ATT = parity_checks(BBQLDPC(9, 6, [3, 1, 2], [3, 1, 2]))
([0 0 … 0 0; 0 0 … 0 0; … ; 0 0 … 0 0; 0 0 … 0 0], [0 0 … 0 0; 1 0 … 0 0; … ; 0 0 … 0 0; 0 0 … 1 0])
julia> BTT == BT
true
julia> ATT == AT
true
julia> mat_file_path = "/home/Feroz/Desktop/matrices2814.mat";
julia> mat_data = matopen(mat_file_path);
julia> AT = read(mat_data, "AT");
julia> BT = read(mat_data, "BT");
julia> BTT, ATT = parity_checks(BBQLDPC(28, 14, [26, 6, 8], [7, 9, 20]))
[...]

PR #387, opened 2024-10-10
2D implementation of Bivariate Bicycle qLDPC codes: Toric Layouts

Fe-r-oz:
This PR implements the 2D implementation of Bivariate Bicycle Codes via Toric Layout and Embeddings in ℤ² from the paper: https://arxiv.org/pdf/2404.17676. All these Toric layouts has been tested and cross checked with the original implementation.  Some details are presented here: https://github.com/QuantumSavory/QuantumClifford.jl/pull/352#issuecomment-2395768519. The original python implementation is quite slow and causes many allocations. I have used sparse arrays and optimized using sparse matrices which causes almost 0 allocations and speed-up in run-time as well.

I think these low-hanging fruits from the paper will be useful in circuit simulation of these codes, but I have not familiar with those part of the paper yet.

`Toric Layouts:`
The paper has not tried plotting these Toric Layouts. They only plotted the Embedding in  ℤ² (the second graph). But I think, we can also plot the Toric Layouts as following:
Screenshot_select-area_20241010073452

`Embeddings in ℤ²`
Towards the end, we will be able to Plot the Embeddings of these codes in 2D. Probably, this will be divided into smaller PR so each PR is well documented and tested separately. 

my_plot11


Fe-r-oz:
The PR is ready for review, @Krastanov. Thank you.

Krastanov:
Issue logged here https://github.com/QuantumSavory/QuantumClifford.jl/issues/652

PR #389, opened 2024-10-14
examples of generalized bicycle codes

Fe-r-oz:
This PR aims to add tests for 2BGA codes from Reference.  Table 1: https://arxiv.org/pdf/2306.16400. I have verified all the appropriate code instance using this new method as well. I have created a test file for this as well. 

@royess, just pinging you for reference. 

Also, there was a typo, 2GBA was used, instead of 2BGA, fixed this as well.


Fe-r-oz:
This is good to go from my POV @Krastanov

P.S. The failing tests are unrelated to this PR.

royess:
I think there is already a `generalized_bicycle_codes` that does exactly the same thing. So I do not get the point why we need this.

Fe-r-oz:
Oh... My bad. I knew this fact, but somehow overlooked it because working on two separate things simultaneously.

Rebased this PR  fixing the typo and verifying the test cases from aforementioned paper. Thanks.

test/test_ecc_2bga.jl
```diff
@@ -0,0 +1,16 @@
...
+    @test code_k(generalized_bicycle_codes([0 , 15, 16, 18], [0 ,  1, 24, 27], 35)) == 8
+    @test code_n(generalized_bicycle_codes([0 ,  1,  3,  7], [0 ,  1, 12, 19], 27)) == 54
+    @test code_k(generalized_bicycle_codes([0 ,  1,  3,  7], [0 ,  1, 12, 19], 27)) == 6
+    @test code_n(generalized_bicycle_codes([0 , 10,  6, 13], [0 , 25, 16, 12], 30)) == 60
+    @test code_k(generalized_bicycle_codes([0 , 10,  6, 13], [0 , 25, 16, 12], 30)) == 6
+    @test code_n(generalized_bicycle_codes([0 ,  9, 28, 31], [0 ,  1, 21, 34], 36)) == 72
+    @test code_k(generalized_bicycle_codes([0 ,  9, 28, 31], [0 ,  1, 21, 34], 36)) == 8
+    @test code_n(generalized_bicycle_codes([0 ,  9, 28, 13], [0 ,  1, 21, 34], 36)) == 72
+    @test code_k(generalized_bicycle_codes([0 ,  9, 28, 13], [0 ,  1,  3, 22], 36)) == 10
+end
```
Krastanov, line 16: I am a bit confused: these are "generalized_bicycle_codes". Shouldn't the test file be called "test_ecc_generalized_bicycle_codes.jl"?

ext/QuantumCliffordHeckeExt/lifted_product.jl
```diff
@@ -139,19 +139,31 @@ code_n(c::LPCode) = size(c.repr(zero(c.GA)), 2) * (size(c.A, 2) * size(c.B, 1) +
...
+
+```jldoctest
+julia> l = 35;
+
+julia> c1 = generalized_bicycle_codes([0, 15, 16, 18], [0, 1, 24, 27], l);
+
+julia> code_n(c1), code_k(c1)
+(70, 8)
+```
+
```
Krastanov, line 156: This seems like a docstring that should go in the function `generalized_bicycle_codes`, not in `two_block_group_algebra_codes`

Krastanov (changes requested):
This is a great addition, thank you! I am a bit confused about where it is placed. Would it not make more sense in `generalized_bicycle_codes`?

test/test_ecc_2bga.jl
Fe-r-oz: You are right.

ext/QuantumCliffordHeckeExt/lifted_product.jl
Fe-r-oz: You are right. Let's put in docstring of BG codes.

In addition, GB codes are a subset of only abelian 2BGA codes. 2BGA codes are formed by a general, finite group. Also, this test is rather relaxed in a sense that this is not a test for 2BGA codes, only abelian version of 2BGA and also because the original paper used small groups to create 2BGA codes which is done in https://github.com/QuantumSavory/QuantumClifford.jl/pull/406, and this is more like a subset test for GB codes.

PR #390, opened 2024-10-14
tests: Specific Group Presentations Using Oscar's Free Group for Non-Abelian and Abelian Groups

Fe-r-oz:
This PR aims to add new method for the two-block group codes using non-abelian groups. Reference Table 1: https://arxiv.org/pdf/2306.16400. 


Fe-r-oz:
@Krastanov, @royess, There are many groups mentioned in the aforementioned paper which are used to create group algebra using direct products/semi products of following groups which are not currently supported which is the aim of this PR:

1. `D x C`
2. `A x C`
3. `S x C`
4. `C ⋉ C `

where `D` is Dihedral Group, `A` is Alternating Group, `S` is Symmetric Group. Semidirect product is also being utilized by Lin, as shown in Table 1

There are two methods for using these non-abelian groups based on discussion with Oscar Team. Using` Hecke`, this can be done with a `small_group` with a multiplication table which is the aim of this PR.  Lin actually uses the GAP analogue of this which is `small_group(l, #) `in `GAP`.  This is supported in `Hecke` as well. More details about this is presented here: https://docs.gap-system.org/doc/ref/chap39.html#X87BF1B887C91CA2E

This approach does not need `Oscar` as all can be done with `Hecke` but users would have to search the multiplication table to see whether they are getting the right group that is required. Meaning the entries for small group have to be carefully entered other wise the wrong group will the received. 

```
julia> using Oscar;

julia> using Hecke;

# searching the multiplication table
julia> describe(small_group(12, 1))
"C3 : C4", # this means C3 semidirect product with C4

julia> describe(small_group(12, 2))
"C12"

julia> describe(small_group(12, 3))
"A4" # Alternating group

julia> describe(small_group(12, 4)) # this is dihedral group of order 12, used in doctest
"D12"

julia> describe(small_group(12, 5))
"C6 x C2" # C6 direct product with C2
```

This PR aims to handle all the code instances from Appendix Table 2 of aforementioned paper where authors take direct product of Dihedral and cyclic groups (`D_m x C_2`). For example, the doctest, the `small_group(12, 4) `provides` D12` which is mentioned in the paper.  

The users enter the `small_group(entry1,entry2)` for each of the small_groups and then group algebra is constructed under the hood, as following:

`two_block_group_algebra_codes([0, 10], [0, 8, 9, 4, 2, 5], (12, 4), (2, 1));` This represents `D12 x C2`

The other approach is to use semi direct products which is https://github.com/QuantumSavory/QuantumClifford.jl/pull/382 which provides a more hands on approach where we can construct the group structure itself  mainly when semidirect products are required. This is useful when custom group presentations with relators (`R`) are given to us. If a group presentation is custom defined for example,  `C_24 x C_2 `where presentation is `⟨r, s|s^2, r^24, (rs)^8⟩`, as by Lin in Table 1, then the mapping via homomorphism can be defined as well taking the presentation into account.

I will turn this to draft to incorporate any more feedback and comments.

Fe-r-oz:
The PR is nearly implemented. But turning this to WIP as I will add and test the entire non-abelian Table 3 and Table 1 results as well with Hecke.small_group by adding them as test cases.

Fe-r-oz:
Since #391 has been resolved, The Table 1 results has been reproduced!  I have verified the Table 1 results and will push it  the test file once the bug in Issue 394 is resolved. 

`Hecke.small_group` was not the right choice if we only consider the group number` (#) `and order `(l)` of the group .  I considered `small_group` as suggested by Lars, but it works, if we are not given a very specific presentation. 

In this case,  a key piece of information, **presentation column**, is necessary to construct such abelian and non-abelian groups. Simply using `Hecke.small_group,` in this case, will give wrong results.

As pointed out by Tommy, this will be done via `**Free_Group**` from `Oscar.` Table 1 is attached below: 

Screenshot_select-area_20241018104735

PR #393, opened 2024-10-16
fix incompatibility with Julia 1.11.1 and remove the adjoint piracy

royess:
In this PR, the adjoint piracy is removed, which is related to #371 .

After this removal, an inner function `group_algebra_conj` is used in favor of reloading `Base.adjoint`. Some comments are also updated and corrected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,11 @@
 
 # News
 
-## v0.9.11
+## v0.9.12 - 2024-10-18
+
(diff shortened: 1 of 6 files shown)


royess:
@Krastanov please help review this PR. Thanks!

royess:
@Krastanov The new commit fixes the bug reported in #394, an incompatibility problem with Julia 1.11.1.

The fix may affect performance for large codes and require optimization in the future.

Krastanov:
Wonderful, thank you!

PR #397, opened 2024-10-18
Non-Abelian Dihedral Groups via Group Presentation 

Fe-r-oz:
This PR reproduces the Table 3 of the Lin and Pyradko paper: https://arxiv.org/pdf/2306.16400  for non-abelian Dihedral Groups. The structure, order, group identification of Dihedral groups has also been tested along with n, k parameters. 

Since Oscar v1.2.0 has not been released yet, Oscar v1.1.1 will throw GAP.jl CI errors because GAP version is v10.4.0. In this version, GAP does not support multi-threaded garbage collection. 

All the results from this Table 3 has been reproduced.  I think documentation about this will be added in future as well.

Screenshot_select-area_20241018201243


test/test_ecc_dihedral2bga.jl
```diff
@@ -0,0 +1,294 @@
+@testitem "ECC 2BGA Reproduce Table 3 of arXiv:2306.16400" tags=[:ecc, :ecc_bespoke_checks] begin
+    @static if !Sys.iswindows() && Sys.ARCH == :x86_64 && VERSION >= v"1.11"
```
Krastanov, line 2: I am not a bit fan of this conditional being tucked inside of a test file. It makes it difficult to track whether there is definitely a CI runner that executes these tests. A conditional in `runtests.jl` and an appropriate extra tag will make this easier to organize.

Fe-r-oz: I will look into this style of setting up conditional in `runtests.jl` over the weekend.

Krastanov:
> > How much do these tests add to the test runtime? More like a minute or more like 10 minutes?
> 
> More like a minute. I ran the entire tests before pushing. The non-abelian 2BGA codes are small codes, so the distance runner does not take much time to find out the distance.

Thanks!

PR #372, opened 2024-09-29
tests: test consistency of Bivariate Bicycle code using LPCode constructor

Fe-r-oz:
@Krastanov, @royess, please find the Lifted Product Construction of Bivariate Bicycle codes, here the abelian group is : ℤₗ x ℤₘ where ℤⱼ cyclic group of order j. Utilized [cyclic_group](https://docs.oscar-system.org/dev/Groups/pcgroup/#cyclic_group) from Oscar.

ECC Zoo refers to these codes as one of the several (two block group algebra codes) 2BGA codes. Since, we have the constructor for 2BGA, and based on Bravyi description (https://github.com/QuantumSavory/QuantumClifford.jl/pull/352#issuecomment-2379941360), it seems this should be simple when considering the LP construction. There are other construction methods that does not rely on LP construction: 1) the circulant matrix construction in GF(2) or 2) where Group algebra (GA) can be defined as quotient of a bivariate polynomial ring. 

The literature says that BB belong to class of Generalized Bicycle Codes (GB), and GB codes were further generalized into two block group algebra codes (2BGA). That's why ECC Zoo says they are one of several 2BGA codes. 

Initial output:
```
julia> import Hecke

julia> import Oscar

julia> using QuantumClifford.ECC: bivariate_bicycle_codes

julia> bivariate_bicycle_codes([3, 1, 2], [3, 1, 2], 9, 6)
[...]

PR #399, opened 2024-10-21
Implementing Bivariate Bicycle Codes using 2BGA as the parent  via Hecke's Group Algebra

Fe-r-oz:
This PR implements the Bivaraite Bicycle code with parent being 2BGA. This construction is given by ECC Zoo: https://errorcorrectionzoo.org/c/qcga. This is canonically equivalent to #372. This is the third construction method for these codes.


Fe-r-oz:
This PR is ready for review @Krastanov, Thank you!

Tasks Completed:

- [x]  `bivariate_bicycle_codes`  API
- [x] Documentation and Docstring
- [x] Testing via `test_ecc.jl` that calls all other tests such as `test_decoder_all_setups.jl`
- [x] Reproducing three different tables from 3 separate papers 

After this PR is finished, I will add a canonical equivalence test to #372 using the `bivariate_bicycle_codes` API defined here to test the following: ` BB using LP code as parent == BB using 2BGA code as parent. ` They both are equivalent.
```mermaid
graph TD
    subgraph A[LP]
        B[2BGA]
        subgraph B[2BGA]
            C[BB]
        end
    end
```

ext/QuantumCliffordHeckeExt/lifted_product.jl
```diff
@@ -189,10 +189,48 @@ Bicycle codes are a special case of generalized bicycle codes,
...
+
+julia> B = [y^5 , x^3  , x^19];
+
+julia> c = bivariate_bicycle_codes(A,B,GA);
+
+julia> code_n(c), code_k(c)
+(756, 16)
+```
+"""
+function bivariate_bicycle_codes(A::Vector{GroupAlgebraElem{FqFieldElem, GroupAlgebra{FqFieldElem, FinGenAbGroup, FinGenAbGroupElem}}}, B::Vector{GroupAlgebraElem{FqFieldElem, GroupAlgebra{FqFieldElem, FinGenAbGroup, FinGenAbGroupElem}}}, GA::GroupAlgebra{FqFieldElem, FinGenAbGroup, FinGenAbGroupElem})
```
Fe-r-oz, line 231: We can make this more readable as follows:

```
function bivariate_bicycle_codes(A::VectorGroupAlgebraElem, B::VectorGroupAlgebraElem, GA::FqFieldFinGenAbGroupElemGroupAlgebra)
```
by defining these `const` types `VectorGroupAlgebraElem` and `FqFieldFinGenAbGroupElemGroupAlgebra` in `types.jl`.

ext/QuantumCliffordHeckeExt/lifted_product.jl
```diff
@@ -209,10 +209,48 @@ Bicycle codes are a special case of generalized bicycle codes,
...
+```jldoctest
+julia> import Hecke: group_algebra, GF, abelian_group, gens; # hide
+
+julia> l=21; m=18;
+
+julia> GA = group_algebra(GF(2), abelian_group([l, m]));
+
+julia> x, y = gens(GA);
+
+julia> A = [x^3 , y^10 , y^17];
```
Krastanov, line 241: why write this down as `[x^3 , y^10 , y^17]` instead of `x^3 + y^10 + y^17` and use directly `two_block_group_algebra_codes`? What is the advantage of the list notation?

Krastanov (changes requested):
This is probably the addition I am most excited about of the ones I reviewed today. Thank you for looking into it. I have a question about the API design choices.

ext/QuantumCliffordHeckeExt/lifted_product.jl
```diff
@@ -209,10 +209,46 @@ Bicycle codes are a special case of generalized bicycle codes,
...
+julia> c = bivariate_bicycle_codes(A,B);
+
+julia> code_n(c), code_k(c)
+(756, 16)
+```
+"""
+function bivariate_bicycle_codes(A::GroupAlgebraElem, B::GroupAlgebraElem)
+    c = two_block_group_algebra_codes(A,B)
+    return c
+end
```
Krastanov, line 254: a followup on my previous question: if now this is all that is needed, why even have a separate function, why not just have `two_block_group_algebra_codes` and add this as yet another example in the docstring, similar to the other PRs you have on this family of codes

I might be wrong, maybe there is some benefit to keeping a separate function for this class of codes, but such a separate function would need something "special" when compared to `two_block_group_algebra_codes`

ext/QuantumCliffordHeckeExt/lifted_product.jl
Fe-r-oz: You are right. For regular abelian groups via Hecke's group algebra, we can use the list notation.

The advantage of list notation is when using finitely presented groups from Oscar with specific group presentations which is not related to this PR. 

```
julia> b_elts = 1 +  s + r^6 +  s^3 * r + s * r^7 +  s^3 * r^5;
ERROR: MethodError: no method matching +(::Int64, ::FPGroupElem)
The function `+` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...)
   @ Base operators.jl:596
  +(::Oscar.StraightLinePrograms.Lazy, ::Any)
   @ Oscar ~/.julia/packages/Oscar/2lR8w/src/StraightLinePrograms/lazy.jl:98
  +(::Any, ::Oscar.StraightLinePrograms.Lazy)
   @ Oscar ~/.julia/packages/Oscar/2lR8w/src/StraightLinePrograms/lazy.jl:99
  ...

Stacktrace:
 [1] +(::Int64, ::FPGroupElem, ::FPGroupElem, ::FPGroupElem, ::FPGroupElem, ::FPGroupElem)
   @ Base ./operators.jl:596
 [2] top-level scope
   @ REPL[9]:1
```

ext/QuantumCliffordHeckeExt/lifted_product.jl
Fe-r-oz: Thank you for these suggestions. I applied the `Oscar.FPGroup` list notation that Tommy introduced for specific stuff and applied it to `Hecke` out of curiosity for no particular reason.

Fe-r-oz:
Thank you for your suggestions. The PR is ready for review!

By the way, your 2BGA enhancement made the group algebra elements `A` and `B` look much prettier.

PR #400, opened 2024-10-21
add QuantumCliffordOscarExt to provide  convenient API for finitely presented groups via specific group presentation

thofma:
@Fe-r-oz hope I did everything correct

PR #401, opened 2024-10-21
ci: tweak ci.yml

thofma:
After clogging the CI queue with many small commits in #400, I noticed that intermediate builds for pull requests are not cancelled. This fixes it.

@Krastanov If you want this, I can also add it to the other *.yml files.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,14 @@ on:
     branches: [master, main]
     tags: ["*"]
   pull_request:
+
+concurrency:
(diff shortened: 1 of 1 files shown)


Krastanov:
Thank you! This is very much appreciated, I did not know it exists as a feature. And if you have the bandwidth, I would certainly take you up on the offer to add it to the other workflows.

Krastanov:
@Fe-r-oz , thanks to this change, you can now disregard my request to batch multiple small commits in a single large push.

PR #406, opened 2024-10-27
Add documentation of using `LPCode` via Hecke's `small_group`

Fe-r-oz:
This PR introduces Hecke.small_group for 'abelian' 2BGA codes for single cyclic groups. Hecke.small groups are limited, they can only be used single cyclic groups.


Fe-r-oz:
Please help review this PR, Thank you!

Fe-r-oz:
Hi Stefan, just a friendly jump on this PR, Thank you!!

ext/QuantumCliffordHeckeExt/lifted_product.jl
```diff
@@ -265,6 +265,49 @@ julia> code_n(c), code_k(c)
...
+
+julia> G = small_group(l, group_id);
+
+julia> GA = group_algebra(GF(2), G);
+
+julia> r = prod(gens(GA));
+```
+
+!!! note When using `Hecke.small_group`, it is essential to verify that the
+presentation for the single cyclic group is satisfied before proceeding with
```
Krastanov: what does this mean "the presentation for the single cyclic group is satisfied"

ext/QuantumCliffordHeckeExt/lifted_product.jl
```diff
@@ -265,6 +265,49 @@ julia> code_n(c), code_k(c)
...
+the code construction. This method serves as a workaround for creating small
+groups, specifically for single cyclic groups, using a group presentation with
+*no extra relations*, such as `⟨r | r³⁰⟩`. For the construction of *general*
+groups with specific group presentations, the only effective method is to use
+*finitely presented groups* (`Oscar.FPGroup`), which allow for defining direct
+products of two or more *general* groups—something not supported by Hecke.
```
Krastanov: I do not think I understand what this says. Why is this a "workaround"? What are these "extra relations"? What does "general" and "specific" and "presentations" mean in "groups with specific group presentations"? I overall do not really understand

ext/QuantumCliffordHeckeExt/lifted_product.jl
```diff
@@ -265,6 +265,49 @@ julia> code_n(c), code_k(c)
 (108, 12)
 ```
 
+### Small Groups
+
+An abelian `[[60, 6, 10]]` 2BGA code  of order `l = 30` with group ID `4`, represented
+by the group presentation `⟨r | r³⁰⟩`, constructed via `Hecke.small_group(4,30)`. Note:
+Hecke's small groups are limited in scope and should only be used for single cyclic groups.
```
Krastanov: Is this example from a paper or you just made up some code? All the other examples reference papers.

test/test_ecc_small_groups.jl
```diff
@@ -0,0 +1,79 @@
+@testitem "ECC 2BGA Hecke Small Groups" begin
+    using Hecke: group_algebra, GF, abelian_group, gens, quo, one, GroupAlgebra, small_group
+    using QuantumClifford.ECC
+    using QuantumClifford.ECC: code_k, code_n, two_block_group_algebra_codes
+
+    @testset "Hecke Small Groups without extra relations for single cyclic groups" begin
```
Krastanov: What does "without extra relations for single cyclic groups" means?

ext/QuantumCliffordHeckeExt/QuantumCliffordHeckeExt.jl
```diff
@@ -5,7 +5,7 @@ using DocStringExtensions
 import QuantumClifford, LinearAlgebra
 import Hecke: Group, GroupElem, AdditiveGroup, AdditiveGroupElem,
     GroupAlgebra, GroupAlgebraElem, FqFieldElem, representation_matrix, dim, base_ring,
-    multiplication_table, coefficients, abelian_group, group_algebra, rand
+    multiplication_table, coefficients, abelian_group, group_algebra, rand, small_group
```
Krastanov: You do not seem to be using this, except in a docstring. Does it need to be imported?

Krastanov (changes requested):
This looks good, it is just docs/examples/tests, and should be easy to merge once clarified a bit. My main confusion is not being to answer why this was added to the docs. Some elaboration on why someone would want to do this would be helpful. Are these some new codes from the literature?

ext/QuantumCliffordHeckeExt/lifted_product.jl
Fe-r-oz: Thank you for your comments! My apologies, I hope this example clarifies the paragraph. Please let me know if there is anything that still needs clarification. 

Given a group  $G=⟨S∣R⟩$ with $S$ set of generators and $R$ set of relations that the generators satisfy, the two examples are as follows:

Let's take this code as example: 

`[[60, 10, 9]]` abelian 2BGA code with group presentation $⟨r|r^{30}⟩$ from  from https://arxiv.org/pdf/2306.16400. There are no [additional/extra](https://math.stackexchange.com/a/1711747) relations that describe this group $C_{30}$
```
julia> l = 30;
julia> group_id = 4;
julia> G = small_group(l, group_id);
julia> GA = group_algebra(GF(2), G);
julia> r = prod(gens(GA)); 
julia> r^30  ==  1 # presentation satisfied
true
```

This means that `Hecke.small_group` can reliably describe groups without additional relations, typically the single cyclic groups, from table 1, block 1 and block 3. We will see later some discrepancies in case of multiple groups.

Let's take this code as example: 

`[[72, 8, 9]]]` non-abelian 2GBA code that is defined by a group presentation $⟨r, s|s^4, r^9, s^−1rsr⟩$ with extra relations $s^−1rsr$. This group is called $C_9 ⋉ C_4$ (semidirect product of cyclic group of order 9 and cyclic group of order 4). This code is taken from table 1, fourth row block from https://arxiv.org/pdf/2306.16400

The accurate way to create this group is through a finitely presented group (`Oscar.free_group`) since we are given a specific group presentation to work with and we have to satisfy this group presentation given the generators and relations of the group. 

In code, this is as follows:

```
julia> using QuantumClifford.ECC; using Oscar; using QuantumClifford

julia> m = 9; n = 4;

julia> F = free_group(["s", "r"]);

julia> s, r = gens(F);

julia> G, = quo(F, [s^n, r^9, s^-1*r*s*r]);

julia> GA = group_algebra(GF(2), G);

julia> s, r = gens(G);

julia> s^4 == r^9 == s^-1*r*s*r # the group presentation given this set of generators and relations is satisfied
true

julia> gens(GA) # later we will see that Hecke.small_group does have the generators that satisfy this presentation
2-element Vector{GroupAlgebraElem{FqFieldElem, GroupAlgebra{FqFieldElem, FPGroup, FPGroupElem}}}:
 [0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 [0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]

julia> a_elts = [one(G), s, r, s*r^6];

julia> b_elts = [one(G), s^2*r, s^2*r^6, r^2];

julia> a = sum(GA(z) for z in a_elts);

julia> b = sum(GA(z) for z in b_elts);

julia> c = two_block_group_algebra_codes(a,b);

julia> code_n(c), code_k(c)
(72, 8)
```

The wrong way is to create the group $C_9 ⋉ C_4$ without satisfying the presentation. This can lead to 2BGA code that has parameters that does not satisfy the states` [[72, 8, 9]]`. We will see how we end up creating a `[[72, 10, d]]`  code instead which does not satisfy the group presentation. 

```
julia> Oscar.describe(Hecke.small_group(36,1))
"C9 : C4"
julia> G = Hecke.small_group(36,1)
Pc group of order 36
julia> Oscar.describe(Hecke.small_group(36,1))
"C9 : C4"
julia> G = Hecke.small_group(36,1)
Pc group of order 36
julia> GA = group_algebra(GF(2), G);
julia> r, s = gens(GA);
julia> gens(GA)
4-element Vector{GroupAlgebraElem{FqFieldElem, GroupAlgebra{FqFieldElem, PcGroup, PcGroupElem}}}:
 [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 [0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 [0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 [0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
``` 

> On a small tangent: We can inspect that this set of generators is missing the following generator which might be reason why this group does not satisfy the group presentation as we see later. This generator is the second generator from the first example.

> ``` 
> [0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
> ```   

Now let's carry on with the second example:
```     
julia> s^4 == r^9 == s^-1*r*s*r # presentation not satisfied
false
julia> r, s = gens(GA)[3], gens(GA)[4];
julia> s^4 == r^9 == s^-1*r*s*r
false
julia> r, s = gens(GA)[4], gens(GA)[3];
julia> s^4 == r^9 == s^-1*r*s*r
false
julia> r, s = gens(GA)[2], gens(GA)[3];
julia> s^4 == r^9 == s^-1*r*s*r
false
julia> r, s = gens(GA)[3], gens(GA)[4];
julia> s^4 == r^9 == s^-1*r*s*r
false
julia> r, s = gens(GA)[1], gens(GA)[4];
julia> s^4 == r^9 == s^-1*r*s*r
false
```

Since the group presentation is not satisfied, the result code is not` [[72, 8, 9]]` but some other `[[72, 10, d]]` code.
 
```
julia> G = Hecke.small_group(36,1)
Pc group of order 36
julia> Oscar.describe(Hecke.small_group(36,1))
"C9 : C4"
julia> G = Hecke.small_group(36,1)
Pc group of order 36
julia> GA = group_algebra(GF(2), G);
julia> r, s = gens(GA);
julia> a = 1 + s + r + s*r^6;
julia> b = 1 + s^2*r + s^2*r^6 + r^2;
julia> c = two_block_group_algebra_codes(a,b);
julia> code_n(c), code_k(c)
(72, 10)
```


> **Edit**: We have to manually verify which set of generators satisfy the group presentation. After a lot of manual searching, we can find the groups that satisfy the group presentation and other criteria. We can check the order of subgroups and whether these subgroups indeed generated G, we have to see whether this group generator the After manual search, we found:

> ```
> julia> G = Hecke.small_group(36,1)
> Pc group of order 36
> julia> GA = group_algebra(GF(2), G);
> julia> s, r = gens(GA)[1], gens(GA)[4];
> julia> s^4 == r^9 == s^-1*r*s*r
> true
> julia> a = 1 + s + r + s*r^6;
> julia> b = 1 + s^2*r + s^2*r^6 + r^2;
> julia> c = two_block_group_algebra_codes(a,b);
> julia> code_n(c), code_k(c)
> (72, 24) # wrong, some other code than required
> julia> gens(G)
> 4-element Vector{PcGroupElem}:
>  f1
>  f2
>  f3
>  f4
> julia> H, _  = sub(G, [gens(G)[1], gens(G)[4]]);
> julia> H == G
> false
> julia> H, _  = sub(G, [gens(G)[4], gens(G)[1]]);
> julia> H == G
> false
> julia> order(gens(G)[1])
> 4
> julia> order(gens(G)[2])
> 2
> julia> order(gens(G)[3])
> 9
> julia> order(gens(G)[4])
> 3
> julia> H, _  = sub(G, [gens(G)[1], gens(G)[3]]);
> julia> H == G
> true
> julia> H, _  = sub(G, [gens(G)[1], gens(G)[3]]);
> julia> H == G
> true
> julia> s, r = gens(GA)[1], gens(GA)[3];
> julia> s^4 == r^9 == s^-1*r*s*r
> true
> julia> a = 1 + s + r + s*r^6;
> julia> b = 1 + s^2*r + s^2*r^6 + r^2;
> julia> c = two_block_group_algebra_codes(a,b);
> julia> code_n(c), code_k(c)
> (72, 8) # this matches with the paper
> ```


So, what I wanted to convey in this paragraph was that when we are given the extra relations such as  $s^−1rsr$, we should first satisfy whether the generators of the group satisfy the group presentation with these extra relations, and only then form a code, otherwise we will end up constructing some other code instead of the required one.  

When we have two or more groups, `Hecke`  do not support this feature, even with `Hecke.small_group` since it's internally assuming some group presentation even though it creates the same $C_9 : C_4$ but it comes with different set of generators that we saw earlier which leads to discrepancy at time of code formation. We can still do a lot of manually inspection to find which set of generators satisfy the group presentation as shown in the above example. (the example added in edit version) 

> Despite this, there are some nice use cases of Hecke.small_group. One is that we can create some non-abelian LP codes which can be used to test whether LP Code support non-commutative algebras. I have used small_groups in #511 for this purpose.

Krastanov: This explanation helps a lot. In terms of the docstring you have added, I would indeed like to keep these improvements, but can we reorganize them a bit.

Could you add a paragraph similar to this (near the beginning of the section you have added):

> Two group algebra codes are defined by specifying polynomials made out of group generators (hence the "group algebra" in the name). It is not sufficient to just pick a group, rather you need specific choices for the generators, usually done by specifying a "group presentation"[wiki link for group presentation]. A "group presentation" is a set of generators together with a list of relations obeyed by these generators. The Oscar package has the tooling necessary to define such presentations succinctly and directly, but it is a rather heavy package. The much lighter package Hecke provides for an easy way to get sets of generators for many "small groups", but it does not necessary provide the generators that obey the relations we want (as there can be many different sets of generators for the same group) -- nonetheless, if we manually confirm the relations, we can use Hecke directly.
>
> Below we show how you can use the lighter package Hecke to pick some of the pre-defined "small groups" in it, however, as we mentioned, it is important to verify that the set of generators you get is actually the one obeying the relations specific to the presentation we want. All examples are of codes discovered in [appropriate citation].
>
> ... all your examples go hear
>
> And now we do the same directly with Oscar
>
> ... new additional examples here, which basically repeat the same code as above, but use Oscar
> 
> ... including a check that the parity checks from the Hecke example are the same as the parity checks from the Oscar example


Additionally, I would suggest showcasing only one single code as an example -- no need to have multiple ones, as long as that single example is easy to understand. More examples can go to the test files.

Fe-r-oz: Thank you 😃 Apologies for the delay on this.

I added the additional tests using `Hecke.small_group` which passes in the CIs where `Oscar` is not imported. But in the CIs where  `Oscar` is also imported to test the `Oscar` related stuff, when using  `Hecke.small_group` which is of type `MultTableGroup` and has different set of generators,  it ends up calling  `Oscar.small_group ` which is of type` PcGroup` with different set of generators.

So, I removed the additional tests in https://github.com/QuantumSavory/QuantumClifford.jl/pull/406/commits/1592609b89f39df3720a6c8674766b557401b987 which pass on CIs only using `Hecke`. For CIs that import Oscar, these test throw error. The example shows how using using `Hecke` and` Oscar` in the same scope causes issue that led me to remove additional tests

```
julia> using Hecke;

julia> G = Hecke.small_group(36,1)

Generic group of order 36 with multiplication table
julia> GA = group_algebra(GF(2), G);

julia> gens(GA)
2-element Vector{GroupAlgebraElem{FqFieldElem, GroupAlgebra{FqFieldElem, MultTableGroup, MultTableGroupElem}}}:
 [0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 [0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]

julia> using Oscar

julia> G = Hecke.small_group(36,1) # this ends up calling Oscar.small_group since Hecke small group is of different type as shown above
Pc group of order 36

julia> gens(GA);

julia> GA = group_algebra(GF(2), G);

julia> gens(GA) # since these are different generator sets, it causes error because we are working with Hecke, but code ends up calling Oscar.small_group
4-element Vector{GroupAlgebraElem{FqFieldElem, GroupAlgebra{FqFieldElem, PcGroup, PcGroupElem}}}:
 [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 [0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 [0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 [0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```

The can be seen from the CI as well: [here](https://github.com/QuantumSavory/QuantumClifford.jl/actions/runs/15833264693/job/44630884377?pr=406#step:7:389). The doctest that  is added in documentation fails because when `Oscar` is imported, the doctest ends up calling `Oscar.small_group` but the doctest was prepared using Hecke.small_group.

```
julia> using QuantumClifford.ECC; using QuantumClifford;
julia> import Hecke: small_group, gens, group_algebra, GF; import Oscar; # to replication the error in CI
julia> m, n = 4, 9;
julia> l = 36;
julia> group_id = 1;
julia> G = Hecke.small_group(l, group_id);
julia> GA = group_algebra(GF(2), G);
julia> s, r = gens(GA);
julia> s^m == r^n == s^-1*r*s*r
false
julia> a = 1 + s + r + s*r^6;
julia> b = 1 + s^2*r + s^2*r^6 + r^2;
julia> c = two_block_group_algebra_codes(a,b);
julia> code_n(c), code_k(c)
(72, 36)
```

I am happy to add back the testset with additional examples if we can allow the Oscar related to 3 CIs to throws errors when using `Hecke.small_group` since that ends up calling `Oscar.small_group`. The tests only fail in CIs where `Oscar` is imported: [here](https://github.com/QuantumSavory/QuantumClifford.jl/actions/runs/15833264693/job/44630884377?pr=406), [here](https://github.com/QuantumSavory/QuantumClifford.jl/actions/runs/15833264693/job/44630884387?pr=406), and [here](https://github.com/QuantumSavory/QuantumClifford.jl/actions/runs/15833264693/job/44630884412?pr=406)

Please let me know how to proceed! Thank you!

**Edit:** removed grammatical errors in the post, and polished it.

> **Edit 2**: In this case, maybe when `Oscar` and `Hecke` are both imported, skip this test.

>```
>    i̶f̶ i̶s̶d̶e̶f̶i̶n̶e̶d̶(̶M̶a̶i̶n̶, :̶O̶s̶c̶a̶r̶)̶ &̶&̶ i̶s̶d̶e̶f̶i̶n̶e̶d̶(̶M̶a̶i̶n̶, :̶H̶e̶c̶k̶e̶)̶
>        @̶i̶n̶f̶o̶ "S̶k̶i̶p̶p̶i̶n̶g̶ H̶e̶c̶k̶e̶ S̶m̶a̶l̶l̶ G̶r̶o̶u̶p̶s̶ t̶e̶s̶t̶s̶ b̶e̶c̶a̶u̶s̶e̶ b̶o̶t̶h̶ O̶s̶c̶a̶r̶ a̶n̶d̶ H̶e̶c̶k̶e̶ a̶r̶e̶ l̶o̶a̶d̶e̶d̶"
>       r̶e̶t̶u̶r̶n̶
>    e̶n̶d̶
>```

Update: Since Oscar small group is called when both hecke and oscar are used, I used `@static if Sys.iswindows() && VERSION < v"1.11"` so that the tests only run on Hecke

Fe-r-oz:
The doctests give error when using both Oscar and Hecke. The doctest is not using Oscar but still the doctests end up assuming Oscar.small_group which causes the doctest to fail because the generators were found and relations satisfied using Hecke.small_group. If there is a way to just make sure the specific doctest only use Hecke, not Oscar, that would be much better than using julia instead of jldoctest which disabled the doctest running. The doctest actually pass when we just use Hecke in interactive environment. Please let me know your feedback, Thank you!

Krastanov:
Which doctest?

Additionally, could you structure an example of the failure, for discussion here, as follows:

```
using Hecke
... # some setup code
output1 = function_call(some_arguments)
using Oscar
output2 = function_call(some_arguments)
# and now output1 and output2 are not the same
```

Fe-r-oz:
I forgot to mention that on the interactive envionment, there is indeed a fix for it which is we pass an additional parameter, Hecke.DefaultSmallGroupDB(), I think the function that is called is `Hecke.small_group`
```
julia> using QuantumClifford.ECC; using QuantumClifford;
julia> import Hecke: small_group, gens, group_algebra, GF, DefaultSmallGroupDB;
julia> m, n = 4, 9;
julia> l = 36;
julia> group_id = 1;
julia> G = small_group(l, group_id; DB = Hecke.DefaultSmallGroupDB())
julia> GA = group_algebra(GF(2), G);
julia> s, r = gens(GA);
julia> s^m == r^n == s^-1*r*s*r
true
julia> a = 1 + s + r + s*r^6;
julia> b = 1 + s^2*r + s^2*r^6 + r^2;
julia> c = two_block_group_algebra_codes(a,b);
julia> code_n(c), code_k(c)
(72, 8)
```

If we now use call Oscar, the doctest will pass since it's not calling `Oscar.small_group`. I think after the CI runs, I will attach the error here if it occurs on CI now.

Fe-r-oz:
Due to the error/discrepancy demonstrated in https://github.com/QuantumSavory/QuantumClifford.jl/pull/406#issuecomment-3028534813, and the work around mentioned in https://github.com/QuantumSavory/QuantumClifford.jl/pull/406#issuecomment-3028575549, I had added the following note to reflect this: The comments in the `()` are just for reference for this discussion. 

When both `Oscar` and `Hecke` are imported, `Oscar`’s `small_group` function overrides `Hecke`’s (as shown here, https://github.com/QuantumSavory/QuantumClifford.jl/pull/406#issuecomment-3028534813, only when we pass the order (l) of the group and the Id of the group).  While `Hecke`’s version returns a `MultTableGroup`, `Oscar`’s returns a `PcGroup` with a different set of generators. If we’ve manually verified that the generators satisfy the presentation using `Hecke`’s `small_group`, we should ensure `Oscar` is not imported, as it may lead to unexpected results due to the override. We can use a workaround (as shown in https://github.com/QuantumSavory/QuantumClifford.jl/pull/406#issuecomment-3028575549) —namely, `Hecke.DefaultSmallGroupDB` — to still employ `Hecke.small_group`, though this is not guaranteed to last forever.

The performance tracking error is unrelated to this PR. It occurs because qeccore version is defined as "`version number `missing ending `" `. This was introduced in the bump commit: https://github.com/QuantumSavory/QuantumClifford.jl/commit/1ea5e64d1956851542477f6733a1ad3f5c0aca93. I have submitted a PR in this regard as well.

test/test_ecc_small_groups.jl
```diff
@@ -0,0 +1,161 @@
+@testitem "ECC 2BGA Hecke Small Groups" tags=[:ecc] begin
+    @static if Sys.iswindows() && VERSION < v"1.11"
```
Krastanov: Why running only on windows at versions older than 1.11?

test/test_ecc_small_groups.jl
```diff
@@ -0,0 +1,161 @@
...
+            # [[72, 8, 9]]
+            l = 36
+            group_id = 2
+            G = small_group(l, group_id)
+            GA = group_algebra(GF(2), G)
+            r = prod(gens(GA))
+            @test r^36  ==  1 # presentation ⟨r|r³⁶⟩ satisfied
+            A = 1 + r^28
+            B = 1 + r + r^18 + r^12 + r^29 + r^14
+            c = two_block_group_algebra_codes(A,B)
```
Krastanov: we renamed this to not be plural anymore

PR #407, opened 2024-10-29
add `check_repr_regular_linear` to verify `F`-linear regular representation for group algebra elements

Fe-r-oz:
This PR provides a useful consistency and diagnostic check that verifies whether F-linear regular representation condition is met. 

**Edit:** 

**Reference [paper](https://arxiv.org/pdf/2306.16400), page 5**

> It is easy to verify that for group elements $g ∈ G$, matrices $L(g)$ form the regular F -linear representation of $G$. Further, for any $a, b ∈ F [G]$, $L(a) L(b) = L(ab)$, $R(a) R(b) = R(ba)$, while any two matrices from different sets commute with each other, $L(a) R(b) = R(b) L(a)$; it is the latter property that gives the CSS orthogonality condition.


nzy1997:
This PR appears to be simple and correct. One minor suggestion: It might be helpful to add a failing test case for check_repr_regular_linear.

Fe-r-oz:
Thanks! I’ve added support for custom `repr` functions, since `LiftedCode` allows users to supply their own custom representations. While `check_repr_regular_linear` will always pass for `Hecke`’s default `representation_matrix`, it would become valuable for users to validate their custom `repr`. For example, fake repr that fails the test that returns a matrix of 1s:

```
function _fake_repr(a::GroupAlgebraElem, side=:left)
     d = length(group(parent(a)))
     return fill(1, d, d) # matrix of 1s
end
G = small_group(36, 1)
GA = group_algebra(GF(2), G)
res = check_repr_regular_linear(GA; repr=_fake_repr)
false
```

nzy1997:
Appreciate the changes! Everything looks great！

Fe-r-oz:
Thanks!

## 

**PR summary**: This PR provides a consistency/property check that verifies whether F-linear regular representation condition is met, helpful when user provides custom representations (`repr`), instead of Hecke's `representation_matrix`.

**cc:** @Krastanov

Krastanov:
Thanks!

PR #408, opened 2024-10-29
noncliff: conjugate destabs

Fe-r-oz:
Based on the `_stabmixdestab` function, the mixture now demonstrates the effect of the `CliffordOperator`, specifically updating the destabs entries.

Based on the current logic, this seems reasonable; hopefully, it is logically sound. 

Before:
```
julia> sm = GeneralizedStabilizer(S"-X")
A mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is
𝒟ℯ𝓈𝓉𝒶𝒷
+ Z
𝒮𝓉𝒶𝒷
- X
with ϕᵢⱼ | Pᵢ | Pⱼ:
 1.0+0.0im | + _ | + _

julia> apply!(sm, CliffordOperator(tHadamard))
A mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is
𝒟ℯ𝓈𝓉𝒶𝒷
+ X
𝒮𝓉𝒶𝒷
- Z
with ϕᵢⱼ | Pᵢ | Pⱼ:
 1.0+0.0im | + _ | + _
```
After:

```
julia> apply!(sm, CliffordOperator(tHadamard))
A mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is
𝒟ℯ𝓈𝓉𝒶𝒷
+ X
𝒮𝓉𝒶𝒷
- Z
with ϕᵢⱼ | Pᵢ | Pⱼ:
 1.0+0.0im | + X | + X
```

### DIFF
--- a/src/nonclifford.jl
+++ b/src/nonclifford.jl
@@ -89,6 +89,19 @@ function _stabmixdestab(mixeddestab, d)
     p
 end
 
+function _conj_destabs(state::GeneralizedStabilizer)
+    dict = state.destabweights
+    dtype = valtype(dict)
+    tzero = zero(dtype)
+    newdict = typeof(dict)(tzero)
+    for ((dᵢ, dⱼ), χ) in dict
+        dᵢ′ = .!dᵢ
+        dⱼ′ = .!dⱼ
+        newdict[(dᵢ′, dⱼ′)] += χ
+    end
+    state.destabweights = newdict
+end
+
 """
 Apply a Clifford gate to a generalized stabilizer state, i.e. a weighted sum of stabilizer states.
 
@@ -109,13 +122,14 @@ A mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is
 𝒮𝓉𝒶𝒷
 - Z
 with ϕᵢⱼ | Pᵢ | Pⱼ:
- 1.0+0.0im | + _ | + _
+ 1.0+0.0im | + X | + X
 ```
 
 See also: [`GeneralizedStabilizer`](@ref)
 """
-function apply!(state::GeneralizedStabilizer, gate::AbstractCliffordOperator) # TODO conjugate also the destabs
+function apply!(state::GeneralizedStabilizer, gate::AbstractCliffordOperator)
     apply!(state.stab, gate)
+    _conj_destabs(state)
     state
 end
 


src/nonclifford.jl
```diff
@@ -109,13 +122,14 @@ A mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is
 𝒮𝓉𝒶𝒷
 - Z
 with ϕᵢⱼ | Pᵢ | Pⱼ:
- 1.0+0.0im | + _ | + _
```
Krastanov, line 112: I am not sure I follow. Why is `_conj_destabs` necessary? Isn't this now giving a wrong or more-complicated result?

Fe-r-oz: I implemented the `TODO` based on your comment:

```
function apply!(state::GeneralizedStabilizer, gate::AbstractCliffordOperator) # TODO conjugate also the destabs
```
In the formula shown in the screenshot, the paper illustrates how a Clifford circuit `C `updates both the stabilizers `S` and destabilizers `D` by conjugation. When we apply `C`and `C†`, we effectively perform a change of basis on the stabilizer and destabilizer terms, transforming them according to `C`.

Screenshot_select-area_20241102120807



To see why `conj_destabs!`  is necessary, let’s examine what happens with the terms `dᵢ` and `dⱼ`:

- In the equation, each `dᵢ` and `dⱼ`​ term is conjugated by `C` as `CdᵢC†` and `CdⱼC†`.
- This conjugation process updates `dᵢ` and `dⱼ`​ to reflect the transformation under the Clifford circuit, keeping them aligned with the updated stabilizers `S`.
- By conjugating both `S` and `D`,  we maintain the overall structure of the full tableau, ensuring that the entire state evolves consistently under the Clifford operation.

This is why your comment "conjugate also the destabs" is relevant. Without it, the destabilizers `dᵢ` and `dⱼ` wouldn’t be updated correctly in response to a Clifford gate, leading to an inaccurate representation of the destabweights.

Krastanov (changes requested):
Thanks, that clarifies why some change in this direction is necessary, but it does not answer my question about correctness. Here are a few things that work on `HEAD~1` -- the commit just before the one in this PR:

```julia-repl
julia> using QuantumClifford, QuantumOpticsBase

julia> s = S"-X"; sm = GeneralizedStabilizer(s);

julia> Ket(s)
Ket(dim=2)
  basis: Spin(1/2)
  0.7071067811865475 + 0.0im
 -0.7071067811865475 + 0.0im

julia> Operator(sm)
Operator(dim=2x2)
  basis: Spin(1/2)
  0.5+0.0im  -0.5+0.0im
 -0.5+0.0im   0.5+0.0im

julia> dm(Ket(s)) ≈ Operator(sm)
true

julia> Operator(tHadamard)*Operator(sm)*Operator(tHadamard)'
Operator(dim=2x2)
  basis: Spin(1/2)
 0.0+0.0im  0.0+0.0im
 0.0+0.0im  1.0+0.0im

julia> Operator(tHadamard)*Operator(sm)*Operator(tHadamard)' ≈ Operator(apply!(s),tHadamard))
true
```

and here are what they look like after this PR:


```julia-repl
julia> using QuantumClifford, QuantumOpticsBase

julia> s = S"-X"; sm = GeneralizedStabilizer(s);

julia> Ket(s)
Ket(dim=2)
  basis: Spin(1/2)
  0.7071067811865475 + 0.0im
 -0.7071067811865475 + 0.0im

julia> Operator(sm)
Operator(dim=2x2)
  basis: Spin(1/2)
  0.5+0.0im  -0.5+0.0im
 -0.5+0.0im   0.5+0.0im

julia> dm(Ket(s)) ≈ Operator(sm)
true

julia> Operator(tHadamard)*Operator(sm)*Operator(tHadamard)'
Operator(dim=2x2)
  basis: Spin(1/2)
 0.0+0.0im  0.0+0.0im
 0.0+0.0im  1.0+0.0im

julia> Operator(tHadamard)*Operator(sm)*Operator(tHadamard)' ≈ Operator(apply!(sm,tHadamard))
false

julia> Operator(sm) # careful, we modified sm in the above line
Operator(dim=2x2)
  basis: Spin(1/2)
 1.0+0.0im  0.0+0.0im
 0.0+0.0im  0.0+0.0im
```

Could you confirm, the last two commands above give the wrong result with this PR, right?

Krastanov:
a couple more notes:

- Your "conjugation" step does not seem to depend on the gate C. That seems suspicious unless there is some trick related to why it does not need to depend.
- A potential trick that might matter here is that the destabilizers are already represented in terms of rows inside of the tableau structure. So it is quite possible that the conjugation is already implicitly performed anyway.

Krastanov:
If it turns out that the conjugation is already being performed, you can just add tests and remove the todo.

Fe-r-oz:
darn.... Thank you for the tests. I should have tried these tests before opening the PR...

> A potential trick that might matter here is that the destabilizers are already represented in terms of rows inside of the tableau structure. So it is quite possible that the conjugation is already implicitly performed anyway.

Yeah. Indeed, based on the tests results before applying this PR.

I think I am just copying the destabs in place of `dᵢ `and `dⱼ​ `to which does indeed refer to your comment _Your "_conjugation" step does not seem to depend on the gate C._ I agree with the assessment.


In light of the feedback, let's just close the PR.

Issue #411, opened 2024-10-31
Apply a PauliMeasurement to a PauliFrame


J-C-Q:
**Describe the bug 🐞**

It is not possible to apply a PauliMeasurement to a PauliFrame.

**Expected behavior**

Using the `pftrajectories()` function, I thought I would be able to simulate a circuit that includes PauliMeasurements, like `PauliMeasurement(P"ZZ", 1)`

**Minimal Reproducible Example 👇**

```julia
circuit = [sHadamard(1),sHadamard(2),PauliMeasurement(P"ZZ", 1)]
pftrajectories(circuit; trajectories=1000)
```

**Error & Stacktrace or other complete output produced by the MRE ⚠️**

```julia
┌ Warning: Could not compactify the circuit, falling back to a slower version of the simulation. Consider reporting this issue to the package maintainers to improve performance. The offending gate was `PauliMeasurement{Array{UInt8, 0}, Vector{UInt64}}(+ ZZ, 1)`.
└ @ QuantumClifford ~/.julia/packages/QuantumClifford/haClW/src/pauli_frames.jl:203
ERROR: MethodError: no method matching apply!(::PauliFrame{Stabilizer{…}, Matrix{…}}, ::PauliMeasurement{Array{…}, Vector{…}})
The function `apply!` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  apply!(::Register, ::PauliMeasurement{A, B}) where {A, B}
   @ QuantumClifford ~/.julia/packages/QuantumClifford/haClW/src/classical_register.jl:86
  apply!(::Register, ::Any, Any...; kwargs...)
   @ QuantumClifford ~/.julia/packages/QuantumClifford/haClW/src/classical_register.jl:36
  apply!(::QuantumClifford.AbstractQCState, ::NoisyGate)
   @ QuantumClifford ~/.julia/packages/QuantumClifford/haClW/src/noise.jl:125
  ...

Stacktrace:
 [1] pftrajectories(state::PauliFrame{Stabilizer{QuantumClifford.Tableau{Vector{UInt8}, LinearAlgebra.Adjoint{UInt64, Matrix{…}}}}, Matrix{Bool}}, circuit::Vector{QuantumClifford.AbstractOperation})
   @ QuantumClifford ~/.julia/packages/QuantumClifford/haClW/src/pauli_frames.jl:229
[...]

PR #412, opened 2024-11-01
Simulate circuits containing PauliMeasurements using PauliFrames

J-C-Q:
This PR adds a layer of abstraction that uses option 2 of #411 to make it easier to simulate circuits containing PauliMeasurements using PauliFrames.

- The PauliFrame contractor adds an extra qubit (at the end), which then gets used as an ancillary qubits for the PauliMeasurements.
- The `apply!` function has a method for PauliMeasurements that basically reuses `ECC.naive_ancillary_paulimeasurement` to convert a PauliMeasurement to a list of gates (which already have `apply!` methods) that perform the measurement using the ancillary qubit.
- Since PauliMeasurement is parametric, add a `concrete_typeparams` method for PauliMeasurements to make `CompactifiedGate` possible. (I'm not sure if the one possible type parameter I specify here is exhaustive... probably not.)


src/pauli_frames.jl
```diff
@@ -120,6 +121,24 @@ function apply!(frame::PauliFrame, op::sMRZ) # TODO sMRY, and faster sMRX
...
+        if op.pauli[qubit] == (1, 0)
+            apply!(frame, sXCZ(qubit, n + 1))
+        elseif op.pauli[qubit] == (0, 1)
+            apply!(frame, sCNOT(qubit, n + 1))
+        elseif op.pauli[qubit] == (1, 1)
+            apply!(frame, sYCX(qubit, n + 1))
+        end
+    end
+    op.pauli.phase[] == 0 || apply!(frame, sX(n + 1))
+    apply!(frame, sMRX(n + 1, op.bit))
```
Krastanov, line 137: I think this should be `sMRZ`

src/pauli_frames.jl
```diff
@@ -26,7 +26,8 @@ $(TYPEDSIGNATURES)
 Prepare an empty set of Pauli frames with the given number of `frames` and `qubits`. Preallocates spaces for `measurement` number of measurements.
 """
 function PauliFrame(frames, qubits, measurements)
-    stab = fastcolumn(zero(Stabilizer, frames, qubits)) # TODO this should really be a Tableau
+    # one extra qubit for ancilla measurements
+    stab = fastcolumn(zero(Stabilizer, frames, qubits + 1)) # TODO this should really be a Tableau
```
Krastanov, line 30: I agree with this change. I think now we need to update `show` and `nqubits` and maybe others. Thankfully, for `show` it can be as simple as using `@view frame[:,1:end-1]`

src/pauli_frames.jl
```diff
@@ -120,6 +121,24 @@ function apply!(frame::PauliFrame, op::sMRZ) # TODO sMRY, and faster sMRX
     return frame
 end
 
+function apply!(frame::PauliFrame, op::PauliMeasurement)
+    # this is inspired by ECC.naive_ancillary_paulimeasurement. Not sure if it's better to import and call that.
```
Krastanov, line 125:
```suggestion
    # this is inspired by ECC.naive_syndrome_circuit
```

Better to keep it the way you did it here, as this is non-allocating while `naive_syndrome_circuit` would create a temporary array

Krastanov (changes requested):
This is awesome, thank you!

I think we just need to add some tests (potentially including a test that uses `@allocated` to verify that memory is not churned).

src/pauli_frames.jl
Krastanov: actually, we do not even have a show method implemented, so this probably only needs a correct `nqubits`

src/pauli_frames.jl
J-C-Q: Of course...

src/pauli_frames.jl
J-C-Q: I subtracted the ancillary qubit in the `nqubits(f::PauliFrame)` function. I think that should be fine.

Krastanov:
The test failures happen due to this: https://github.com/QuantumSavory/QuantumClifford.jl/blob/f69e49ccfb91d7fd85bda16a846c34d22c500399/test/test_ecc_syndromes.jl#L33-L36

The cleanest way to fix this is probably to do `mul_left!(tab(naive_frames), pₙ)` instead of `mul_left!(naive_frames.frame, pₙ)`

with a newly defined

```
tab(p::PauliFrame) = @view p.frame[:,1:end-1] # to account of the buffer ancillary qubit
```

J-C-Q:
Somehow this is more complicated because the `view` implementations for `PauliFrame`, `Stabilizer` and `Tableau` only allow one dimensional indexing as far as I can tell.

Krastanov:
darn, you are right. Thankfully, here we are adding a qubit at the very end, so there is still an easy workaround

```
tab(pf::PauliFrame{Stabilizer{Tableau{V,F}}, M}) where {V,F,M} = Tableau{V,F}(pf.frame.tab.phases, pf.frame.tab.xzs, nqubits(pf))
```

If I am not getting this wrong, the internal indexing utilities of `Tableau` will take care of skipping over the extra qubit. This would need some testing though

J-C-Q:
Sadly, this doesn't seem to work.

I guess the internal indexing doesn't skip over the last qubit. And removing it from the `xzs` reference is not feasible...

I don't think there are many other options... Probably, silently adding one extra qubits is not such a good idea.

Krastanov:
I still like the way you have set this up and I suspect there is something pretty minor left to fix. I should be able to look into it around the end of the week.

Krastanov:
finally got some time to look into this. Just as an FYI, the only example that causes errors are the LP118 family of LDPC codes (which also happen to be the largest codes we are considering, so it is probably related to some off-by-one error when we need more than 1 uint word to store the qubits)

Krastanov:
The fix ended up being extremely simple -- just adding a +1 to the size of a random pauli generated in one of the tests (due to the new buffer qubit). This is an internal detail that was used in the test, not breaking and public API.

But then I spent another hour or two adding extra tests just to make sure nothing was missed.

If the tests pass here, this should be merged and released shortly.

Thank you so much for the contribution! Do not hesitate to suggest other improvements as you play with this tool

J-C-Q:
Thank you for the incredible effort you put into this! Your dedication and hard work are truly appreciated. It was a pleasure to collaborate on such a fantastic project!

PR #413, opened 2024-11-02
noncliff: add tests for conjugate destabs

Fe-r-oz:
This PR removes the TODO comment and add tests for conjugation. This PR is inspired from this comment: https://github.com/QuantumSavory/QuantumClifford.jl/pull/408#issuecomment-2452981604

### DIFF
--- a/test/test_nonclifford_quantumoptics.jl
+++ b/test/test_nonclifford_quantumoptics.jl
@@ -51,35 +51,25 @@ qo_tgate.data[2,2] = exp(im*pi/4)
     end
 end
 
-@testset "Conjugate destabs" begin
-    test_cases = [
-        (1, [tHadamard, tPhase, tId1]),
-        (2, [tCNOT, tCPHASE, tSWAP]),
-        (3, [enumerate_cliffords(3, clifford_cardinality(3)), CliffordOperator(sHadamard(3), 3), CliffordOperator(sCNOT(1, 2), 3)]),
-        (4, [enumerate_cliffords(4, clifford_cardinality(4)), CliffordOperator(sHadamard(4), 4), CliffordOperator(sCNOT(2, 1), 4)]),
-        (5, [enumerate_cliffords(5, clifford_cardinality(5)), CliffordOperator(sHadamard(5), 5), CliffordOperator(sCNOT(2, 3), 5)])
-    ]
-
-    for (num_qubits, gates) in test_cases
-        @testset "Conjugate destabs test using $num_qubits-qubit Clifford gate" begin
-            for _ in 1:10
-                s = random_stabilizer(num_qubits)
-                sm = GeneralizedStabilizer(s)
-                @test dm(Ket(s)) ≈ Operator(sm)
-                for C in gates
-                    @test Operator(C) * Operator(sm) * Operator(C)' ≈ Operator(apply!(sm, C))
-                end
-            end
+@testset "apply!" begin
+    for n in [1,2,3,4] # exponential cost in this term
+        s = random_stabilizer(n)
+        sm = GeneralizedStabilizer(s)
+        @test dm(Ket(s)) ≈ Operator(sm)
+        # test clifford gates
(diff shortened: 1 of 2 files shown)


test/test_nonclifford_quantumoptics.jl
```diff
@@ -50,3 +50,15 @@ qo_tgate.data[2,2] = exp(im*pi/4)
         end
     end
 end
+
+@testset "conjugate destabs" begin
+    @testset "Single qubit Clifford gate" begin
+        for s in [S"X", S"Y", S"Z", S"-X", S"-Y", S"-Z"]
```
Krastanov, line 56: to make the tests more trustworthy we should have multi-qubit cases and cases of non-clifford operations -- it can be in a separate testset to keep this already nice testset easy to read. The extra testset can contain stuff like:

```
for n in [5, 10, 100]
    for g in [tHadamard, tPhase, pcT]
        eg = embed(g,...)
        sm = GeneralizedStabilizer(random_stabilizer(n))
        ...
```

Krastanov (changes requested):
That would be very useful, thank you! I added a comment on a few improvements that can be made to make the tests more rigorous

test/test_nonclifford_quantumoptics.jl
Fe-r-oz: `embed` is not defined for `CliffordOperator`, only for `PauliOperator`, so instead used: `enumerate_cliffords(4, clifford_cardinality(4))` for multi-qubit` CliffordOperators`. I hope you like it.

Fe-r-oz:
Thank you for your speedy comments and insights. That is much appreciated! 

I hope this is satisfactory. `n=100` crashed my PC so just used `n=10` max.

test/test_nonclifford_quantumoptics.jl
Fe-r-oz: Instead of `embed` for `CliffordOperator`, we can use `CliffordOperator(sHadamard(2), 5)`... darn...

test/test_nonclifford_quantumoptics.jl
```diff
@@ -50,3 +50,36 @@ qo_tgate.data[2,2] = exp(im*pi/4)
...
     end
 end
+
+@testset "Conjugate destabs" begin
+    test_cases = [
+        (1, [tHadamard, tPhase, tId1]),
+        (2, [tCNOT, tCPHASE, tSWAP]),
+        (3, [enumerate_cliffords(3, clifford_cardinality(3)), CliffordOperator(sHadamard(3), 3), CliffordOperator(sCNOT(1, 2), 3)]),
+        (4, [enumerate_cliffords(4, clifford_cardinality(4)), CliffordOperator(sHadamard(4), 4), CliffordOperator(sCNOT(2, 1), 4)]),
+        (5, [enumerate_cliffords(5, clifford_cardinality(5)), CliffordOperator(sHadamard(5), 5), CliffordOperator(sCNOT(2, 3), 5)])
```
Krastanov, line 60: this seems unnecessarily complicated. And the number of qubits is something that does not need to be specified separately

in particular, is there something special about `enumerate_cliffords(n, clifford_cardinality(n))` that you are picking this choice

I will submit a few simplification suggestions later today

Fe-r-oz: > is there something special about enumerate_cliffords(n, clifford_cardinality(n)) that you are picking this choice

Nope. No particular reason for the choice of parameters for `enumerate_cliffords`.


Thank you!

Krastanov: to keep in the spirit of having a random check you can also just do `enumerate_clifford(3, rand(1:clifford_cardinality(3)))`, but at this point it is just simpler to stick to `random_stabilizer`

Krastanov:
thanks for pushing through this, it is very helpful cleanup. Let's wait to see what the new tests say

Issue #151, opened 2023-07-11
implement `embed` for all types


Krastanov:
Currently embed is implemented for `PauliFrames`.

We need it for all tableaux, and for dense clifford operators, and for single- and two-qubit gates.

Fe-r-oz:
I think there is a small typo here: Currently `embed` is implemented for `PauliOperator` not `PauliFrames` at that moment in time. It is the `paulioperator.jl`.

PR #415, opened 2024-11-02
embed for single-qubit and two-qubit symbolic Clifford Operators

Fe-r-oz:
The PR #413 revealed to us that we don't have `embed` for single-qubit  dense/symbolic Clifford Operator or two-qubit dense/symbolic Clifford Operator.  Thus, this PR attemps to define `embed` for single-qubit Clifford Operator, thereby addressing a chunk of #151

Edit:

I think is is straightforward as we have already defined implicit `embed` as `CliffordOperator(sCNOT(1, 2), 4)` for instance! 

```
julia> embed(5, (2,3), sCNOT)
X₁ ⟼ + X____
X₂ ⟼ + _XX__
X₃ ⟼ + __X__
X₄ ⟼ + ___X_
X₅ ⟼ + ____X
Z₁ ⟼ + Z____
Z₂ ⟼ + _Z___
Z₃ ⟼ + _ZZ__
Z₄ ⟼ + ___Z_
Z₅ ⟼ + ____Z
```

### DIFF
--- a/src/symbolic_cliffords.jl
+++ b/src/symbolic_cliffords.jl
@@ -416,6 +416,42 @@ LinearAlgebra.inv(op::sInvSQRTXX) = sSQRTXX(op.q1, op.q2)
 LinearAlgebra.inv(op::sSQRTYY)    = sInvSQRTYY(op.q1, op.q2)
 LinearAlgebra.inv(op::sInvSQRTYY) = sSQRTYY(op.q1, op.q2)
 
+"""
+Embed a single-qubit Clifford operator in a larger Clifford operator.
+
+```jldoctest
+julia> embed(5, 4, sHadamard)
+X₁ ⟼ + X____
+X₂ ⟼ + _X___
+X₃ ⟼ + __X__
+X₄ ⟼ + ___Z_
+X₅ ⟼ + ____X
+Z₁ ⟼ + Z____
+Z₂ ⟼ + _Z___
+Z₃ ⟼ + __Z__
+Z₄ ⟼ + ___X_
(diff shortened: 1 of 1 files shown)


Fe-r-oz:
The PR is ready for review. Thank you!

Krastanov (changes requested):
Thanks for looking into this. Do we need an embed for these though? They already work fine on arbitrary number of qubits, i.e. they are already implicitly embedded. You can run `apply!(stab, sCNOT(1,2))` on any size `stab` (if anything, an embed here would make thinks slower). You can not do that with `apply!(stab, tCNOT)` so an embed is needed. 

I guess adding this functionality makes sense just for completeness, but there should be a warning in the docstring that this is a bad way to do it.

Also, it seems there is a single docstring for one of the methods that includes what is supposed to be a docstring for the other method. This should be two separate docstrings (if they are method dosctrings).

Lastly, the single-qubit embed should still permit one-tuples as argument types for `i` to be consistent with the rest of the `embed` methods

Krastanov:
also, new functionality like this needs tests

Fe-r-oz:
Thank you for your comments.

> Do we need an embed for these though? 

Actually, I considered closing this PR since the embedding is already handled implicitly. While it’s worth exploring alternative approaches while looking into #151, the final outcome here doesn’t offer much added value. 

Instead of adding code that SLOW DOWN performance || is a BAD way to do things || might introduce redundant checks, we can simply update the documentation to clarify that the embedding is implicitly supported.

This also seems like an inefficient use of your valuable review time.

So, let's just close it :)

PR #417, opened 2024-11-03
noncliff: enhancements to printing

Fe-r-oz:
There is extra space that is printed in REPL when printing genstab. This is noticeable as there is a chasm between Pi and rho. Please do try on REPL, I hope it's not a linux thing.  This PR provides a consistent printing scheme that matches with how destab and stabs are printed and fixes this printing gap in REPL.

Before:
Screenshot_select-area_20241103145851

Screenshot_select-area_20241103150340

After:
The pretty printing is now consistent with the stab, destab printing, and there is no extra space in REPL.

Screenshot_select-area_20241103145922

Screenshot_select-area_20241103150409

### DIFF
--- a/src/nonclifford.jl
+++ b/src/nonclifford.jl
@@ -14,31 +14,31 @@
 """
 $(TYPEDEF)
 
-Represents mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is a pure stabilizer state.
+Represents mixture 𝛴 𝜙ᵢⱼ 𝒫ᵢ 𝜌 𝒫ⱼ† where 𝜌 is a pure stabilizer state.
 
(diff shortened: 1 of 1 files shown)


Fe-r-oz:

Krastanov:
The calligraphic P is indeed prettier, but it is not a common choice for denoting down an operator in this subfield of science. The calligraphic S and D kinda make sense because they are commonly used that way. I would prefer to stick to that convention to avoid confusion.

The extra space seems to not exist in the source code, so it is probably an issue with some terminals not supporting unicode fonts well. That is something we should not be fixing internally here but report as bugs in the terminal emulators.

The Gnome terminal seems to be ok:
image


I will close this for now, but do not hesitate to re-open if similar issues appear

PR #419, opened 2024-11-03
noncliff: add copy, == for GeneralizedStabilizer, UnitaryPauliChannel

Fe-r-oz:
This PR is inspired from this comment: https://github.com/QuantumSavory/QuantumClifford.jl/pull/355#discussion_r1826568457

### DIFF
--- a/src/nonclifford.jl
+++ b/src/nonclifford.jl
@@ -64,6 +64,8 @@ function GeneralizedStabilizer(state)
 end
 
 GeneralizedStabilizer(s::GeneralizedStabilizer) = s
+Base.copy(sm::GeneralizedStabilizer) = GeneralizedStabilizer(copy(sm.stab),copy(sm.destabweights))
+Base.:(==)(sm₁::GeneralizedStabilizer, sm₂::GeneralizedStabilizer) = sm₁.stab==sm₂.stab && sm₁.destabweights==sm₂.destabweights
 
 function Base.show(io::IO, s::GeneralizedStabilizer)
     println(io, "A mixture ∑ ϕᵢⱼ Pᵢ ρ Pⱼ† where ρ is")
@@ -145,8 +147,11 @@ with ϕᵢⱼ | Pᵢ | Pⱼ:
  0.853553+0.0im | + _ | + _
  0.146447+0.0im | + Z | + Z
 
-julia> expect(P"-X", sm)
+julia> χ′ = expect(P"-X", sm)
 0.7071067811865475 + 0.0im
+
+julia> prob = (real(χ′)+1)/2
+0.8535533905932737
 ```
 
 """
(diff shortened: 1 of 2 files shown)


src/nonclifford.jl
```diff
@@ -342,6 +347,8 @@ struct UnitaryPauliChannel{T,S,P} <: AbstractPauliChannel
 end
 
 PauliChannel(p::UnitaryPauliChannel) = p.paulichannel
+Base.copy(p::UnitaryPauliChannel) = deepcopy(p)
```
Krastanov, line 350: using `deepcopy` as an implementation of `copy` is a bit dangerous. It frequently works fine, but if you have views into some other memory it can produce unexpected results. The more boilerplate-heavy "Constructor on copied slot entries" is much safer here.

Krastanov (changes requested):
That is very useful, thank you! I have a minor worry about the use of deepcopy, explained below.

src/nonclifford.jl
Fe-r-oz: My bad...

Fe-r-oz: Since tuples are immutable, I used `p.weights` or `p.paulis` instead of `copy(p.weights)`, `copy(p.paulis)` etc.

src/nonclifford.jl
```diff
@@ -342,6 +347,8 @@ struct UnitaryPauliChannel{T,S,P} <: AbstractPauliChannel
 end
 
 PauliChannel(p::UnitaryPauliChannel) = p.paulichannel
+Base.copy(p::UnitaryPauliChannel) = UnitaryPauliChannel(p.paulis, p.weights)
+Base.:(==)(p₁::UnitaryPauliChannel, p₂::UnitaryPauliChannel) = p₁.paulis==p₂.paulis && p₁.weights==p₂.weights
```
Krastanov, line 351: The tuples might be immutable, but the objects they contain are not-necessarily immutable. E.g. if I do something stupid like `mul_left!(p.paulis, some_other_pauli)` on a copied `p`, the original `p` would also be modified. I am stupid to do that to begin with, but nonetheless I would be right to be surprised that modifying the copy modified the original as well.

More generally, let the compiler worry about immutability unless you know for a fact that the compiler is doing a bad job. The compiler can figure out when immutability guarantees let it reuse memory:

Lastly, you are probably doing this because `Tuple` does not have a `copy` method. That is a frustration I have with julia -- the story behind it can be found starting in this discussion: https://discourse.julialang.org/t/why-can-i-deepcopy-nothing-but-not-copy-nothing/112429 (in particular see the links by stevengj).

Anyway, to make sure this is an actual copy, the clunky but necessary `map(copy, p.paulis)` can be used.

Krastanov (changes requested):
You encountered an interesting quirk of julia when you were trying to copy a tuple. See the comment I added below.

src/nonclifford.jl
Fe-r-oz: Thank you for these insights! I avoided using `copy(p.paulis)` so it wouldn’t look stupid.

Krastanov:
thank you!

PR #422, opened 2024-11-07
noncliff: Cauchy Schwarz Inequality check  for GeneralizedStabilizers

Fe-r-oz:
This PR implements the  Cauchy Schwarz Inequality for GeneralizedStabilizers. I have a test that check this for `n =1:10`

Screenshot_select-area_20241107133753

Inner product between two Generalized Stabilizers is the LHS: $Tr[sm'sm]$  without the absolute value. 

Except about importance of inner product algorithm from TJ Yoder:

_The inner product algorithm,  allows us to determine whether two generalized stabilizer states are equal. This is not always a trivial thing to do, because two generalized stabilizers with different stabilizer bases and different χ-matrices may represent the same state._

Edit:

The inner product is implemented in #423 as well.

### DIFF
--- a/ext/QuantumCliffordQOpticsExt/QuantumCliffordQOpticsExt.jl
+++ b/ext/QuantumCliffordQOpticsExt/QuantumCliffordQOpticsExt.jl
@@ -6,6 +6,7 @@ using QuantumOpticsBase
 using Graphs
 using DocStringExtensions
 import QuantumOpticsBase: Ket, Operator
+import QuantumClifford: cauchy_schwarz_check
 
 const _b2 = SpinBasis(1//2)
 const _l0 = spinup(_b2)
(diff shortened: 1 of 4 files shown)


Krastanov (changes requested):
That would indeed be a usable check. But as mentioned in the PR that implements the dot product, this should not be relying on exponentially-expensive conversion into a density matrix.

This should not be exported either, as it is only for tests (generally it is important to be very conservative with exports, as otherwise you are promising to keep a stable API).

Given that it is internal and single-purpose, let's also rename it to something more detailed. E.g. `generalizedstab_cauchyschwarz_check`.

I will mark this as a draft as it can not be finished until the other related PR is finished.

Fe-r-oz:
Indeed. Thank you for your comments!

Fe-r-oz:
Thank you, I will keep these points in mind. I need to implement stabilizer union (theorem 16) first for the fast, in-place implementation of inner product. Details at: https://github.com/QuantumSavory/QuantumClifford.jl/pull/423#issuecomment-2492741456

PR #416, opened 2024-11-03
noncliff: enhanced error message for inapplicable Project! of GeneralizedStabilizer

Fe-r-oz:
I have incorporated the relevant details from the paper to create an informed error message. This PR is inspired from this comment: https://github.com/QuantumSavory/QuantumClifford.jl/pull/355#discussion_r1826569150

I think we only need one `proj,` so removed `_proj₋ `or `_proj₊` in favor of `proj`. 

- [x] The code is properly formatted and commented.
- [x] Substantial new functionality is documented within the docs.
- [x] All new functionality is tested.
- [x] All of the automated tests on github pass.
- [x] We recently started enforcing formatting checks. If formatting issues are reported in the new code you have written, please correct them. <small>There will be plenty of old code that is flagged as we are slowly transitioning to enforced formatting. Please do not worry about or address older formatting issues -- keep your PR just focused on your planned contribution.</small>

PR #423, opened 2024-11-07
noncliff: Inner Product between two GeneralizedStabilizers

Fe-r-oz:
This PR implements the Inner Product between two Generalized Stabilizers. Inspired from V.Section B of the base paper.

### DIFF
--- a/ext/QuantumCliffordQOpticsExt/QuantumCliffordQOpticsExt.jl
+++ b/ext/QuantumCliffordQOpticsExt/QuantumCliffordQOpticsExt.jl
@@ -1,11 +1,12 @@
 module QuantumCliffordQOpticsExt
 
 using QuantumClifford
-import QuantumClifford: mul_left!, mul_right!
+import QuantumClifford: mul_left!, mul_right!, dot
 using QuantumOpticsBase
 using Graphs
 using DocStringExtensions
 import QuantumOpticsBase: Ket, Operator
+using LinearAlgebra
 
 const _b2 = SpinBasis(1//2)
 const _l0 = spinup(_b2)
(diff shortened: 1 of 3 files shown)


Fe-r-oz:
The PR is ready for review. There is an unrelated doctest errors that were fixed in #416 that is why one extra CI-error.

Thank you!

ext/QuantumCliffordQOpticsExt/QuantumCliffordQOpticsExt.jl
```diff
@@ -234,6 +235,41 @@ function Operator(c::CliffordOperator)
     cliff_to_unitary(c)
 end
 
+"""
+The inner product of two [`GeneralizedStabilizer`](@ref) states, `sm₁` and `sm₂`.
+
+```jldoctest
+julia> using QuantumOpticsBase; using LinearAlgebra; # hide
```
Krastanov, line 242: do not hide these, otherwise the reader might not know how to run these examples

ext/QuantumCliffordQOpticsExt/QuantumCliffordQOpticsExt.jl
```diff
@@ -234,6 +235,41 @@ function Operator(c::CliffordOperator)
...
+ 0.0-0.353553im | + Z | + _
+ 0.853553+0.0im | + _ | + _
+ 0.146447+0.0im | + Z | + Z
+
+julia> dot(sm, sm)
+0.9999999999999994
+```
+"""
+function LinearAlgebra.dot(sm₁::GeneralizedStabilizer, sm₂::GeneralizedStabilizer)
+    return real(tr(Operator(sm₁)' * Operator(sm₂)))
```
Krastanov, line 270: This just converts the two generalized stabilizers to density matrices. This is an exponentially expensive operation that does not seem to be implementing what is shown in the paper. This can be a useful correctness test in the test suite, but it is not an appropriate implementation for this library.

test/test_nonclifford_quantumoptics.jl
```diff
@@ -82,6 +82,7 @@ end
         apply!(sm, embed(n, i, pcT))
         smcopy = copy(sm)
         @test smcopy == sm
+        @test dot(sm, smcopy) ≈ 1
```
Krastanov, line 85: having tests that do not always use the same state on both sides would make this much more trustworthy

Krastanov (changes requested):
Thanks for starting this! I think there is a bit of a misunderstanding of what the paper is actually doing. The approach implemented here is exponentially slower than the appropriate algorithm from the paper.

Fe-r-oz:
> I think there is a bit of a misunderstanding of what the paper is actually doing.

Yup. The stabilizer union between two generalized stabilizers is a prerequisite for the original in-place implementation described in the paper. The algorithm relies on the subroutine detailed in Theorem 16, called 'stabilizer union.' I think we can leverage the` pauli_group` (G(U)), as part of the process involves decomposing the generators and checking their presence in G(U). Given this, it makes sense to close this for now and prioritize completing and submitting the stabilizer union implementation first.

PR #425, opened 2024-11-13
Bicycle and Unicycle codes: Fixing doc and method errors

Benzillaist:
Hi! I finally got around to addressing the problems with my previous commits. I made a few changes:

- Updated doc strings to be consistent with the rest of the code
- Added appropriate citation links
- Fixed issue with outdated method calls in `reduce_unitary`

If you have the chance to review this and let me know if you want me to fix anything else, that would be great! Thanks so much :)


docs/src/references.bib
```diff
@@ -561,3 +561,22 @@ @article{haah2011local
   pages={042330},
   year={2011},
 }
+
+@article{mackay_sparse-graph_2004,
```
Krastanov, line 565: this reference already exists in the bib file, no need to add it again

test/test_ecc.jl
```diff
@@ -68,5 +68,10 @@
...
         @test isdegenerate(Steane7(), 2) == true
         @test isdegenerate(Bitflip3()) == true
+        @test isdegenerate(Bicycle(6, 4)) == true
+        @test isdegenerate(Bicycle(18, 16)) == true
+        @test isdegenerate(Bicycle(200, 120)) == true
+        @test isdegenerate(Unicycle(21, [1, 3, 8, 9, 12])) == true
     end
-end
+
+function
```
Krastanov, line 77: this seems to be causing a syntax error in the test runner

src/ecc/codes/simple_sparse_codes.jl
```diff
@@ -0,0 +1,337 @@
...
+        li = circ_arr[end]
+        circ_arr[2:end] = circ_arr[1:end-1]
+        circ_arr[1] = li
+    end
+    comp_matrix[1:n,1:n] = circ_matrix
+    comp_matrix[1:n,n+1] = one_col
+    return comp_matrix
+end
+
+"""Attempts to generate a list of indices to be used in a bicycle code using a search method"""
```
Krastanov, line 242: Where is this algorithm from? Could you add that to the docstring?

src/ecc/codes/simple_sparse_codes.jl
```diff
@@ -0,0 +1,337 @@
...
+            diff_arr = copy(temp_diff_arr)
+        end
+    end
+    return circ_arr
+end
+
+"""Attempts to generate a list of indices to be used in a bicycle code using a randomized check method
+
+Note: This is very slow for large N"""
+function bicycle_set_gen_rand(N::Int, d::Int)
```
Krastanov, line 288: Is this still used anywhere? If not, we should probably remove it.

Krastanov (changes requested):
Thank you, this is shaping really well! I have not reviewed everything yet, but here is a first pass of things that can be polished a bit.

Krastanov:
I converted it to draft to make it a bit easier for me to manage my review queue.

src/ecc/codes/simple_sparse_codes.jl
Benzillaist: I came up with it myself using the definition of a perfect difference set -- should I include a link to some article or definition on perfect difference sets?

src/ecc/codes/simple_sparse_codes.jl
Benzillaist: Sounds good! I'm not really using it for anything, potentially thought it could be used to see some interesting results, but I agree that it's not superbly useful

Benzillaist:
Hi Stefan, thanks for the review! I went back through and updated a few things with a focus on the function headers. Please let me know if there are any other things that you'd like me to check for!

docs/src/references.bib
```diff
@@ -560,4 +560,4 @@ @article{haah2011local
   number={4},
   pages={042330},
   year={2011},
-}
+}
```
Krastanov, line 563: A couple of files have their ending byte modified. For consistency and to keep the git history clean, it helps to set your editor to automatically take care of these.

In vscode for instance:

image

you can include these entries:

```
{
    "files.trimFinalNewlines": true,
    "files.trimTrailingWhitespace": true,
    "editor.rulers": [92],
    "files.eol": "\n"
}
```

src/ecc/codes/simple_sparse_codes.jl
```diff
@@ -0,0 +1,284 @@
...
+        li = circ_arr[end]
+        circ_arr[2:end] = circ_arr[1:end-1]
+        circ_arr[1] = li
+    end
+    comp_matrix[1:n,1:n] = circ_matrix
+    comp_matrix[1:n,n+1] = one_col
+    return comp_matrix
+end
+
+"""Attempts to generate a list of indices to be used in a bicycle code using a search method given a matrix width (N)"""
```
Krastanov, line 243: why "attempts"? Is this algorithm from the paper? Can you specify where in it if it is. If not, could you explain what the user should worry about with it?

src/ecc/codes/simple_sparse_codes.jl
```diff
@@ -0,0 +1,284 @@
...
+function bicycle_set_gen(N::Int)
+    circ_arr = Int[0]
+    diff_arr = Int[]
+    circ_arr[1] = 0
```
Krastanov, line 247: `circ_arr` already has its first element set to `0` in the first line

Krastanov (changes requested):
A few more points. Mostly interested in learning more about the generating algorithm for the bicycle case.

test/test_ecc_bivaraite_bicycle_as_twobga.jl
```diff
@@ -1,173 +1,173 @@
...
-        l=6; m=15
-        GA = group_algebra(GF(2), abelian_group([l, m]))
-        x, y = gens(GA)
-        A = x^3 + y   + y^2
-        B = y^6 + x^4 + x^5
-        c = two_block_group_algebra_codes(A,B)
-        @test code_n(c) == 180 && code_k(c) == 8
-    end
-end
+# @testitem "ECC Bivaraite Bicycle as 2BGA" begin
```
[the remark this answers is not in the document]
Benzillaist: Oops! Restored

Benzillaist:
Hi Stefan, I added a little description to the index generation method covering some information I thought was relevant! Please let me know if you think anything is missing from the description :)

src/ecc/codes/simple_sparse_codes.jl
```diff
@@ -0,0 +1,297 @@
...
++ ZZ_Z_Z
++ Z_Z_ZZ
+```
+"""
+struct Bicycle
+    "Width of array, should be >= 2. Equal to number to number of physical qubits (N)"
+    n::Int
+    "Height of array, should be >= 2 and a multiple of 2. Equal to number of parity checks (M)"
+    m::Int
+end
```
Krastanov, line 29: I think my main remaining request for change is related to the distinction between a structure and a function.

An instance of the type `Bicycle` should be a fully defined code where quickly and deterministically one can get a parity check matrix. The "quick" part is nice to have, but "deterministic" is mandatory.

Here, I think this is not deterministic. On the other hand the output of `bicycle_set_gen` seems to be all you need to be able to deterministically do the other steps. Thus I would suggest having `set` be an entry here, basically the way you have already done it for `Unicycle`.

That way both `Bicycle` and `Unicycle` will be fully defined codes by their fields, without any randomness. And the documentation can spell out the difference in what `set` needs to be. For `Bicycle` it needs to just be a set of integers (right? or maybe there are more constraints) while for `Unicycle` it has to be perfect difference set?

And in the docstring you can mention "we have a heuristic for finding useful sets in the function `yadayada`"

src/ecc/codes/simple_sparse_codes.jl
```diff
@@ -0,0 +1,297 @@
...
+
+function parity_checks(u::Unicycle)
+    n = u.N
+    set = u.set
+    usc = circ_to_unicycle_h0(set, n)
+    rusc = reduce_unicycle(usc) # reduced unicycle code
+    return parity_checks(CSS(rusc, rusc))
+end
+
+"""Takes an untrimmed bicycle matrix and removes the row which keeps the spread of the column weights minimal.
```
Krastanov, line 91: Is this something you invented or something from the paper? In either case, explain why it is needed. If it is from the paper, give a specific reference where in the paper it is.

src/ecc/codes/simple_sparse_codes.jl
```diff
@@ -0,0 +1,297 @@
...
+        li = circ_arr[end]
+        circ_arr[2:end] = circ_arr[1:end-1]
+        circ_arr[1] = li
+    end
+    comp_matrix[1:n,1:n] = circ_matrix
+    comp_matrix[1:n,n+1:2*n] = transpose(circ_matrix)
+    return comp_matrix
+end
+
+"""Takes an untrimmed unicycle matrix and removes linearly dependent rows.
```
Krastanov, line 164: same question about where this is coming from and why is it necessary

src/ecc/codes/simple_sparse_codes.jl
```diff
@@ -0,0 +1,297 @@
...
+        li = circ_arr[end]
+        circ_arr[2:end] = circ_arr[1:end-1]
+        circ_arr[1] = li
+    end
+    comp_matrix[1:n,1:n] = circ_matrix
+    comp_matrix[1:n,n+1] = one_col
+    return comp_matrix
+end
+
+"""
```
Krastanov, line 241: same question about provenance and invention and reference

Krastanov: if this is your own thing, you should brag about it a bit more loudly :D

src/ecc/codes/simple_sparse_codes.jl
```diff
@@ -0,0 +1,297 @@
...
+        tr = LinearAlgebra.rank(matrix(rrzz, tm))
+        if(tr == r)
+            m = tm
+            if(size(m)[1] == r)
+                break
+            end
+        else
+            i += 1
+        end
+    end
```
Krastanov, line 196: This is an n^4 algorithm, right? n^3 for the Gaussian elimination in `rank` and then another factor of n for the loop?

Add something to the documentation saying that "a naive implementation of row-reduction currently bottlenecks the use on very large codes".

src/ecc/codes/simple_sparse_codes.jl
```diff
@@ -0,0 +1,297 @@
...
+
+```
+"""
+function reduce_bicycle(H0::Matrix{Bool})
+    m, n = size(H0)
+    r_i = 0
+    std_min = Inf
+    for i in 1:m
+        t_H0 = vcat(H0[1:i-1, :], H0[i+1:end, :])
+        std_temp = std(convert(Array, sum(t_H0, dims = 1)))
```
Krastanov, line 114: The `convert(Array, ` is pretty suspicious. Do you need it?

Krastanov (changes requested):
Finally got around to reviewing, thank you for your patience!

I think this is the last round of review -- this is very close to merging.

Do you think you would have a chance of looking into the change requests? I would be interested in adding this code to the wiki

PR #438, opened 2024-11-29
fix: broken doctests and typos

Fe-r-oz:
This PR implements minor fixes in doctests.

Edit:

### DIFF
--- a/ext/QuantumCliffordHeckeExt/lifted_product.jl
+++ b/ext/QuantumCliffordHeckeExt/lifted_product.jl
@@ -219,6 +219,10 @@ julia> import Hecke: group_algebra, GF, abelian_group, gens; # hide
 
 julia> l=4; m=6;
 
+julia> GA = group_algebra(GF(2), abelian_group([l, m]));
+
+julia> x, y = gens(GA);
+
 julia> z = x*y;
 
 julia> A = x^3 + y^5;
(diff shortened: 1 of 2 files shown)


Krastanov:
These fixes are much appreciated! But I am also a bit confused: any idea how come these problems were not detected by the test runner?

Fe-r-oz:
The test runner didn't detect these errors because it did not construct the code to check the parameters: the parameters (108,12) were given to it on the platter.

https://github.com/QuantumSavory/QuantumClifford.jl/blob/7b444e90f2474f7e762ff2acdc70142c58bd25ca/ext/QuantumCliffordHeckeExt/lifted_product.jl#L254-L255
To the test runner, this was correct because it was the expected result. It didn't calculate the parameters to verify it, as the calculation step - the code construction => to calculate the parameters was missing. This was a silly error. My bad.

Fe-r-oz:
In the first doctest fix, the group algebra, `GA` and the generators `x` and `y` are not constructed. For multivariate bicycle code, the condition is to have a `z = x*y `as the third generator.  This doctest should have thrown an error because `x` and `y` are not provided to it when constructing the third generator. 

https://github.com/QuantumSavory/QuantumClifford.jl/blob/7b444e90f2474f7e762ff2acdc70142c58bd25ca/ext/QuantumCliffordHeckeExt/lifted_product.jl#L218-L230

Krastanov:
That does not seem to be the correct explanation. It seems these doctests, because they were in an extension, were actually never executed. Hopefully the last commit I pushed fixes that.

Thanks for flagging this issue! That would have been unresolved and untested for a long time otherwise. Good catch!

PR #439, opened 2024-11-30
compute the minimum distance of QLDPC using Mixed Integer Programming

Fe-r-oz:
This PR implements the Mixed Integer programming (MIP) approach to compute the minimum distance of QLDPC using GNU's Linear Programming Kit. This method is used by [Panteleev and Kalachev]( https://arxiv.org/pdf/1904.02703) and by [Bravyi et. al.](https://arxiv.org/pdf/2308.07915). The latter reference mentioned that this MIP approach was originally developed in 2011 and used in this [reference](https://arxiv.org/pdf/1108.5738)

This PR allows the users to compute the minimum distance with ease and also provide the necessary documentation along with the relevant modern usecases.  The tests have been conducted in #435

ILP/MILP methods: https://quantumcomputing.stackexchange.com/questions/37289/compute-the-exact-minimum-distance-of-a-qecc-with-integer-linear-programming-met

P.S. Mixed Integer Programming (MIP) or mixed integer linear program refer to same thing in the literature, in this context.

Edit:
- [x]  The code is properly formatted and commented.
- [x] Substantial new functionality is documented within the docs.
- [x] All new functionality is tested.
- [x] All of the automated tests on github pass.
- [x] We recently started enforcing formatting checks. If formatting issues are reported in the new code you have written, please correct them.

PR #463, opened 2025-01-20
fix: random circuit codes are non-CSS so return `iscss(...) = false` instead of `nothing`

Fe-r-oz:
Hi, @royess, The random_circuit codes are non-CSS so, I am wondering why the  `iscss(::Type{...})` returns `nothing`.  Shouldn't it return `false`? Hopefully, that is not a bad proposition?

Could you explain if it was a typo or there is some reason behind it? Thanks! I agree with you that codes are non-CSS. 

> BP and BP-OSD in PyQDecoders.jl (PyMatchingDecoder ) are not applicable to random circuit codes because they are non-CSS.

https://github.com/QuantumSavory/QuantumClifford.jl/blob/0d137912dc1f4828a3d1f68fdb211f66001d9793/src/ecc/codes/random_circuit.jl#L19

Bravyi introduces a Mixed Integer Programming for CSS/QLDPC codes so his approach don't work for non-CSS codes. Hence, I want to exclude non-CSS codes so user does not end up calculating `distance` for non-CSS codes which throw solver error.

`iscss(::Type{...}) = false` would be helpful here because in a later PR, I want to do something like

```
function distance(c::AbstractECC; kwargs...)
    if !iscss(c)
        throw(ArgumentError("Computing the minimum distance of QLDPC using Mixed Integer Programming works only for CSS codes."))
    end
    return distance(parity_checks(c); kwargs...)
end
```

@Krastanov, Please help review this PR, Thank you!

### DIFF
--- a/src/ecc/codes/random_circuit.jl
+++ b/src/ecc/codes/random_circuit.jl
@@ -16,7 +16,7 @@ struct CircuitCode <: AbstractECC
     encode_qubits::AbstractArray
 end
 
-iscss(::Type{CircuitCode}) = nothing
+iscss(::Type{CircuitCode}) = false
 
 code_n(c::CircuitCode) = c.n
 


royess:
@Krastanov and I discussed this during PR. The reason is that random circuit codes are not necessarily CSS but could still possibly be CSS.

Fe-r-oz:
There are a couple of places, (in the bullet point 2 in intro message) and also in the like I quoted where you mention these codes are non-CSS.

I will let Stefan decide the fate of this PR.

royess:
Let me clarify more: `CircuitCode`'s are non-CSS in general. (That is what I meant in the sentences you quoted.) But some instances of them could be CSS.

Therefore, `iscss(::Type{CircuitCode})` is undetermined, for which we assign `nothing`. If we return `false` all the time, we will make mistakes in these CSS instances.

Regarding your later PR: Does it make sense that you also exclude `nothing` when trying to calculate the code distance?

Fe-r-oz:
Thanks! Yeah, I can try checking !iscss(...) and also check if nothing which can also work and also does not change anything here. The PR is #439, it's a previous PR, but I ended up saying it's a later PR. 

I will close this PR to be safe.

PR #465, opened 2025-01-20
`[[4p, 2(p − 2), 4]]` Delfosse repetition code

Fe-r-oz:
The paper *[Short Shor-style syndrome sequences](https://arxiv.org/pdf/2008.05051)* by has many interesting generalized quantum codes from base that could be implemented. This PR implements one such code: `[[4p, 2(p − 2), 4]]` Delfosse repetition code.  This a one of the generalized codes that were developed by Delfosse and Reichardt. Interestingly, ECC Zoo don't have entries for these 'generalized codes' yet.   These codes are tested using the `test_ecc_base.jl` for properties checkup.  These codes were defined as `[[4p, 2(p − 2), 4]]`  codes so I named them `DelfosseRepCode`.  Internally, it uses extended blocks of [4,1,4] repetition code which is different in symmetry/structure from the `RepCode` that is already implemented.

Please find the following schematics for these codes:

display

display

- [x] The code is properly formatted and commented.
- [x] Substantial new functionality is documented within the docs.
- [x] All new functionality is tested.
- [x] All of the automated tests on github pass.
- [x] We recently started enforcing formatting checks. If formatting issues are reported in the new code you have written, please correct them.

PR #466, opened 2025-01-22
Delfosse-Reichardt codes from classical self-orthogonal Reed-Muller seeds

Fe-r-oz:
The paper [Short Shor-style syndrome sequences](https://arxiv.org/pdf/2008.05051) has many interesting generalized quantum codes from base that could be implemented. This PR implements the second code: `[[8p, 6(p-1), 4]]` Delfosse Generalized [8,4,4] code. This is one of the generalized codes that were developed by Delfosse and Reichardt. The first quantum stabilizer code is implemented in #465

Interestingly, ECC Zoo don't have entries for these 'generalized codes' yet. These codes are tested using the `test_ecc_base.jl` for properties checkup. These codes were defined as `[[8p, 6(p-1), 4]]` codes so I named them `DelfosseGeneralized844`. Internally, it uses extended blocks of [8,4,4] Reed Muller code.

Please find the following schematics for these codes:

display

What's left is to add some explicit tests from the paper.


Fe-r-oz:
Delfosse and Reichardt uses:

- **[8, 4, 4]** Reed-Muller code to create **[[8p, 6(p−1), 4 ]]** self-dual CSS quantum codes, for `p ≥ 2`
- **[16,11, 4]** Reed-Muller code to create **[[16p, 14p − 8, 4]]** self-dual CSS quantum codes, for `p ≥ 1`

Currently(https://github.com/QuantumSavory/QuantumClifford.jl/pull/466/commits/105877a5c5a8174c0caea9c7c9a5b3312d42cb0c), I have implemented the Delfosse Generalized [8,4,4] code as `[[8p, 6(p-1), 4]]` quantum stabilizer code. The next step is to create the second bullet-point code. 

To enhance the flexibility of the code, a better approach would be to generalize these codes using the Reed-Muller code as the base matrix. Instead of hardcoding the `[8,4,4]` or `[16,11, 4]` Reed-Muller codes as base matrices, we could allow users to input parameters `r` and `m`, making the code more generalized. 

So, we can **invent** :)  the following code: **[[8rp, (8r−2)p−2m, 4]] Delfosse Generalized ReedMuller code**

- `[8, 4, 4]` Reed-Muller code to create` [[8p, 6(p−1), 4 ]] `self-dual CSS quantum codes, for `p ≥ 2`
[8,4,4] base code, r =1, m =3

```julia
julia> parity_checks(ReedMuller(1,3))
4×8 Matrix{Bool}:
 1  1  1  1  1  1  1  1
 1  1  1  1  0  0  0  0
 1  1  0  0  1  1  0  0
 1  0  1  0  1  0  1  0
```

[[8×1×p = 8p,  (8-2)×p - 2×3 =  6p - 6 = 6(p-1), 4]] quantum stabilizer code which is the same as  `[[8p, 6(p−1), 4 ]] `

- `[16,11, 4]`  Reed-Muller code to create `[[16p, 14p − 8, 4]]` self-dual CSS quantum codes, for `p ≥ 1`

[16, 11,4]  base code, r =2, m =4

```julia
julia> parity_checks(ReedMuller(2,4))
5×16 Matrix{Bool}:
 1  1  1  1  1  1  1  1  1  1  1  1  1  1  1  1
 1  1  1  1  1  1  1  1  0  0  0  0  0  0  0  0
 1  1  1  1  0  0  0  0  1  1  1  1  0  0  0  0
 1  1  0  0  1  1  0  0  1  1  0  0  1  1  0  0
 1  0  1  0  1  0  1  0  1  0  1  0  1  0  1  0
```

[[8×2p = 16p,  (16-2)×p - 2×4 =  14p - 8, 4]] quantum stabilizer code which is the same as  `[[16p, 14p − 8, 4]] `

Fe-r-oz:
Please find the schematic for [[8rp, (8r − 2)p − 2m, 4]] Delfosse Generalized Reed-Muller code 
display

Fe-r-oz:
Please help review this PR, thank you so much :) 

There is a Julia error with `allocated` as it throws false negatives. So, I have commented out `allocated` that was throwing errors to avoid CI failures. 🙏🏼  Please find the detail about this Julia CI error here: https://github.com/QuantumSavory/QuantumClifford.jl/pull/467

src/ecc/codes/delfosse_reichardt_code.jl
```diff
@@ -0,0 +1,124 @@
...
+    end
+    return extended_H
+end
+
+function parity_matrix(c::DelfosseReichardt)
+    extended_mat = _generalize_delfosse_reichardt_code(c.blocks, c.r, c.m)
+    hx, hz = extended_mat, extended_mat
+    code = CSS(hx, hz)
+    Stabilizer(code)
+end
```
nzy1997: The `parity_matrix` function should return the parity check matrix, while `parity_checks` returns the corresponding Pauli strings. Our implementation provides a fallback mechanism that automatically derives the Pauli strings from the parity check matrix. Therefore, for new codes, you only need to define the `parity_matrix` - the fallback will handle the conversion to Pauli strings when `parity_checks` is called. 

https://github.com/QuantumSavory/QuantumClifford.jl/blob/b135bdabf3d718bd8ed3c63d299048e37aca0f4e/src/ecc/ECC.jl#L88

Additionally, for `AbstractCSSCode` types, we only require the definition of `parity_matrix_x` and `parity_matrix_z` - this function handles the rest.

https://github.com/QuantumSavory/QuantumClifford.jl/blob/b135bdabf3d718bd8ed3c63d299048e37aca0f4e/lib/QECCore/src/codes/quantum/css.jl#L47

In summary, you can simply remove this function and make minor adjustments to the tests - everything should work as expected.

Fe-r-oz: Thanks!

Should we remove this `parity_matrix` function that is in `Surface.jl` given that we can generate it via  `parity_matrix(c::AbstractCSSCode) = parity_matrix(CSS(parity_matrix_x(c), parity_matrix_z(c))`)?


https://github.com/QuantumSavory/QuantumClifford.jl/blob/23eea6410357c3b5588b84a8d1036a811ca81d91/lib/QECCore/src/codes/quantum/surface.jl#L64

nzy1997: Yeah, I think we can remove them, including `parity_matrix` in `toric.jl` and `surface.jl`.

nzy1997:
Thank you very much for moving the code into `QECCore.jl`. This PR looks fine to me!

src/ecc/codes/delfosse_reichardt_code.jl
Fe-r-oz: Removed `parity_matrix` for `Triangular488` and `Triangular666` as well.

Fe-r-oz:
Upon further investigation, the reason why paper used `RM(1,3)` and `RM(2,4`) is because these two codes are self-orthogonal, so the procedure results in self-dual CSS codes by ensuring commutating stabilizers. Arbitrary RM codes, as seen from the  code snippet, are not self-orthogonal, as it would result in invalid stabilizers,

```
julia> iszero(mod.(parity_matrix(ReedMuller(1,3))*parity_matrix(ReedMuller(1,3))',2)) # paper
true

julia> iszero(mod.(parity_matrix(ReedMuller(2,4))*parity_matrix(ReedMuller(2,4))',2)) # paper
true

julia> iszero(mod.(parity_matrix(ReedMuller(1,4))*parity_matrix(ReedMuller(1,4))',2)) # arbitrary
false

julia> iszero(mod.(parity_matrix(ReedMuller(1,5))*parity_matrix(ReedMuller(1,5))',2)) # arbitrary
false
```

In order to generalize this code construction a bit further, we can do searching for good parameters similar to what is done for BB or coprime BB codes. So, by using a function that searches for good parameters, we can generate many more code families than `[[8p, 6(p-1), 4]]` codes and` [[16p, 14p-8, 4]]`. Interestingly, these families still have distance of `4`.

```
function self_orthogonal_rm_codes(mmax::Int)
    good_params = Tuple{Int, Int}[] 
    for m in 1:mmax
        for r in 0:m
            try
                RM = ReedMuller(r, m)
                H = parity_matrix(RM)
                # Check self-orthogonality?
                if all(iszero, mod.(H * H', 2))
                    push!(good_params, (r, m))
                end
            catch
                continue 
            end
        end
    end
    return good_params
end
```

All of the code families construct with these `RM(r,m)` construct valid commutating stabilizers!
```
julia> self_orthogonal_rm_codes(10)
30-element Vector{Tuple{Int64, Int64}}:
 (0, 1)
 (1, 2)
 (1, 3)
 (2, 3)
 (2, 4)
 (3, 4)
 (2, 5)
 (3, 5)
 (4, 5)
 (3, 6)
 (4, 6)
 (5, 6)
 (3, 7)
 (4, 7)
 (5, 7)
 (6, 7)
 (4, 8)
 (5, 8)
 (6, 8)
 (7, 8)
 (4, 9)
 (5, 9)
 (6, 9)
 (7, 9)
 (8, 9)
 (5, 10)
 (6, 10)
 (7, 10)
 (8, 10)
 (9, 10)
```

PR #471, opened 2025-01-31
La-cross quantum LDPC codes

Fe-r-oz:
This PR implements **La-cross codes** from [High-rate quantum LDPC codes for long-range-connected neutral atom registers](https://www.nature.com/articles/s41467-025-56255-5). The [latest v2 version](https://arxiv.org/pdf/2404.13010) of this paper was published on **January 31, 2025**!  

The authors propose a **near-term implementation** of quantum LDPC codes in neutral atom registers that is both **static** (i.e., it does **not** require qubit shuttling) and **fast** compared to existing approaches. They demonstrate that these quantum LDPC codes outperform the surface code at the circuit level using only moderate nonlocal resources.  

This paper may be particularly relevant to Anthony's (@amicciche) work, as it is the first to propose:  

"_By targeting electromagnetic shifts at multiple lattice ranges, one can measure non-local stabilizers of quantum LDPC codes, **without needing to shuttle atoms**._" Please check it out — hopefully, you find it interesting! Thanks so much!

Please find the decoder graphs for these codes: :)

The first two schematics represents the codes from the appendix:
display
display

```julia
julia> codes
4-element Vector{Lacross}:
 Lacross(7, [1, 1, 0, 1], false)
 Lacross(7, [1, 1, 0, 1], true)
 Lacross(14, [1, 1, 0, 1], false)
 Lacross(7, [1, 1, 0, 1], true)
```

display

What's left is to improve the documentation as I just only skimmed the paper and add more tests. I have ensured that the construction is generalized and not limited to the specific codes they implemented.

### DIFF
--- a/docs/src/references.bib
+++ b/docs/src/references.bib
@@ -584,3 +584,14 @@
   journal={arXiv preprint arXiv:1811.06246},
   year={2018}
 }
+
+@article{pecorari2025high,
+  title={High-rate quantum LDPC codes for long-range-connected neutral atom registers},
+  author={Pecorari, Laura and Jandura, Sven and Brennen, Gavin K and Pupillo, Guido},
+  journal={Nature Communications},
+  volume={16},
+  number={1},
+  pages={1111},
+  year={2025},
+  publisher={Nature Publishing Group UK London}
+}
--- a/docs/src/references.md
+++ b/docs/src/references.md
@@ -45,6 +45,7 @@ For quantum code construction routines:
 - [lin2024quantum](@cite)
 - [bravyi2024high](@cite)
 - [haah2011local](@cite)
+- [pecorari2025high](@cite)
 
 For classical code construction routines:
 - [muller1954application](@cite)
--- a/src/ecc/ECC.jl
+++ b/src/ecc/ECC.jl
@@ -29,7 +29,7 @@ export parity_checks, parity_checks_x, parity_checks_z, iscss,
     Shor9, Steane7, Cleve8, Perfect5, Bitflip3,
     Toric, Gottesman, Surface, Concat, CircuitCode, QuantumReedMuller,
     LPCode, two_block_group_algebra_codes, generalized_bicycle_codes, bicycle_codes,
-    haah_cubic_codes,
+    haah_cubic_codes, Lacross,
     random_brickwork_circuit_code, random_all_to_all_circuit_code,
     evaluate_decoder,
     CommutationCheckECCSetup, NaiveSyndromeECCSetup, ShorSyndromeECCSetup,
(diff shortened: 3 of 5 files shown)


Fe-r-oz:
I think we should test the minimum distance for these codes as well :)

Fe-r-oz:
The implementation of these La-cross codes is completed and the codes are reproduced from the paper.  

- [ ] **TODO**: Test the minimum distance of La-cross quantum ldpc codes using MILP solver #439 as that would be a very valuable cross-verification check! 

 
Please help review this PR, Thank you so much! I hope you find this code exciting :)

nzy1997 (commented):
Thank you for your code. Everything looks good to me. I appreciate your detailed docstrings and comprehensive tests.

Fe-r-oz:
Thank you for your feedback and prompt reviews -- It is much appreciated. Apologies for any inconvenience caused by these last-minute commits that increases review overhead.

While adding changelog, I pushed some minor documentation/tests:

- documentation tweaks -- reference refer to these codes "La-cross", some sentences were using "La-Cross", so standardised this.
- previously we were only testing `k = 3` polynomials, we now test `k = 2, 3, and 4` (to test all variants of codes). 
- consistency checks that test `N,K` formulas with the general `code_n`, `code_k` that act on stabilizer.

 Documentation related Clifford deformation of La-cross codes is left for the future work.

ext/QuantumCliffordHeckeExt/lacross.jl
```diff
@@ -0,0 +1,182 @@
...
+julia> code_n(c), code_k(c), distance(c, DistanceMIPAlgorithm(solver=HiGHS))
+(400, 16, 8)
+```
+
+The ECC Zoo has an [entry for this family](https://errorcorrectionzoo.org/c/lacross).
+
+### Fields
+    $TYPEDFIELDS
+"""
+struct Lacross <: AbstractCSSCode
```
nzy1997: I think "La-cross" should be `LaCross` in PascalCase.

nzy1997: Just a minor note. Looks good to merge!

Fe-r-oz:
**PR Summary**

Implements `[[2n², 2k², d]]` and `[[(n - k)² + n², k², d]]` La-cross code using `Hecke`'s univariate polynomial ring

 **Future Work**

Test Clifford deformation on La-Cross codes [[reference]](https://arxiv.org/pdf/2502.20189) (see related issue #504). 

Context: Although Clifford-deformed La-Cross codes retain the same `[[N, K, D]]` parameters as their CSS counterparts, they yield non-CSS constructions. The aforementioned reference work does not apply code deformation to Bivariate Bicycle codes because "_these codes do not enjoy the same parity symmetry as XZZX/La-Cross codes_". Interestingly, this  referenced work generalizes the Clifford deformation framework from the XZZX surface code to arbitrary hypergraph product LDPC codes, including surface/La-Cross codes, which suggest broader applicability of the code deformation technique.

**cc:** @Krastanov

ext/QuantumCliffordHeckeExt/lacross.jl
Fe-r-oz: Changed to PascalCase!

PR #476, opened 2025-02-10
`[[n² + m², (n - rank([C ∣ M]))² + (m − rank([C ∣ M]ᵀ))², d]]` quantum Tillich-Zémor code

Fe-r-oz:
Dear Stefan,

Thank you so much for your valuable insights—they have been incredibly helpful in shaping this PR. I am also grateful to your cool feedback on [[8rp, (8r − 2)p − 2m, 4]] Delfosse-Reichardt code as well that helped me provide a better approach in this PR about new code construction (hopefully). The previous code was more of a generalization rather than a new construction. Your feedback made me realize I needed to improve the process and strive to do better than previous attempt. None of this would have been possible without your support and feedback.

With your insights in mind, I read [this paper](https://www.arxiv.org/pdf/2501.19125) on improved minimum distance bounds for structured LDPC codes, and it inspired me to explore the possibility of developing a new quantum LDPC code, namely **(n,m,r)-structured random quantum** LDPC codes. I would be deeply grateful for your feedback. I’ve shared a of summary via email for your review before submitting the PR as well.

Please find the following schematics for one sub-familty for these codes:
display

As you highlighted, what really matters is maintaining a constant rate, ensuring linear distance, and using effective decoding algorithms. 🙏🏼 My focus should remain on these critical aspects. I hope this framework feels both exciting and promising! :) I welcome any criticism, as it helps me catch potential oversights.

I would greatly appreciate your help and guidance. I will keep the PR in the draft stage in the mean time so that I incorporate your valuable feedback. Thank you!

- [ ] We recently started enforcing formatting checks. If formatting issues are reported in the new code you have written, please correct them


Krastanov:
Hi, Feroz! This looks interesting, good job on setting it up. Here are a few comments in no particular order:

- It would be important to do a bit of a literature review to see whether similar ideas have been explored (probably under different names); This exact idea might or might not be completely new, but there would certainly be similar constructions, and informing the reader about these would be valuable (and it will also make documenting and naming this construction easier for you, making it also easier for search engines to find it when someone uses a different name for it)
- The mapping `parity_checks(StructDefiningTheCode(parameters))` should be deterministic. See how in the rest of the library structures are used to fully define a code (and whenever there is some randomness to the construction we just use a helper). I do not know whether a complete new structure is needed here (maybe one that has enough information to define the circulant matrices), but a helper function for the actual construction would be a more idiomatic API (that then just returns and instance of a css or hgp code, whichever is simpler).
- The table decoder is a very limited decoder. The various belief propagation decoders would probably have better performance and provide more informative benchmarks.

Fe-r-oz:
Thank you so much!  I'm truly honored that you find it interesting. I’ll definitely conduct a more thorough literature review—your support means A LOT :) 

My sincere apologies for the multiple emails today. Also, you might this [discussion post](https://github.com/QuantumSavory/QuantumClifford.jl/discussions/475) is very stupid of me so please just skip it. I had some problems with using double forward slashes today... which I am glad was corrected, all thanks to Tommy 🙏🏼 My apologies.

Thank you so much again for your guidance!

Fe-r-oz:
Thank you so much for your thoughtful and insightful feedback! I’m pleased to share that I’ve incorporated point 2 into the implementation. Specifically, I’ve ensured a deterministic construction for `StructuredQLDPC`, alongside random constructors through `random_structured_qldpc_code`. To validate both approaches, I’ve added tests for the deterministic and random versions of the code. :)

It will likely take me a few more weeks to thoroughly review all the relevant literature, but I’m genuinely excited and motivated to dive deeper into this area—thanks so much in large part to your wonderful feedback! :)

Fe-r-oz:
Thank you so much for sharing your insights! That is much appreciated!

Regarding the term "Structured," it can be quite broad and may carry different meanings across various fields, such as classical wireless communication systems or coding theory. To avoid any ambiguity, I think using a more specific name like "Quantum Tillich-Zémor codes" would be a better choice. The previous name was just too long, broad and general...Thank you for the help!

Fe-r-oz:
Please help review this PR, Thank you! 

Thank you for your wonderful feedback and suggestions. They really helped in improving the constructors and help write better code. I hope that the constructors are now defined now as per your requirements and suggestions.

Based on my limited literature review, this exact idea seems completely new. However, in the literature, there are quantum LDPC codes by Tillich-Zémor which are described in https://arxiv.org/pdf/0903.0566. These quantum ldpc codes are pretty interesting as they are based on tanner graph product so we can use `Graphs.jl` to build them.  I have constructed them as well (currently and documenting testing them), maybe useful for us to determine how these codes compare with other quantum LDPC codes. There are similar ideas of how classical seed codes can be used to form new quantum LDPC codes such as the recent La-cross codes https://arxiv.org/pdf/2404.13010. Classical TZ codes have linear encoding rate and sub-linear distance which can be improved, so currently going through that literature. I will continue with more literature review to identify similar ideas from both the same and different authors. Thank you!

lib/QECCore/Project.toml
```diff
@@ -8,12 +8,14 @@ Combinatorics = "861a8166-3701-5b0c-9a16-15d98fcdc6aa"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 Graphs = "86223c79-3864-5bf0-83f7-82e725a168b6"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Nemo = "2edaba10-b0f1-5616-af89-8c11ac63239a"
```
nzy1997: Since `Nemo` is a relatively heavy dependency, we will include it in an extension of QECCore. You can set it up now or integrate this code into `QuantumClifford.j`l at this stage.

Fe-r-oz: I will set up a Ext that provides this accurate rank computation. The reason for using `Nemo` is for the accurate calculation for computed rank. The implementation for these new codes  is pure linear algebra, but I found out that using `Nemo` for rank calculating provides accurate results while deriving the formula for `code_k`. 

```
 H_gf2 = matrix(GF(2), H) # Nemo mat
 Ht_gf2 = transpose(H_gf2) # transpose called on Nemo mat
 k = c.n - rank(H_gf2) # rank called on Nemo mat
 kT = c.m - rank(Ht_gf2) # rank called on Nemo mat
```

Fe-r-oz: Done!

lib/QECCore/ext/QECCoreNemoExt.jl
```diff
@@ -0,0 +1,23 @@
...
+
+using QECCore
+using DocStringExtensions
+
+import Nemo
+import Nemo: GF, matrix, rank, transpose
+
+import QECCore: TillichZemor, code_k, _create_circulant_matrix, _create_matrix_M_deterministic
+
+function QECCore.code_k(c::TillichZemor)
```
nzy1997: Thank you for the changes! I think we can make this function work for all codes, like `distance` in `QuantumCliffordJuMPExt`. Would replacing `H = hcat(C, M)` with `H = party_matrix(c)` be sufficient?

nzy1997:
Everything looks fine—just a quick question: Why add the quantum Tillich-Zémor code to QECCore, while `random_TillichZemor_code` is in QuantumClifford?

lib/QECCore/ext/QECCoreNemoExt.jl
Fe-r-oz: I don’t think this substitution would work in general because this is a particular kind of classical parity check matrix  is of the form $H = [C \\mid M]$. So, the formula is specific to these codes that have  "structured" classical parity check matrix because `M` is a matrix that obeys certain existence conditions  specific to these codes and `C` is the circulant matrix. 

Note that the `H = party_matrix(c)` in the quantum cases would provide the boolean version of the stabilizer matrix, but here we use the classical parity check matrix  of the seed code. Also, since we are using the same classical structured parity check matrices as seeds `hgp(H, H)`, there are codes where `H1` is not the same as `H2` so $k^2_1$ does not apply, it's rather $k_1k_2 + k^T_1 k^T_2$

Indeed, this reminds me we can define the general `code_k`, `code_n` for quantum LDPC codes that formed `hgp` product formulation only but that is left for future work as it's unrelated to this PR.

Fe-r-oz:
Thank you for the feedback!

### PR Summary

This PR implements novel $[[n² + m², (n - rank([C ∣ M]))² + (m − rank([C ∣ M]ᵀ))², d]]$ quantum Tillich-Zémor qLDPC code family where  $H = [ C ∣ M]$ is the parity-check matrix of the underlying $(n, m, n)$ -Structured classical LDPC code. To the best of my knowledge, I didn't find this exact idea in the existing literature.

### Future work
- benchmark these `[[n² + m², (n - rank([C ∣ M]))² + (m − rank([C ∣ M]ᵀ))², d]]` against other codes.
- implement generalized versions of `code_n`, `code_k`,` distance` for qLDPC codes formed using hypergraph product construction.

#### [Reference](https://arxiv.org/pdf/2502.20189) for generalizing `code_n, code_k` for qLDPCs formed via hgp construction

> Let $C_i = [n_i, k_i, d_i]$ be a classical code encoding $k_i$ logical bits in $n_i$ physical bits with Hamming distance $d_i$. Let $r_i$ be the number of its checks and $H_i$ $\in$ $F_2^{r_i \times n_i}$ its parity-check matrix, i.e., the matrix having entries $H_i)_{ab} = 1$ iff the $a$-th check acts non-trivially on the $b$-th bit, so that $k_i = n_i - \text{rank}(H_i)$. Let $C^T_i$ be the transposed code of $C_i$ with parity-check matrix $H^T_i \in F_2^{n_i \times r_i}$. We denote $C^T_i \equiv [n^T_i, k^T_i, d^T_i]$ with obvious meaning of the code parameters. The hypergraph product (HGP)  takes the parity-check matrices  $H_i \in F_2^{r_i \times n_i}$  of two classical codes $C_i$ together with their transposed codes $C^T_i$,  $i = 1, 2$, and yields a $[[N, K, D]]$ quantum stabilizer code  with quantum parity-check matrix

> ```math
> H_Q = \begin{pmatrix}
> 0 & 0 & H_1 \otimes I_{n_2} & I_{r_1} \otimes H_2^T \\
> I_{n_1} \otimes H_2 & H_1^T \otimes I_{r_2} & 0 & 0 
> \end{pmatrix}
> ```

>where now $(H_Q)_αβ = 1$ iff the $α$th stabilizer acts nontrivially on the $β$th qubit. Quantum parameters read $N = n_1n_2 + r_1r_2$, $K = k_1k_2 + k^T_1 k^T_2$, $D \geq \min\{d_1, d_2, d^T_1, d^T_2\}$. ~ quoted from page 3 of reference.

**cc:** @Krastanov

lib/QECCore/ext/QECCoreNemoExt.jl
Fe-r-oz: I have defined the general version for `code_k` $k = n - rank(H_x) - rank(H_z)$ and moved the theoretical derivation to test file. This formulation is more general than `code_k` defined for quantum LDPC codes via `hgp` construction, in which case, we have  $K = k_1k_2 + k^T_1 k^T_2$.

nzy1997 (commented):
Appreciate your edits. The code appears more organized now. I don't have any other concerns.

Fe-r-oz:
Thanks for your suggestions!

PR #473, opened 2025-02-05
fix: `RGB` color initialization in `QuantumCliffordMakieExt` colormap

Fe-r-oz:
This PR aims to fix the `ArgumentError` when using functions from `QuantumCliffordMakieExt` module that now occurs on CI as CI Documentation error. It causes all the `@example` blocks to fail that calls the `stabilizerplot_axis`. The following stack trace on CI is the following for all example blocks that throw error:

```julia
Error: failed to run `@example` block in src/canonicalization.md:16-21
@example
using QuantumClifford, CairoMakie
f=Figure()
stabilizerplot_axis(f[1,1], canonicalize!(random_stabilizer(20,30)))
f

  exception = 
    ArgumentError: component type FixedPointNumbers.N0f8 is an 8-bit type representing 256 values from 0.0 to 1.0,
      but the values (0x001b9e77, 0x001b9e77, 0x001b9e77) do not lie within this range.
      See the READMEs for FixedPointNumbers and ColorTypes for more information.
    Stacktrace:
      [1] throw_colorerror_(::Type{FixedPointNumbers.N0f8}, values::Tuple{UInt32, UInt32, UInt32})
        @ ColorTypes ~/.julia/packages/ColorTypes/0RO5O/src/types.jl:692
      [2] throw_colorerror(::Type{ColorTypes.RGB{FixedPointNumbers.N0f8}}, values::Tuple{UInt32, UInt32, UInt32})
        @ ColorTypes ~/.julia/packages/ColorTypes/0RO5O/src/types.jl:700
      [3] checkval
        @ ~/.julia/packages/ColorTypes/0RO5O/src/types.jl:670 [inlined]
      [4] _new_colorant
        @ ~/.julia/packages/ColorTypes/0RO5O/src/types.jl:582 [inlined]
      [5] Color
        @ ~/.julia/packages/ColorTypes/0RO5O/src/types.jl:537 [inlined]
      [6] _new_colorant
        @ ~/.julia/packages/ColorTypes/0RO5O/src/types.jl:554 [inlined]
      [7] ColorTypes.RGB(g::UInt32)
        @ ColorTypes ~/.julia/packages/ColorTypes/0RO5O/src/types.jl:547
      [8] (::QuantumCliffordMakieExt.var"#3#4")(scene::Makie.Scene)
        @ QuantumCliffordMakieExt ~/work/QuantumClifford.jl/QuantumClifford.jl/ext/QuantumCliffordMakieExt/QuantumCliffordMakieExt.jl:18
[...]

PR #478, opened 2025-02-14
resolve #474

### LINKED ISSUE #474: fastrow and fastcolumns need to be documented, benchmarked (as part of docs and tests), and the defaults for memory layout need to be stated clearly
Below I give more context and possible places where to add the docs, benchmarks, and tests. But first, a quick list of what can be done:

- add discussion of fastrows and fastcolumns to the datastructure choice page (or split that page)
- add benchmarks to the test that verify that the default we have chosen are the fastest (for canonicalization with/without phase, for application of small named gates, for application of dense clifford operators)
- have fastrows and fastcolumns work on clifford operators as well (and add that to the benchmark matrix above)
- have the benchmarks auto-executed as part of the doc building and have them reported in the docs (in addition to the tests)

background:

First a quick point on storage. The very low-level storage format for a tableau with n qubit and r rows is `Vector{UInt8}` of length r for the phases and a `Matrix{UIntX}` (usually UInt64 as that is fastest for a typical CPU) of size `2*(n÷64)` by `r`. The second dimension of that matrix (r, its columns) enumerates as the first dimension of the tableau (its rows). The first dimension of that matrix enumerates the n x-bits, and then n z-bit of a given tableau row (i.e. it enumerates the columns of the tableau). This weird inversion is to ensure cache locality in the inner-most loop of canonicalization or gate application operations (i.e. in memory we have all the entries of a given tableau column as closely together, leading to ). It is the fast orientation for canonicalization and multiplication with other dense tableaux and it is the the slow orientation for application of small sparse gates as seen below. There are also the functions `fastcolumn` and `fastrow` that change these orientations internally without changing the interface the rest of the library sees, if one needs another type of default. It happens by using `Transpose{Matrix}` instead of `Matrix`, where the `Transpose` wrapper simply swaps the axes in the `Base.getindex` and `Base.setindex!` functions.

```julia-repl
julia> @time canonicalize!(sr);
  0.042747 seconds (1 allocation: 32 bytes)

julia> @time canonicalize!(sc); # canonicalization is faster with fastrow (the default)
  0.054066 seconds (1 allocation: 32 bytes)

julia> @time apply!(sr, sCNOT(1,2));
  0.000046 seconds (1 allocation: 32 bytes)

julia> @time apply!(sc, sCNOT(1,2)); # applying named gates is faster with fastcow (not default)
  0.000024 seconds (1 allocation: 32 bytes)
```

and for dense gates (below) they seem mostly the same, but this might also be because I have not set up a fastrow and fastcolumn for CliffordOperator

```julia-repl
julia> s = random_stabilizer(1000);

julia> c = random_clifford(1000);

julia> sr = fastrow(copy(s));

julia> sc = fastcolumn(copy(s));

julia> @time apply!(sr, c);
  0.052028 seconds (9 allocations: 800 bytes)

julia> @time apply!(sc, c);
  0.053642 seconds (9 allocations: 800 bytes)
```

And then there is also the question of whether we are accounting for phases (are you computing those?); Then the correct benchmark will be with `phases=false`.

- some discussion of this is at the bottom of this page: https://qc.quantumsavory.org/stable/datastructures/
- and some correctness tests for both memory layouts are here https://github.com/QuantumSavory/QuantumClifford.jl/blob/master/test/test_bitpack.jl

alilavas:
I have added a preliminary discussion of fastrow and fastcolumn to the data structure choice page, with hard-coded naive benchmarks. I will update it after I write proper benchmarks to incorporate auto-generated benchmarks. 

--------


alilavas:
Many GitHub automated tests have failed, but it appears they are not related to what I have changed. Should I worry about them or just ignore them?

Krastanov:
The `@allocations` errors are due to an issue with Julia 1.11 that appeared a week or so ago. Please mark it with `@test_broken` (and probably an if-else for the julia version as this works ok on 1.10).

The documentation failure is related to a change in Makie that I did not catch when changing compat bounds, also in the last week or two. I think it is fixed here: https://github.com/QuantumSavory/QuantumClifford.jl/pull/473 -- so do not worry about it, that fix will be merged separately.

Thank you for setting this up! I will need a couple of days to get back to it to provide comments (a few other urgent things happening).

Krastanov:
closing and reopening to rerun the tests now that the breaking from 1.11.3 is fixed on master

Krastanov:
Looks like a good start, thank you for engaging with this

Besides the documentation and benchmarks, the last thing would be implementing these functions for CliffordOperator (should be just a few lines, just calling into the implementation for Tableau)

PR #481, opened 2025-02-20
Basic GPU implementation of phaseless apply! of dense clifford operators

Eciccarelli0016:
This pull request is a little different in nature. I worked on my honors thesis at UMass Amherst relating to QuantumClifford.jl. All my work is contained in the one file I am adding, gpu_apply.ipynb. I don't expect it to be incorporated as a primary function for the library, but it is a nice start to attempting implementing GPU programming to this library with the hopes of a speed increase, especially for dense Clifford operators. This file contains 2 main function apply_gate_gpu! and apply_gate_gpu_64plus!. The second function works for any number of qubits while the first one, apply_gate_gpu! only works for 1-64 qubits. There is also benchmarking code, graphs, and other testing code in the notebook to really document everything I did, but the core of the project is really those two functions which are meant to mimic the current apply! function for PHASELESS operations. Also I did not add this to CHANGELOG, so upon review if that could be done (if necessary for this slightly different in nature pull request) that would be appreciated.


Krastanov:
[Final Honors Project Submission.pdf](https://github.com/user-attachments/files/19036696/Final.Honors.Project.Submission.pdf)

attached is the thesis that details this work for future reference; Thanks Evan!

PR #485, opened 2025-03-07
add gpu canonicalize! support

riteshbhirud:
This PR is for the gpu canonicalize! support

### DIFF
--- a/ext/gpu_canon.jl
+++ b/ext/gpu_canon.jl
@@ -1,0 +1,412 @@
+using CUDA
+using BenchmarkTools
+using LinearAlgebra
+using QuantumClifford  
+
+const MIN_BLOCKS = 36
+
+###########################################################################
+# Custom Linear Copy Kernel
+###########################################################################
+# This kernel copies elements from a source device array (src) to a destination device array (dest).
+function linear_copy_kernel!(dest, src, N)
+    i = (blockIdx().x - 1) * blockDim().x + threadIdx().x
+    stride = blockDim().x * gridDim().x
+    while i <= N
+        dest[i] = src[i]
+        i += stride
+    end
+    return
+end
+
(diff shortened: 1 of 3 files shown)


riteshbhirud:
added a test file

Krastanov (changes requested):
Apologies for the slow response. This is wonderful work!

A few quick points before a real in-depth review.

- For tests, comment out all of the `println` statements. The test runner has its own status printer.
- There should be no `include` statements in the test runner. This should be just a normal use of the library
- If you make new files in `src`, they probably should be added in an `include("./filename")` statement inside of `src/QuantumClifford.jl`
- The implementation should be in the CUDA extension inside of the corresponding subfolder in the `ext` folder, not in the main `src`. The library as a whole is supposed to NOT depend on CUDA, so that it can be installed without downloading many GBs of CUDA runtime libraries. But CUDA-specific capabilities can be added and conditionally loaded with a package extension (see https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions) )
- I am not sure what is the purpose of the file currently in the `ext` folder. It seems like it is a list of examples? That might make more sense in documentation or tests.
- the `if abspath...` things should not be part of the src or ext implementations, they should be in documentation (if they are documentation) or tests
- `src` probably would have very little added to it, I expect the majority of the code would actually be in the extension
- we also need to make these things work automatically on the appropriately typed stabilizer structures -- the GPU should be automatically used if the stabilizer tableau is stored in GPU memory, it should not be necessary to use functions with a `gpu_` prefix in their name. On the other hand, it is fine to have such internal functions if that helps with organizing the code.
- The tests themselves should probably have some naive simple benchmark, just to make sure that the GPU code is faster than some fixed constant (or than the CPU code)  -- more as a sanity check, less as a benchmark.

PR #501, opened 2025-06-05
apply_inv!

sagnikpal2004:
## Checklist

## Functionality
- inv_qubit_kernel: Returns a tuple for the inverted qubit kernel for all AbstractSingleQubitOperator (sHadamard, sPhase, etc)
- apply_inv!: Returns stabilizer after applying the inverted version of a certain operation.
```
apply_inv!(S"X", sHadamard(1)) # returns S"Z"
apply_inv!(S"Y", sPhase(1)) # returns S"X"
```

## TODO
- Fix code duplication (_apply! and _apply_inv! have a lot of same code)
- `_apply_inv!(stab, op::SingleQubitOperator)` in `symbolic_cliffords.jl`

## Notes
- Used python's `sympy.logic` to symbolically solve for inverse Boolean expressions and manually optimized the result by collapsing XOR chains

## Fixes
- Updated docstring for apply! and apply_inv!
- Changed parameter type from generic to fixed for _apply! and _apply_inv!

## Changelog
- added `apply_inv!` to apply the inverse of any quantum operation to a stabilizer state
- added `inv_qubit_kernel` to return the kernel for applying the inverse of a symbolic operation
- directly typed the parameter in `_apply!` and `_apply_inv!` instead of using a type parameter


test/test_inv.jl
```diff
@@ -0,0 +1,325 @@
+@testitem "Apply Inv" begin
+    @testset "Apply Inv" begin
+       
+        @test apply_inv!(S"X", sHadamard(1)) == apply!(S"X", inv(CliffordOperator(sHadamard(1), 1)))
```
Krastanov: I would suggest automating these tests a bit. You can just loop over all objects in `subtypes(appropriate_abstract_type)`. That way the tests also do not need to be updated when a new gate is added.

src/QuantumClifford.jl
```diff
@@ -1102,6 +1102,23 @@ function _apply!(stab::AbstractStabilizer, p::PauliOperator, indices; phases::Va
     stab
 end
 
+
+function apply_inv! end
```
Krastanov, line 1129: short docstring because you have made this public

src/QuantumClifford.jl
```diff
@@ -1102,6 +1102,23 @@ function _apply!(stab::AbstractStabilizer, p::PauliOperator, indices; phases::Va
...
+
+function apply_inv!(stab::AbstractStabilizer, op::AbstractCliffordOperator; phases::Bool=true)
+    @valbooldispatch _apply_inv!(stab,op; phases=Val(phases)) phases
+end
+function apply_inv!(stab::AbstractStabilizer, op::AbstractCliffordOperator, indices; phases::Bool=true)
+    @valbooldispatch _apply_inv!(stab,op,indices; phases=Val(phases)) phases
+end
+
+function _apply_inv!(stab::AbstractStabilizer, p::PauliOperator; phases::Val{B}=Val(true)) where B
+    # TODO
```
Krastanov: you can actually just use apply here because Pauli * Pauli^dagger = I

src/symbolic_cliffords.jl
```diff
@@ -71,8 +71,22 @@ function _apply!(stab::AbstractStabilizer, gate::G; phases::Val{B}=Val(true)) wh
     stab
 end
 
+function _apply_inv!(stab::AbstractStabilizer, gate::G; phases::Val{B}=Val(true)) where {B, G<:AbstractSingleQubitOperator}
```
Krastanov: discussed in the call: this `G` is not necessary but you mentioned it is set up like that in some other place as well. Could you leave a link for that other place. It is probably a silly oversight there.

Krastanov:
could you edit the PR description and brag about the boolean tool you used to solve the inverses

Krastanov:
@sagnikpal2004 , could you address the merge conflict?

src/symbolic_cliffords.jl
sagnikpal2004: https://github.com/QuantumSavory/QuantumClifford.jl/blob/2d199a80003402f94fac63ca5a108e556b95a453/src/symbolic_cliffords.jl#L60
https://github.com/QuantumSavory/QuantumClifford.jl/blob/2d199a80003402f94fac63ca5a108e556b95a453/src/symbolic_cliffords.jl#L270

src/QuantumClifford.jl
```diff
@@ -1117,6 +1117,25 @@ function _apply!(stab::AbstractStabilizer, p::PauliOperator, indices; phases::Va
     stab
 end
 
+
+"""In `QuantumClifford` the `apply_inv!` function is used to apply the inverse of any quantum operation to 
+a stabilizer state, including unitary Clifford and Pauli operators"""
```
```suggestion
"""
    apply_inv!
    
Apply the inverse of any quantum operation to a stabilizer state.
"""
```

src/symbolic_cliffords.jl
Krastanov:
```suggestion
function _apply_inv!(stab::AbstractStabilizer, gate::AbstractSingleQubitOperator; phases::Val{B}=Val(true)) where {B} # code repetition with the corresponding `_apply`
```

Krastanov: yup, good point

src/symbolic_cliffords.jl
```diff
@@ -184,6 +202,9 @@ function _apply!(stab::AbstractStabilizer, op::SingleQubitOperator; phases::Val{
     end
     stab
 end
+function _apply_inv!(stab::AbstractStabilizer, op::SingleQubitOperator; phases::Val{B}=Val(true)) where B
+    # TODO
+end
```
Krastanov:
```suggestion
```

That way we will get an understandable method error

src/symbolic_cliffords.jl
```diff
@@ -292,8 +313,33 @@ function _apply!(stab::AbstractStabilizer, gate::G; phases::Val{B}=Val(true)) wh
     stab
 end
 
+function _apply_inv!(stab::AbstractStabilizer, gate::G; phases::Val{B}=Val(true)) where {B, G<:AbstractTwoQubitOperator}
```
Krastanov:
```suggestion
function _apply_inv!(stab::AbstractStabilizer, gate::AbstractTwoQubitOperator; phases::Val{B}=Val(true)) where {B} # code repetition with the corresponding `_apply`
```

PR #508, opened 2025-06-12
Move some simple codes into QECCore.jl

nzy1997:
Move the following codes into QECCore.jl:
`Toric`, `Surface`, `RepCode`, `CSS`, `Shor9`, `Steane7`, `Cleve8`, `Perfect5`, `Bitflip3`.


lib/QECCore/src/codes/surface.jl
```diff
@@ -14,33 +17,33 @@ The surface code has open boundary conditions, unlike the toric code. To this en
...
-+ ____ZZ_Z
+The parity checks of `Surface(3,2)` are:
+```
+X__X__X_
+_X__X_XX
+__X__X_X
+ZZ____Z_
+_ZZ____Z
+___ZZ_Z_
+____ZZ_
```
Krastanov: this looks like a typo, a missing Z

src/ecc/ECC.jl
```diff
@@ -49,6 +49,7 @@ Only CSS codes have this method.
 
 See also: [`parity_checks`](@ref)"""
 function parity_checks_x(code::AbstractECC)
+    return parity_matrix_x(code)
     throw(lazy"Codes of type $(typeof(code)) do not have separate X and Z parity checks, either because they are not a CSS code and thus inherently do not have separate checks, or because its separate checks are not yet implemented in this library.")
```
Krastanov: I do not think I understand this change. The throw will never be reached.

src/ecc/ECC.jl
```diff
@@ -58,6 +59,7 @@ Only CSS codes have this method.
 
 See also: [`parity_checks`](@ref)"""
 function parity_checks_z(code::AbstractECC)
+    return parity_matrix_z(code)
     throw(lazy"Codes of type $(typeof(code)) do not have separate X and Z parity checks, either because they are not a CSS code and thus inherently do not have separate checks, or because its separate checks are not yet implemented in this library.")
```
Krastanov: same here, this change seems it will just cause infinite recursion and the throw will never be reached

src/ecc/ECC.jl
```diff
@@ -84,6 +86,7 @@ function generator_polynomial end
 function generator end
 
 parity_checks(s::Stabilizer) = s
+parity_checks(c::AbstractECC) = Stabilizer(parity_matrix(c))
```
Krastanov, line 87: thanks, this removes a lot of worries about backward compat

src/ecc/ECC.jl
```diff
@@ -110,11 +113,11 @@ end
...
-"""Parity matrix of a code, given as a stabilizer tableau."""
-function parity_matrix(c::AbstractECC)
-    paritym = stab_to_gf2(parity_checks(c::AbstractECC))
-    return paritym
-end
+# """Parity matrix of a code, given as a stabilizer tableau."""
+# function parity_matrix(c::AbstractECC)
+#     paritym = stab_to_gf2(parity_checks(c::AbstractECC))
+#     return paritym
+# end
```
Krastanov: any reason not to just delete this?

test/test_aqua.jl
```diff
@@ -2,3 +2,5 @@
     using Aqua
     Aqua.test_all(QuantumClifford)
 end
+root_project_path, found = Aqua.root_project_toml(Base.PkgId(Base.UUID("0525e862-1e90-11e9-3e4d-1b39d7109de1"), "QuantumClifford"))
+Aqua.precompile_wrapper(root_project_path,10,quote end)
```
Krastanov: could you document why this is necessary?

lib/QECCore/src/QECCore.jl
```diff
@@ -8,6 +8,8 @@ using Combinatorics
 export distance, parity_matrix, code_n, code_s, code_k, parity_matrix_x, parity_matrix_z, rate
 export AbstractECC, AbstractQECC, AbstractCECC, AbstractCSSCode
 
+export code_s_temp
```
Krastanov: could you document why this is needed and what it means

lib/QECCore/src/codes/surface.jl
Krastanov: I was reviewing commit by commit -- I see this has been fixed in latter commits, marking comment as resolved

Krastanov:
there are a few comments on outdated pieces of code because I was revieweing commit by commit. Feel free to disregard them.

Krastanov:
@nzy1997 , could you check that the tests are actually running with the setup you have contributed? Looking through most of the actions, they seem to be finishing in 2 minutes, instead of the usual 30. I suspect TestItems is not working with the project re-activation you are using.

Issue #446, opened 2024-12-07
Generalizing `LPCode` to Non-Commutative and ℓ-Dimensional Associative Algebras


Fe-r-oz:
**Is your feature request related to a problem? Please describe.**

The paper [Asymptotically Good Quantum and Locally Testable Classical LDPC Codes](https://arxiv.org/pdf/2111.03654) developed the LPCode construction for non-abelian groups, where the $R$ is not necessarily commutative.  The current limitation of #356 is that it works only  with "commutative" group algebra R. When R is commutative, "we do not need to distinguish the left and the right representations". What that means that for `A`  and `B`  `GroupAlgebraElemMatrix`, we don't specify any particular group action as we see the default left representation is used for both `A` and `B.`  In the following line, repr is `Hecke.representation_matrix`

https://github.com/QuantumSavory/QuantumClifford.jl/blob/5fcf3531737f54ff8797e27d5cb25b45fb85a9a7/ext/QuantumCliffordHeckeExt/lifted.jl#L93

Hence, the LPCode currently works as Abelian LPCode. To generalize it, Panteleev and Kalachev remarked that for matrices $A$ and $B$, we use the **right** regular representation for $A$ and the **left** regular representation for $B$ to form the block matrices $\hat{A}$ and $\hat{B}$, defining the CSS code $LP(A, B)$.  This approach also works well with any $\ell$-dimensional associative algebra $R$ over $\mathbb{F}_q$, not necessarily commutative, if we use the right (resp. left) regular matrix representation of its elements as the entries of $\hat{A}$ (resp. $\hat{B}$). When $R$ is not commutative, it becomes necessary to distinguish between the right and left regular representations to ensure the correct construction of the block matrices. 

This construction method is presented in foot note 4, page 3 and detailed in appendix B  (Page 48) of the [aforementioned paper](https://arxiv.org/pdf/2111.03654)

Related Issue: https://github.com/QuantumSavory/QuantumClifford.jl/issues/405

**Describe the solution you’d like**

- [ ] For non commutative algebra (non-abelian groups), use the right (resp. left) regular matrix representation of its elements as the entries of $\hat{A}$ (resp. $\hat{B}$). 
- [ ] Resolve #405 
[...]

PR #402, opened 2024-10-23
tests:  non-abelian dihedral group based `LPCode` via `semidirect_product` Dₘ= Cₘ⋉C₂ of cyclic groups

Fe-r-oz:
This PR reproduces the Table 3 of Lin and Pryadko's paper by using Semidirect product of two cyclic groups instead of group presentation to create non-abelian Dihedral group which was done in: https://github.com/QuantumSavory/QuantumClifford.jl/pull/397

As $D_{m} = C_m \ltimes C_2 = \langle x, s \mid x^m = s^2 = (r*s)^2 = 1 \rangle$, there are two ways to reproduce Table 3:

1) $C_m \ltimes C_2$ (This PR)
2) $\langle x, s \mid x^m = s^2 = (r*s)^2 = 1 \rangle$ (done in #397)

This PR implements the 1) approach. This first approach was not mentioned in the caption of Table 3 but was referenced in the text of the Appendix. Both gives the same results.

Related PR:  #397, #511

- [x] The code is properly formatted and commented.
- [x] Substantial new functionality is documented within the docs.
- [x] All new functionality is tested.
- [x] All of the automated tests on github pass.
- [x] We recently started enforcing formatting checks.

PR #511, opened 2025-06-20
support non-abelian `LPCode` by generalizing to non-commutative group algebras

Fe-r-oz:
This PR aims to fix stabilizer non-commutativity in `LPCode` for non-abelian groups by introducing two representations functions: `A_repr` (the right regular representation) and `B_repr` (a representation derived from [arXiv:2111.03654, Appendix B](https://arxiv.org/pdf/2111.03654)). Previously, we used left regular representation (which is default in `Hecke`) which works for commutative algebras because left regular representation is same as right regular representation.

The PR implements @royess insightful suggestions in this [discussion](https://github.com/QuantumSavory/QuantumClifford.jl/issues/446#issuecomment-2745130382). As suggested,  I attempted using `hx, hz = concat_lift_repr(c.A_repr, hx), concat_lift_repr(c.B_repr, hz)` but this preserved non-commutativity errors, so adopted the Appendix B approach with explicit handling of CSS orthogonality using `A_repr`, and `B_repr`.

@royess, thank you for your guidance on this! Whenever you have time, please let me know your feedback. We can submit for review once you approve the approach.

The tests that were throwing non-commutativity error when using non-abelian groups are fixed.
```
julia> using QuantumClifford: check_allrowscommute, stab_looks_good; using QuantumClifford.ECC: two_block_group_algebra_codes, code_n, code_k, parity_checks;
julia> import Hecke: gens, quo, group_algebra, GF, one;
julia> import Oscar: free_group, small_group_identification, describe, order, dihedral_group;
julia> m = 6;
julia> G = dihedral_group(2*m);
julia> GA = group_algebra(GF(2), G);
julia> s, r = gens(G);
julia> r^m == s^2 == (r*s)^2
true
julia> s, r = gens(GA);
julia> a_elts = [one(G), r^4];
julia> b_elts = [one(G), s*r^4, r^3, r^4, s*r^2, r];
julia> a = sum(GA(z) for z in a_elts);
julia> b = sum(GA(z) for z in b_elts);
julia> c = two_block_group_algebra_codes(a,b);
julia> check_allrowscommute(parity_checks(c))
true
julia> stab_looks_good(parity_checks(c), remove_redundant_rows = true) 
true
julia> order(G)
12
julia> describe(G)
"D12"
julia> code_n(c), code_k(c)
(24, 8)
```

Semdirect product based group algebras also pass:
```
julia> m = 6;
julia> Cₘ = cyclic_group(m);
julia> C₂ = cyclic_group(2);
julia> A = automorphism_group(Cₘ);
julia> au = A(hom(Cₘ,Cₘ,[Cₘ[1]],[Cₘ[1]^-1]));
julia> f = hom(C₂,A,[C₂[1]],[au]);
julia> G = semidirect_product(Cₘ,f,C₂);
julia> GA = group_algebra(GF(2), G);
julia> s, r = gens(GA);
julia> r^m == s^2 == (s*r)^2;
julia> a_elts = [one(r), r^4];
julia> b_elts = [one(r), s*r^4, r^3, r^4, s*r^2, r];
julia> a = sum(GA(x) for x in a_elts);
julia> b = sum(GA(x) for x in b_elts);
julia> c = two_block_group_algebra_codes(a,b);
julia> check_allrowscommute(parity_checks(c))
true
julia> stab_looks_good(parity_checks(c), remove_redundant_rows = true) 
true
julia> order(G)
12
julia> describe(G)
"D12"
julia> code_n(c), code_k(c)
(24, 8)
```

⚠️ B̶u̶t̶ I̶ g̶e̶t̶ a̶ d̶i̶s̶c̶r̶e̶p̶e̶n̶c̶y̶ f̶o̶r̶ c̶o̶d̶e̶_̶k̶ v̶a̶l̶u̶e̶ o̶f̶ c̶o̶d̶e̶s̶ f̶r̶o̶m̶ h̶t̶t̶p̶s̶:̶//a̶r̶x̶i̶v̶.o̶r̶g̶/p̶d̶f̶/2̶1̶1̶1̶.0̶7̶0̶2̶9̶. F̶o̶r̶ a̶l̶l̶ o̶t̶h̶e̶r̶ t̶y̶p̶e̶s̶ o̶f̶ L̶P̶C̶o̶d̶e̶s̶, t̶h̶e̶ n̶, k̶ v̶a̶l̶u̶e̶s̶ a̶r̶e̶ c̶o̶n̶s̶i̶s̶t̶e̶n̶t̶. A̶f̶t̶e̶r̶ i̶n̶s̶p̶e̶c̶t̶i̶o̶n̶, I̶t̶ a̶p̶p̶e̶a̶r̶s̶ t̶h̶a̶t̶ n̶o̶w̶ t̶h̶i̶s̶ i̶s̶ n̶o̶n̶-̶c̶o̶m̶m̶u̶t̶a̶t̶i̶v̶e̶ n̶o̶w̶ w̶h̶i̶c̶h̶ i̶s̶ a̶ b̶u̶g̶ 🐛 

Commutative cases work as usual as it uses the approach used previously: 
```
julia> import Hecke; using QuantumClifford.ECC;

julia> base_matrix = [0 0 0 0; 0 1 2 5; 0 6 3 1]; l = 7;

julia> c2 = LPCode(base_matrix, l .- base_matrix', l);

julia> code_n(c2), code_k(c2)
(175, 19)
```

For all the other families of LPCodes, there appears to be no error:
```
julia> import Hecke: group_algebra, GF, abelian_group, gens; import LinearAlgebra: diagind; using QuantumClifford.ECC;

julia> l = 63; GA = group_algebra(GF(2), abelian_group(l)); x = gens(GA)[];
julia> A = zeros(GA, 7, 7);
julia> A[diagind(A)] .= x^27;
julia> A[diagind(A, -1)] .= x^54;
julia> A[diagind(A, 6)] .= x^54;
julia> A[diagind(A, -2)] .= GA(1);
julia> A[diagind(A, 5)] .= GA(1);
julia> B = reshape([1 + x + x^6], (1, 1));
julia> c1 = LPCode(A, B);
julia> stab_looks_good(parity_checks(c1))
false
julia> stab_looks_good(parity_checks(c1), remove_redundant_rows=true)
true
julia> code_n(c1), code_k(c1)
(882, 24)
```

Edit: I added the same links for the two papers in the post, so corrected this.

- [x] We recently started enforcing formatting checks.


ext/QuantumCliffordHeckeExt/lifted_product.jl
```diff
@@ -98,53 +151,76 @@ struct LPCode <: AbstractECC
...
     end
 
-    function LPCode(c₁::LiftedCode, c₂::LiftedCode; GA::GroupAlgebra=c₁.GA, repr::Function=c₁.repr)
-        # we are using the group algebra and the representation function of the first lifted code
+    function LPCode(c₁::LiftedCode, c₂::LiftedCode; GA::GroupAlgebra=c₁.GA, A_repr::Function=x->representation_matrix(x, :right), B_repr::Function=x->representation_matrix(x, :left))
         c₁.GA == GA && c₂.GA == GA || error("The base rings of both lifted codes must be the same as the group algebra")
-        new(c₁.A, c₂.A, GA, repr)
+        new(c₁.A, c₂.A, GA, A_repr, B_repr)
     end
 end
```
Krastanov, line 192: let's add a few constructors for backward compatibility:

```
function LPCode(A::GroupAlgebraElemMatrix, B::GroupAlgebraElemMatrix; GA::GroupAlgebra=parent(A[1,1]), repr::Function)
  if !check_algebras_are_comm
      throw(...)
  end
  return LPCode(A, B, GA, repr, repr)
end
```

Krastanov: and for the other constructor you modified

test/test_ecc_base.jl
```diff
@@ -47,6 +47,31 @@ test_hcubic_codes = [
     haah_cubic_codes([0, 15, 20, 28, 66], [0, 58, 59, 100, 121], 3)
 ]
 
+# Lifted product codes using non-commutative algebras
```
Krastanov: could you add a reference in the comment where these come from or if they are not from a paper could you say something like "these are just made up, they are not known to specifically be good, they are not from a paper"

src/nonclifford.jl
```diff
@@ -338,7 +338,7 @@ function projectrand!(sm::GeneralizedStabilizer, p::PauliOperator)
     # Apply the corresponding projection
     sm, _ = _projectrand_notnorm(sm, p, res)
     # Normalize the state
-    trace = tr(sm)
```
Krastanov: undo this if not needed anymore (maybe it requires merge with master)

ext/QuantumCliffordHeckeExt/lifted_product.jl
Fe-r-oz: Thank you! T̶h̶e̶r̶e̶ i̶s̶ n̶o̶ n̶e̶e̶d̶ f̶o̶r̶ b̶a̶c̶k̶w̶a̶r̶d̶ c̶o̶m̶p̶a̶t̶i̶b̶i̶l̶i̶t̶y̶ b̶e̶c̶a̶u̶s̶e̶ w̶e̶ a̶r̶e̶ b̶a̶c̶k̶ t̶o̶ u̶s̶i̶n̶g̶ `̶r̶e̶p̶r̶`̶ 😄   

Edit: Reverted back to two custom representations, `A_repr` and `B_repr`

Fe-r-oz:
@Krastanov, Thank you for your feedback! Please help review this PR, Thank you! I completely overhauled the previous approach and implemented a (cleaner?) solution. While this differs from the approach @royess kindly suggested earlier, I believe it’s a (more streamlined way ?) with no need for backward compatibility. Forgive the ping — I wanted to surface this alternative approach. Now, we use a single representation `repr` and control how we manage the use of right and left representations, rather than maintaining separate ones. This should provide a more maintainable and lasting solution. Apologies for the ping, 🛰️ 📡

ext/QuantumCliffordHeckeExt/lifted_product.jl
```diff
@@ -131,20 +185,43 @@ end
...
+    mb, nb = size(B)
+    # Apply regular representations to ensure CSS orthogonality for non-commutative algebras:
+    # Â = ρ(A) (right regular representation: x ↦ xa)
+    # B̂ = λ(B) (left regular representation: x ↦ bx)
+    Â = concat_lift_repr(x->representation_matrix(x, :right), A) # ρ(A) right regular repr
+    B̂ = concat_lift_repr(x->representation_matrix(x, :left), B) # λ(B) left regular repr
+    # Compute adjoints with representations:
+    # B̂* = λ(B*) and Â* = ρ(A*)
+    B̂_conj = concat_lift_repr(x->representation_matrix(x, :left), permutedims(group_algebra_conj.(B))) # λ(B*)
+    Â_conj = concat_lift_repr(x->representation_matrix(x, :right), permutedims(group_algebra_conj.(A))) # ρ(A*)
```
Krastanov: I think I disagree with this approach and preferred the previous version of this PR. Here is my reasoning:

The whole point of having the `repr` method is that some users might to use a group representation that is not the default `Hecke.representation_matrix`. So we might very well have a user who uses a different representation with a non-commutative group algebra, and this code with disregard the representation the user has specified.

Krastanov: A separate question: do you have an understanding of why this more general `_build_parity_matrix_xz(::NonCommutative)` can not be used also for the commutative case?

Fe-r-oz: >  So we might very well have a user who uses a different representation with a non-commutative group algebra, and this code with disregard the representation the user has specified.

Indeed, I preferred the previous version as well as that provided two separate representations that can be custom as well. In that case, the user needs to have two custom representations, `A_repr` and `B_repr`, instead of single custom representation for non-commutative algebra.

Fe-r-oz: > A separate question: do you have an understanding of why this more general _build_parity_matrix_xz(::NonCommutative) can not be used also for the commutative case?

Thank you for the great question! Please let me know your thoughts!

We maintain two separate implementations because the commutative and non-commutative cases rely on fundamentally different assumptions about how group algebra elements are stored, lifted, and represented in constructing parity-check matrices. 

While the non-commutative version `_build_parity_matrix_xz(::NonCommutative)`  is more general in its handling of raw group algebra elements, it cannot fully replace the commutative version due to how legacy construction (i.e. base-matrix LPCode construction)  handle preprocessing and implicit inverses.

In the commutative case, the parity-check matrices are constructed using a straightforward structure:

hx = [A ⊗ I | I ⊗ B]
hz = [I ⊗ B* | A* ⊗ I]

Here, conjugation is applied explicitly using `group_algebra_conj`, and the representation map (through `concat_lift_repr`) is applied after forming Kronecker products. This late-lifting approach works well because, in commutative algebras, representation maps distribute over tensor products. This construction does not involve double conjugation and yield correct results to how the `LPCode` supports commutative group algebra.

This approach also naturally supports the base-matrix convention used in constructs like `LPCode(base_matrix, l .- base_matrix', l)`. In this construction, the second matrix is implicitly treated as the inverse of the first by subtracting shifts modulo l. Because representation is applied only after these structural operations, the inversion trick remains compatible. This makes the commutative implementation ideal for supporting legacy and base-matrix LP code paths without modifying the internal data format.

By contrast, the non-commutative case imposes 'stricter' rules to preserve CSS orthogonality. It follows an early lifting approach, where the group algebra elements A and B are mapped via representations (e.g., ρ(A) and λ(B)) before forming the Kronecker products. The resulting parity-check matrices take the form:

hx = [ρ(A)⊗I | I⊗λ(B)]
hz = [I⊗λ(B*) | ρ(A*)⊗I]

This strategy ensures that orthogonality conditions are enforced at every step, but it relies on `A` and `B` being raw, unprocessed group algebra elements. If instead the matrices were constructed with preprocessing conventions—such as storing` B⁻¹` via `l .- base_matrix'`—then the assumptions are broken, and the generated stabilizers become invalid.

This issue was highlighted very first [commit](https://github.com/QuantumSavory/QuantumClifford.jl/pull/511/commits/004dcebfebba7af89dcc75721b2a0f11a023f02e) of this PR, where I attempted to use `_build_parity_matrix_xz(::NonCommutative)` for both commutative and non-commutative cases. While it worked for LP codes built from raw group algebra elements, it failed when applied to codes constructed using base matrices. The CSS orthogonality was violated, leading to incorrect stabilizers and erroneous code_k values.

There are currently two major pathways for constructing LP codes. The first is the group-algebra-based method, which is used in most of our examples and works seamlessly with the non-commutative implementation. 

https://github.com/QuantumSavory/QuantumClifford.jl/blob/def729fe59d8f9843ea31f7129045e21c47cf957/ext/QuantumCliffordHeckeExt/lifted.jl#L74

The second, used less frequently, is the base-matrix method which relies on implicit inverse encoding. This pathway is incompatible with the early-lifting strategy of the non-commutative approach because it violates the assumption that inputs are raw.

https://github.com/QuantumSavory/QuantumClifford.jl/blob/def729fe59d8f9843ea31f7129045e21c47cf957/ext/QuantumCliffordHeckeExt/lifted.jl#L87

I also experimented with a unified strategy based on this [issue comment](https://github.com/QuantumSavory/QuantumClifford.jl/issues/446#issuecomment-2745130382), where representations were applied in-place without altering hgp. Unfortunately, this still produced invalid stabilizers in the non-commutative group algebra setting. I will also debug the legacy code.

The commutative version succeeds because it  accommodates the legacy shift-inversion that base-matrix constructor of LPCode requires, deferring lifting until after the structural form is complete. The non-commutative version is general (works for both commutative and non-commutative group algebra) but strictly correct only when the inputs are unprocessed group algebra elements. 

Therefore, `_build_parity_matrix_xz(::NonCommutative)` could be be used for both cases, doing so would require a standardization  and the removal of all implicit preprocessing from legacy code.

Until such normalization is in place, it might be both safer to keep both implementations, each handling their respective cases. The commutative version supports inverse-encoding via late lifting, while the non-commutative version guarantees orthogonality for non-commutative group algebra through the use of left and right representation.


##

Additional context:

Some rambling:

The current code in repo:

Note that the double-conjugation pattern when the group algebra is commutative. This appears in:

https://github.com/QuantumSavory/QuantumClifford.jl/blob/def729fe59d8f9843ea31f7129045e21c47cf957/ext/QuantumCliffordHeckeExt/lifted_product.jl#L146

Second conjugation (inside hgp):

https://github.com/QuantumSavory/QuantumClifford.jl/blob/def729fe59d8f9843ea31f7129045e21c47cf957/ext/QuantumCliffordHeckeExt/lifted_product.jl#L140

For commutative algebras, the implementation achieves the effective behavior `[A⊗I | I⊗B]` for X-checks (through the double-conjugation cancellation where `conj(conj(B)) = B`) and `[I⊗B* | A*⊗I]` for Z-checks (via single conjugation).. This structure inherently supports the base-matrix LP code construction  `l.-base_matrix'` inversion through its late lifting approach (`concat_lift_repr` applied after Kronecker products). In contrast, the non-commutative version requires distinct representations (`[ρ(A)⊗I | I⊗λ(B)]` for X-checks and `[I⊗λ(B*)|ρ(A*)⊗I]` for Z-checks) to maintain orthogonality, necessitating early lifting before Kronecker products. 

> **Edit:** The permalinks were swapped so I unswapped them at one place so they refer to correct location. 

> Some code examples that highlight the aforementioned incompatible assumptions.
 
> I also experimented with a unified strategy based on this [issue comment](https://github.com/QuantumSavory/QuantumClifford.jl/issues/446#issuecomment-2745130382), where representations were applied in-place without altering hgp. Unfortunately, this still produced invalid stabilizers in the non-commutative group algebra setting.  I will also debug the legacy code.

After some debugging, `group_algebra_conj(group_algebra_conj(x)) == x` (̶o̶n̶l̶y̶ t̶r̶u̶e̶ f̶o̶r̶ c̶o̶m̶m̶u̶t̶a̶t̶i̶v̶e̶ a̶l̶g̶e̶b̶r̶a̶)̶, so looks like the culprit that [issue comment](https://github.com/QuantumSavory/QuantumClifford.jl/issues/446#issuecomment-2745130382) strategy didn't work because of the implicit assumptions with which the base-matrix `LPCode` construction is working for only commutative case.

The small groups used in this example form non-commutative group algebra. The `LPCode` base matrix construction assumes abelian-group under the hood, so it form commutative group algebra.

> Edit^3:  These examples after running on new terminal:
 
```
julia> using LinearAlgebra
julia> import Hecke; using QuantumClifford.ECC;
julia> using Hecke
julia> import Hecke; using QuantumClifford.ECC;
julia> base_matrix = [0 0 0 0; 0 1 2 5; 0 6 3 1]; l = 7;
julia> c = LPCode(base_matrix, l .- base_matrix', l);
julia> a_conj = permutedims(group_algebra_conj.(c.A));
julia> a_conj_conj = permutedims(group_algebra_conj.(a_conj));
julia> a_conj_conj == c.A
true
julia> b_conj = permutedims(group_algebra_conj.(c.B));
julia> b_conj_conj = permutedims(group_algebra_conj.(b_conj));
julia> b_conj_conj == c.B
true

julia> G = small_group(36,1);
julia> GA = group_algebra(GF(2), G);
julia> r, s  = gens(GA);
julia> A = 1 + r;
julia> B = 1 + s + r^6 + s^3*r + s*r^7 + s^3*r^5;
julia> c = two_block_group_algebra_codes(A,B);
julia> a_conj = permutedims(group_algebra_conj.(c.A));
julia> a_conj_conj = permutedims(group_algebra_conj.(a_conj));
julia> a_conj_conj == c.A
true
julia> b_conj = permutedims(group_algebra_conj.(c.B));
julia> b_conj_conj = permutedims(group_algebra_conj.(b_conj));
julia> b_conj_conj == c.B
true

julia> G = small_group(40,8);
julia> GA = group_algebra(GF(2), G);
julia> r, s  = gens(GA);
julia> A = 1 + s*r^5 + r^5 + s*r^6;
julia> B = 1 + s^2 + r + s^2*r^3;
julia> c = two_block_group_algebra_codes(A,B);
julia> a_conj = permutedims(group_algebra_conj.(c.A));
julia> a_conj_conj = permutedims(group_algebra_conj.(a_conj));
julia> a_conj_conj == c.A
true
julia> b_conj = permutedims(group_algebra_conj.(c.B));
julia> b_conj_conj = permutedims(group_algebra_conj.(b_conj));
julia> b_conj_conj == c.B
true
```

Why does the (`_build_parity_matrix_xz(::NonCommutative`)  works with `LPCode` pathway 1 and not with base-matrix `LPCode` construction pathway 2?

In the group element construction (`LiftedCode(group_elem_array))`, elements are stored raw and inverses are explicitly computed using `group_algebra_conj,` which works naturally with the generalized approach (`_build_parity_matrix_xz(::NonCommutative`) ). 

However, the base matrix construction (`LiftedCode(shift_array, l)`) uses an implicit inverse encoding scheme through the` l .- base_matrix'` convention, which assumes that double conjugation (index-based inversion via the multiplication table) will perfectly cancel out `(i.e., group_algebra_conj(group_algebra_conj(x)) == x.`  The non-commutative version more general so it does have these assumptions. T̶h̶i̶s̶ a̶s̶s̶u̶m̶p̶t̶i̶o̶n̶ o̶n̶l̶y̶ h̶o̶l̶d̶s̶ f̶o̶r̶ c̶o̶m̶m̶u̶t̶a̶t̶i̶v̶e̶ a̶l̶g̶e̶b̶r̶a̶s̶, a̶s̶ d̶e̶m̶o̶n̶s̶t̶r̶a̶t̶e̶d̶ b̶y̶ a̶f̶o̶r̶e̶m̶e̶n̶t̶i̶o̶n̶e̶d̶ e̶x̶a̶m̶p̶l̶e̶s̶ w̶h̶e̶r̶e̶ n̶o̶n̶-̶c̶o̶m̶m̶u̶t̶a̶t̶i̶v̶e̶ c̶a̶s̶e̶s̶ f̶a̶i̶l̶ t̶h̶i̶s̶ p̶r̶o̶p̶e̶r̶t̶y̶ (̶s̶h̶o̶w̶n̶ b̶y̶ `̶b̶_̶c̶o̶n̶j̶_̶c̶o̶n̶j̶ =̶=̶ b̶_̶c̶o̶n̶j̶`̶ r̶e̶t̶u̶r̶n̶i̶n̶g̶ f̶a̶l̶s̶e̶)̶. 

Hence, this makes the base-matrix `LPCode `construction fundamentally incompatible with the more general non-commutative version that requires explicit representation handling and is not relying on double conjugation cancellation. a̶n̶d̶ c̶a̶n̶n̶o̶t̶ r̶e̶l̶y̶ o̶n̶ d̶o̶u̶b̶l̶e̶ c̶o̶n̶j̶u̶g̶a̶t̶i̶o̶n̶ (̶i̶n̶d̶e̶x̶-̶b̶a̶s̶e̶d̶ i̶n̶v̶e̶r̶s̶i̶o̶n̶ v̶i̶a̶ t̶h̶e̶ m̶u̶l̶t̶i̶p̶l̶i̶c̶a̶t̶i̶o̶n̶ t̶a̶b̶l̶e̶)̶ c̶a̶n̶c̶e̶l̶l̶a̶t̶i̶o̶n̶.

The base matrix construction makes three critical commutative-specific assumptions: 
1. inverses can be encoded via modular arithmetic,
2. double conjugation (strictly speaking index-based inversion via the multiplication table) cancels w̶h̶i̶c̶h̶ i̶t̶ d̶o̶e̶s̶ n̶o̶t̶ f̶o̶r̶ n̶o̶n̶-̶c̶o̶m̶m̶u̶t̶a̶t̶i̶v̶e̶ c̶a̶s̶e̶,  but non-commutative version does not rely on double conjugation.
3. representations can be applied after Kronecker products. 

These incompatibility implicit assumptions is why we have two separate implementations for the two cases: commutative case that handle both the pathways (group algebra elements, and base matrix construction) to construct LPcodes and the non-commutative case for non-commutative group algebra.

ext/QuantumCliffordHeckeExt/lifted_product.jl
Fe-r-oz: Thank you! I initially implemented aforementioned two functions with  `repr` parameter, but they caused method overwriting errors due to identical signatures with the new `A_repr/B_repr` representation functions.

Fe-r-oz: Added `repr` for backwards compatibility.

ext/QuantumCliffordHeckeExt/lifted_product.jl
Fe-r-oz: 📄 The `_build_parity_matrix_xz(::CommutativeLift`) which is defined as 

hx = [A ⊗ I | I ⊗ B]
hz = [I ⊗ B* | A* ⊗ I]

produce the exact same stabilizer matrices as current repo ` hgp + parity_checks_xz`  double conjugation approach defined abelian LPCodes.

hx = [ A ⊗ I   | I     ⊗ (B*)* ]
hz = [ I  ⊗ B* | A* ⊗  I       ]

Maybe we should add a test for this as well to ensure that we have the exactly same stabilizers for commutative algebra case.

ext/QuantumCliffordHeckeExt/lifted_product.jl
```diff
@@ -216,43 +216,34 @@ end
...
-    Â = concat_lift_repr(c.A_repr, A) # ρ(A) right regular repr
-    B̂ = concat_lift_repr(c.B_repr, B) # λ(B) left regular repr
-    # Compute adjoints with representations:
-    # B̂* = λ(B*) and Â* = ρ(A*)
-    B̂_conj = concat_lift_repr(c.B_repr, permutedims(group_algebra_conj.(B))) # λ(B*)
-    Â_conj = concat_lift_repr(c.A_repr, permutedims(group_algebra_conj.(A))) # ρ(A*)
-    hx = [kron(Â, Matrix(LinearAlgebra.I(mb))) kron(Matrix(LinearAlgebra.I(ma)), B̂)] # [ρ(A) ⊗ I | I ⊗ λ(B)]
-    hz = [kron(Matrix(LinearAlgebra.I(na)), B̂_conj) kron(Â_conj, Matrix(LinearAlgebra.I(nb)))] # [I ⊗ λ(B*) | ρ(A*) ⊗ I]
+    Â, B̂ = is_commutative(c.GA) ? (A, B) : (concat_lift_repr(c.A_repr, A), concat_lift_repr(c.B_repr, B))
+    Â_conj, B̂_conj = if is_commutative(c.GA)
```
Krastanov: This is a bit better, it is more legible and it avoids code repetition, but it does not really address the question, rather it hides it a bit. It is still two different code paths for the commutative and noncommutative case, and I want to understand why two different code paths are necessary.

Fe-r-oz: Thanks! I've unified the code into a single function so we can now focus more clearly on why special handling is required for certain operations. In particular, I’ve removed the separate commutative case (which had been simplified by applying `B_repr` to both `A` and `B)`, and instead wrote a general function that applies conditional logic based on commutative/non-commutative algebra. This probably makes distinction more explicit because placing the if checks directly inside the unified function, it will probably be helpful to better reason about the specific steps for non-commutative algebras.

Fe-r-oz:
Hi Stefan,

While digging into this today and pondering about the root causes, the possible fixes (late-lifting for commutative GA, early-lifting for non-commutative GA), two general questions came up after https://github.com/QuantumSavory/QuantumClifford.jl/pull/511#issuecomment-3030781792, -- I’ll ponder on the second question before we next meet. I hope these two questions are helpful. Apologies for the noise.


1) Why does the late-lifting approach (pathway 1) (which take Kronecker product of group algebra elements, and then lift them to matrices nicely) does not work with the non-commutative group algebra? 

```
# late-lifting approach applied to non-commutative GA has the following affect
hx, hz = hgp(c.A, permutedims(group_algebra_conj.(c.B)))
hx = concat_lift_repr(c.A_repr, hx)  # ρ to [A⊗I | I⊗B*]
hz = concat_lift_repr(c.B_repr, hz)  # λ to [I⊗B | A*⊗I]

instead of
hx = [ρ(A)⊗I | I⊗λ(B)]
hz = [I⊗λ(B*) | ρ(A*)⊗I]
```
This means that representations have to be applied before the kronecker product in the non-commutative GA case which leads to commutativity between parity check matrices in non-commutative case. 👀

2) Why does the early-lifting approach  (pathway 2) (where the Kronecker products are between lifted matrices) fails to accommodate the base_matrix construction method (commutative) of LPCodes?

Early lifting approach entails that the Kronecker product are now between matrices, not group algebra elements. I will now ponder about this case from many aspects as it leads to non-commutativity of parity check matrices in base matrix case which is the issue.

Fe-r-oz:
🎨  Turning to draft to do a minor style commit: adding comments, removing typos in comments inside `parity_check_xz` function.

Krastanov:
Everything else in this PR looks great. I just want to understand better the differences that are happening.

Fe-r-oz:
Thank you for adding a useful test! closed the PR to trigger CI again.

The `test_twobga_codes` contains some codes (`nonabel1-4`) that use non-abelian groups/non-commutative algebra, 

 > append!(test_twobga_codes, [t1b1, t1b3, tb21, tb22, dprod1, dprod2,` nonabel1, nonabel2, nonabel3, nonabel4])`

The test uses codes which contains these non-abelian codes but we are only testing codes built from abelian groups, so this might the reason for the error.

```
codes = vcat(LP04, LP118, test_gb_codes, test_bb_codes, test_mbb_codes, test_coprimeBB_codes, test_hcubic_codes, test_twobga_codes, test_honeycomb_color_codes);
```

Here, we are using the `_parity_matrix_xz_if_comm_algebra(c)` on `twobga_codes` where some of them are built using non-abelian groups which could explain the error.


```
for c in codes
   @test QECCore.parity_matrix_xz(c) == QuantumCliffordHeckeExt._parity_matrix_xz_if_comm_algebra(c)
end
```

The 4 failure in the` buildkite` ECC is due to these 4 codes that use non-abelian groups.

Fe-r-oz:
The Buildkite ECC tests are now passing. The previous four failures occurred because of the non-abelian LPCodes I added in `test_twobga_codes` as part of this PR. These triggered errors in `test/test_ecc_lpcode_simpler_repr.jl` due to `_parity_matrix_xz_if_comm_algebra(c)`, which relies on an older implementation that doesn’t support non-abelian LPCodes.  We can add additional tests from #397 and #402 to further validate non-abelian group functionality.

PR #519, opened 2025-06-23
Implement GPU accelerated Pauli/Tableau multiplication.

Hamiltonian-Action:
Sending out a ping to @Krastanov.

As discussed a few days ago, here is how mul_leftright would look like in the minimal "idiomatic style" without any bells, whistles, allocations, or type safety. The same kernel functions on both Paulis and Tableaus just by altering the launch configuration, due to how Julia handles superfluous indices. All extraneous type information has been omitted, sans those required to match the existing definitions.

This is purposefully marked as a draft because I am entirely aware, and have verified, that this does not function unless some liberty is granted to workaround some of the issues that I had already solved previously. Beyond Metal which does not yet support atomics, see: `KA.supports_atomics(::MetalBackend) = false` in the relevant repository, both CUDA, ROCm, and OneAPI only support atomic operations on values of certain sizes. Namely, by consulting the relevant documentation, they are 32, 64, and 128 bit wide (vectors of) integers and floats that are naturally aligned. Consequently, atomically modifying the UInt8 phase will raise a runtime exception and hang the relevant device.

I will present the available alternatives that I managed to devise after having pondered this issue over the past two days, as well as some brief commentary on each of them.

1. Allocate an auxiliary buffer and perform two kernel launches. The first kernel will update the operator bits and write the partial phases to the buffer, whilst the second one will accumulate the buffer and write the updated phase at the appropriate destination.

This solution is the most encompassing as it does not require atomics at all and can properly work on all backends, though the size of the required buffer may become relatively large depending on the operands, as it necessitates `cld(prod(size(xzs)), 256)` values to be stored.

2. Allocate an auxiliary buffer and perform two kernel launches. The first kernel will update the operator bits and atomically accumulate on the buffer, whilst the second one will just copy the phase value from the buffer and to the appropriate destination.

This is a variation of the previous solution that loses on the flexibility side, given that atomics are now required, but reduces the buffer memory requirements to merely `size(xzs, 2)`.

3. Re-define QuantumClifford such that the phase becomes a UInt32 value.

The only solution on the list that makes this draft work right out of the box. This is definitely ill-advised due to how much work would be required elsewhere in order to make this function properly, as well as potentially causing a breaking change wherever a value of UInt8 is assumed for the phase.

4. Implement a mutex and then utilise it to update the phase.

The worst "solution" of the bunch and I am only including it for completeness. Disregarding the scope issues, this will necessitate the same atomicity and additional storage requirements as the second solution, whilst also absolutely ruining any hope of proper GPU utilisation due to having the resident threads enter a busy waiting loop until they can acquire the mutex.


src/mul_leftright.jl
```diff
@@ -125,29 +126,57 @@ end
...
+        s = mul_right!(l.xz, r.xz, phases=phases)
+        B && (l.phase[] = (s+r.phase[]+l.phase[])&0x3)
+    end
     l
 end
 
 @inline function mul_left!(r::PauliOperator, l::Tableau, i::Int; phases::Val{B}=Val(true)) where B
-    s = mul_left!(r.xz, (@view l.xzs[:,i]), phases=phases)
-    B && (r.phase[] = (s+r.phase[]+l.phases[i])&0x3)
+    if KA.get_backend(r.xz) isa KA.GPU
```
Krastanov: Why is this check necessary? Can't we just dispatch on array type?

Seeing `isa` generally makes me suspicious.

Krastanov (changes requested):
Thanks for setting this draft up! I have a minor question about the current code structure (mentioned inline as well):

I am imagining that simply by having an appropriate new method for `mul_ordered!` with GPU array types, all the other methods will just specialize as appropriate without having to change any of their code (or at least having less invasive code than something like an `isa` if-else block). Anything stopping you from doing it that way?

You are probably already thinking about something like this, but could you add OpenCL tests (so that we can check the kernels for correctness without having a GPU)?

src/mul_leftright.jl
Hamiltonian-Action: I am doing this to prevent the existing code from indexing device arrays from the host. Backends will throw an error if you attempt to do that unless some other flag is set, but I do not know if this can be done portably. In any case, setting said flag will force the code to synchronise whenever there is a host-side indexing operation on a device array, which is definitely undesirable.

Krastanov:
> That is, I would need to define functions that are explicitly `PauliOperator{<: AbstractGPUArray{UInt8, 0}, <: AbstractGPUVector{<: Unsigned}}`

There was probably a misunderstanding. Doing this type of dispatch is very much the correct way to structure the code.

Although for this particular case it can be done probably even more simply: `mul_ordered!` (which currently has only two methods) can receive a new method specialized on GPU Array and no other existing functions need to be modified.

> Perhaps I can pre-emptively write the test files and then freeze this PR until that gets released?

Yup, that sounds good.

But even if we can not use OpenCL for tests, we can use `dev=CPU()` for such tests right now, right?

Something like this:

- create a new `mul_ordered_KA_kernel` (potentially with a better name)
- test its correctness with CPU arrays
- create a `mul_ordered!(::GPU Array, ...)`  that calls into `mul_ordered_KA_kernel`
- add a test for `mul_ordered!` that runs only when CUDA is available and it uses a GPU array
- (in the future) when there is OpenCL support, add tests that call `mul_ordered` with an OpenCL array -- for now it can just be marked with `broken` so it is present in the test runner but known to not work

Krastanov:
Good point on mul_ordered being used inside of other kernels. That would indeed lead to some complications.

> I reckon the only tests should be for `CPU()` and possibly `OpenCL`

Definitely not. Supposedly the value of KernelAbstractions is that it is trivial to run the code on various GPUs. If that is the case, we will definitely have at the very least CUDA tests. AMDGPU tests would be nice to have as well. For Metal we will need to figure out whether we can find CI runners, but having the tests exist should be easy.

Krastanov:
It seems a bit wasteful to go for uint32, but this is negligible in the asymptotic regime, so for now let's stick to option 3 (switch to uint32 for phases) as it is simplest. Try out a PR that makes just that change (presumably only a few lines of code need to be changed), and if it raises too much havoc we can get back to option 2 (having a uint32 buffer) which would have proven to be simpler by that point.

Could you confirm that your current plan with this PR is to make `mul_*!` into a kernel but keep all consumers of that kernel as normal functions (as long as they do not need to do indexing)? So presumably this will lead to:

- significant speedup in `mul_*!` operations on very large states
- small speedup to `apply!(stab, clifford)` and `apply!(stab, pauli)`
- no or negligible speedup to `canonicalize*!`
- it is not touching `apply!(stab, small_named_clifford)`

src/mul_leftright.jl
```diff
@@ -130,45 +131,147 @@ end
...
+    mul_device!(
+        l.phase, l.xz, r.phase, r.xz;
+        phases = phases, order_right_left = Val(false)
+        )
+    return l
+
+end
+
 @inline function mul_left!(r::PauliOperator, l::Tableau, i::Int; phases::Val{B}=Val(true)) where B
+    nqubits(l)==nqubits(r) || throw(DimensionMismatch("The Pauli operator and the Tableau should have the same length!"))
```
Krastanov: This is an internal function used in a lot of low-level code. This type of checks are certainly valuable and should be added at some point, but it should be done in a way that can be elided, as the check can be expensive and I do not believe we can rely on the compiler to do reliable constprop for it:

1. We can use the `@inbounds` capabilities from `Base`. That is probably the cleanest way to do it, but it would require sprinkling annotations wherever `mul_left!` is used.
2. Or we can move the equality check to the user-facing public API (that would be a part of solution 1 anyway, but it is valid to just stick to this, without implementing the rest of solution 1)

test/runtests.jl
```diff
@@ -5,6 +5,7 @@ if Sys.iswindows() || Sys.ARCH != :x86_64
 elseif get(ENV, "GPU_TESTS", "") == "true"
     @info "running with GPU tests"
     Pkg.add("CUDA")
+    Pkg.add("AMDGPU")
```
Krastanov: I do not expect there are many machines that have both CUDA and AMDGPU hardware. These should be under two separate flags (we can just change the flag to something like `GPU_CUDA_TESTS` and `GPU_AMDGPU_TESTS`).

test/test_mul_leftright.jl
```diff
@@ -35,4 +35,245 @@
       @test np==cp
     end
   end
+
+  @testset "OpenCL" begin
```
Krastanov: Let's move these to a separate file, something like `test_KernelAbstractions_OpenCL.jl`. Or maybe even create a whole new subfolder just for KA tests.

test/test_mul_leftright.jl
```diff
@@ -35,4 +35,245 @@
...
+            @test h_o.phases[i] == Array(d_o.phases)[i]
+            @test (@view h_o.xzs[:, i]) == (@view Array(d_o.xzs)[:, i])
+          end
+        end
+      end
+    end
+  end
+
+  @testset "ROCm" tags = [:gpu] begin
+    using AMDGPU
```
Krastanov: separate file and better filtering so AMDGPU and CUDA can be filtered separately

test/test_mul_leftright.jl
```diff
@@ -35,4 +35,245 @@
...
+          @test (@view h_o.xzs[:, i]) == (@view Array(d_o.xzs)[:, i])
+        end
+      end
+    end
+  end
+
+end
+
+@testitem "Mul leftright -- GPU" tags = [:gpu] begin
+  @testset "CUDA" begin
```
Krastanov: separate file and better filtering so AMDGPU and CUDA can be filtered separately

test/test_mul_leftright.jl
```diff
@@ -35,4 +35,245 @@
...
   end
+
+  @testset "OpenCL" begin
+    using pocl_jll, OpenCL
+    using KernelAbstractions: synchronize, get_backend
+    using QuantumClifford: mul_left!, mul_right!, Tableau
+    # Small sizes for encoding issues, large sizes for race conditions.
+    test_sizes = [31, 32, 33, 63, 64, 65, 127, 128, 129,
+                  64 * 1023, 64 * 1024, 64 * 1025,
+                  64 * 2047, 64 * 2048, 64 * 2049]
```
Krastanov: Creating a random stabilizer that is `64*2049`-wide would take quite a while. This would make our slow test runner even slower. I think equivalent tests can be executed with something like a random PauliFrame. Even much larger random PauliFrames can be constructed extremely fast, so there will be no slowdown.

test/test_mul_leftright.jl
```diff
@@ -35,4 +35,245 @@
...
+
+    skip_flag = true
+    if !skip_flag
```
Krastanov: why is this flag here?

test/test_mul_leftright.jl
```diff
@@ -35,4 +35,245 @@
...
+
+      using KernelAbstractions: synchronize, get_backend
+      using QuantumClifford: mul_left!, mul_right!, Tableau
+      # Small sizes for encoding issues, large sizes for race conditions.
+      test_sizes = [31, 32, 33, 63, 64, 65, 127, 128, 129,
+                    64 * 1023, 64 * 1024, 64 * 1025,
+                    64 * 2047, 64 * 2048, 64 * 2049]
+
+      skip_flag = true
+      if !skip_flag
```
Krastanov: same question about the flag

src/mul_leftright.jl
Krastanov: that can be done in a separate PR

src/mul_leftright.jl
```diff
@@ -130,45 +131,147 @@ end
     r
 end
 
+@inline function mul_left!(
```
Krastanov: let's keep these new methods in a separate file (maybe put all of the new kernelabstractions work in a separate folder)

test/test_mul_leftright.jl
Krastanov, line 1: all of the new tests in this file seem to be fairly similar in style. Let's restructure them so that when we need to add something new to them, we edit only one location. E.g. there can be a `setup_KA_test_mul_leftright.jl` that implements `function mul_leftright_tests(datatype)` and then `test_KernelAbstractions_OpenCL.jl` just imports that file and runs the test function with an appropriate data type. No need to follow this suggestion exactly, there might well be issues with the specific example I am giving, but something in that style would be helpful.

test/test_mul_leftright.jl
Hamiltonian-Action: Because the tests, as they stand, are incompatible with how the phase is handled at the time of writing. Once that issue is cleared up, the tests can be enabled through a simple flick.

test/test_mul_leftright.jl
Hamiltonian-Action: Same answer as the above.

src/mul_leftright.jl
Hamiltonian-Action: Will remove, add a comment, and revisit later.

src/mul_leftright.jl
Hamiltonian-Action: I believe that grouping similar definitions together would provide more clarity than separating things out, though I guess having `KA/mul_leftright.jl` is not the worst thing imaginable.

test/runtests.jl
Hamiltonian-Action: It is actually harmless, since the tests themselves query the devices lists, but having actual environment variables set up is even better. This was only done in this manner since I did not get a response around the time I was fleshing it out.

test/test_mul_leftright.jl
Krastanov: During development this makes a lot of sense, but no need to keep this when preparing for merge. The merge of this PR will be waiting on whatever other prerequisite PRs are being submitted anyway.

src/mul_leftright.jl
Krastanov: Having them separate would make it easy to move this to a pkg ext if we decide to. KernelAbstractions is not that light of a dependency (bringing in LLVM and GPUCompiler). The load time of QuantumClifford.jl is already pretty bloated and KA adds another 10%-20% to it.

Krastanov:
No worries about imprecision in the benchmarks. Those are really not that important, I just wanted a rough idea of the speedup. Normalization of the benchmarks is not necessary and time spent on that would be much better used working on the rest of the kernels. The benchmarks are there just to give an order of magnitude idea.

The CI server I will be setting up will mostly be consumer/workstation hardware.

As a baseline comparison I am also curious about the following:

- comparing `b .= b .⊻ a` vs `mul_left!(b,a; phases=false)` both on CPU and GPU on a log-log scale
- comparing `reduce(⊻, b)` vs `mul_left!(b,a; phases=true)` both on CPU and GPU on a log-log scale

That would give a good idea of whether there are some common optimizations still missing from the current implementation of these kernels.

Krastanov:
> as it is unable to smoothly parse `reduce(xor, b)` due to some complaints about not being aware of the neutral element.

My recollection is that there is a standard way to provide the neutral element, either by defining a new method or with some keyword argument.

The legend labels are not very clear to me. What does each color correspond to?

Krastanov:
The main reason I was asking about this is because you have already started working on `mul_ordered!` kernels which are used inside of the loop of `canonicalize`, so it would be nice to be able to reuse them.

I am wondering whether the following abstraction might make sense:

- make a macro that provides the equivalent of `mul_ordered!` but it takes the indexing variable as an argument
- use that macro to make `mul_ordered!`
- reuse that macro for the row operations in `canonicalize`

This is just some idle musing of mine -- you are the one in the trenches, so just disregard it if it does not seem like it would actually lead to code simplification.

My main worry is that there are many different types of canonicalization and it would be pretty sad to have to repeatedly code up the equivalent of `mul_ordered!` in their inner loops.

Krastanov:
Sorry, but I would not have the time to help with testing draft code.

Krastanov:
Could you give some small selfcontained examples? Something that does not require a lot of exposition, just a small piece of code and a command to test it.

Krastanov:
This has not been marked as in-need-of-review (upper right corner of this page) and it has failing tests, so I have not looked at it. Generally, if you are waiting on me for some of the PRs, feel free to ping me if you have been waiting for more than a day or two.

The downgrade test is not failing on master, so it should be fixed here (presumably a dependency you have added causes an inconsistency with the lower bounds of something else -- bumping the lower bound of that something else should fix the issue).

"n threads on k cores" where n<k is perfectly fine -- the cores are oversubscribed, but that would not be a cause for a failure.

I will get you the nvidia runner later tonight or tomorrow, and after they pass we should be good to go for a final review.

We might want to move this to a package extension instead of keeping it in the main package, to avoid compilation time bloat, but as long as everything is kept in a separate file that would be trivial to do in a future PR before release.

Krastanov:
This is from the downgrade tests and it seems to be a real issue, not just a fluke:

image

Krastanov:
The ubuntu failures are also not happening on master where we have the same threads=5 > cpus=4, so they seem to be real as well, although having very unhelpful error messages.

Krastanov:
and the Github Action failure on ubuntu with threads=5, cpus=4, also happens in buildkite with threads=16, cpus=16, so this is definitely a new issue

test/runtests.jl
```diff
@@ -56,6 +57,8 @@ testfilter = ti -> begin
 
     if !OpenCL_flag
         push!(exclude, :opencl)
+    else
+        return :opencl in ti.tags
```
Krastanov: as of the very last commit, given that OpenCL_flag is unconditionally set to true, the vast majority of tests are just being skipped, right? So there is no crash simply because we do not run the crashing test at all, in any of the test runners?

Krastanov:
that is a pretty wild.. bug? In what though... it is weird to have a segmentation fault due to an interaction between two packages that are unrelated. Locally I can not trigger the crash with just:

```
]add Oscar GPUArrays OpenCL pocl_jll
using pocl_jll, OpenCL, GPUArrays, Oscar
```

Thanks for looking into this! It is a very annoying sysadmin-ish problem, that has little to do with the fun algorithmic side of things. Let's have just a separate OpenCL runner for now, so that all tests are run, but the environment in which this crash happens is never created. We will just not mark any of this functionality public until we know what is causing it. If we move this functionality to an extension, the crashing environment will not be autocreated either, so it should be safe (we can do that move before release, no need to delay this PR further)

test/runtests.jl
Hamiltonian-Action: That was intentional, just to get the runners to exclusively run the OpenCL test and nothing else. It allowed confirming that they are both individually fine but only error out when used in conjunction. The last few commits are bogus and I will clean them up before merging.

Krastanov:
How do you want to proceed? A few options that come to mind:

- have a separate runner for OpenCL stuff, so it is not present in the Oscar environment where the crashes seem to happen
- move things to an extension in this PR so that from the very beginning we have more compartmentalized compilation and a smaller test environment Project.toml
- other options you prefer?

Krastanov:
The weakdeps should be backend independent, so presumably just KA (and Atomix or similar transitive dependencies if necessary). Just ping me when you make a decision on which way to go (splitting into an extension now or later) and want to do a final review.

Krastanov:
I will make a small commit here shortly, to add the CUDA runner. Feel free to overwrite / modify it.

Krastanov:
Wonderful job, thank you!

PR #521, opened 2025-06-23
Fixed incorrect phases on certain single qubit symbolics.

Hamiltonian-Action:
Previous operator definitions violated unitarity restrictions, namely mapping `I ~> -I`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 
 ## v0.10.0 - 2025-06-23
 
+- **(fix)** The gates `SQRTY`, `CXYZ`, `CZYX` were computing phases incorrectly when acting on `I` stabilizers.
+- **(fix)** `rowdecompose` was not accounting for the phase of the input Pauli string, leading to potential errors in non-Clifford functionality.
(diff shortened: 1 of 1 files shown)


Krastanov:
Good catch, this is a pretty significant correctness bug. Surprising that the tests have not caught this before.

@Fe-r-oz ,  do you remember how you implemented these? I think it happened in #333 . Did you copy them from stim or derive with some boolean formula simplifier? Do not worry about the mistake itself, I did not catch it either. It would be useful to know where these come from so we can fix the mistake in the source or just generally make sure that similar mistakes do not happen in the future.

Fe-r-oz:
Good catch.

> Did you copy them from stim or derive with some boolean formula simplifier? 

I followed the approach mentioned in the following line and used the [dcode](https://www.dcode.fr/boolean-expressions-calculator) boolean simplfier. Stim was used for consistency checks. Thank you for the correctness checks! https://github.com/QuantumSavory/QuantumClifford.jl/blob/2a0c6742d7eb7ba62bcae4beaa2329f7937fed1e/src/symbolic_cliffords.jl#L349

PR #522, opened 2025-06-24
add gpu canonicalize! support

riteshbhirud:
This PR adds support for GPU canonicalization while keeping the core algorithm intact. It replaces our custom converter with the built-in to_gpu utility, incorporates all prior feedback, and includes a series of housekeeping and performance optimizations to make the codebase more concise and maintainable.

@Krastanov this is the updated and enhanced version of my old [PR] (https://github.com/QuantumSavory/QuantumClifford.jl/pull/485#pullrequestreview-2729474364) done for my thesis project. I made a new PR and closed the old one as that branch became a bit cluttered and confusing. 

Changes:

GPU Support: Leverage the framework’s native to_gpu converter for canonicalization.

Code Cleanup: Remove obsolete custom converter code and streamline related modules.

Optimizations:

Reduce intermediate allocations.

Inline helper functions where appropriate.

Improve loop unrolling and memory access patterns.

Feedback Integration: Addressed all comments from the previous review cycles. 

Example usage: 

 cpu_stab = random_stabilizer(6000)
gpu_stab = to_gpu(cpu_stab) # using built-in to_gpu instead of custom converters
 canonicalize!(gpu_stab) #automatic dispatch for gpu type


.vscode/settings.json
Krastanov, line 1: remove this, add to .gitignore

ext/QuantumCliffordGPUExt/canonicalization.jl
```diff
@@ -0,0 +1,235 @@
...
+const GPUTableau = QuantumClifford.Tableau{CuArray{UInt8, 1, CUDA.Mem.DeviceBuffer}, CuArray{UInt32, 2, CUDA.Mem.DeviceBuffer}}
+const GPUStabilizer = QuantumClifford.Stabilizer{GPUTableau}
+
+struct GpuTableau
+    d_mat::CuArray{UInt64,2}         
+    d_phases::CuVector{UInt8}   
+    nqubits::Int                
+    nblocks::Int                
+    r::Int                      
+end
```
Krastanov: Let's remove this struct, and stick to the Tableau struct (that you have specifically parameterized above for a GPU). That would lead to some streamlining

ext/QuantumCliffordGPUExt/canonicalization.jl
```diff
@@ -0,0 +1,235 @@
...
+    d_mat_32 = stab.tab.xzs
+    d_phases = stab.tab.phases
+    nqubits  = stab.tab.nqubits
+    r        = size(d_mat_32, 2)
+    nblocks  = cld(nqubits, 64)
+    
+    d_mat_64 = reinterpret(UInt64, reshape(d_mat_32, (:,)))
+    d_mat_64 = reshape(d_mat_64, (size(d_mat_32, 1) ÷ 2, size(d_mat_32, 2)))
+    
+    gt = GpuTableau(d_mat_64, d_phases, nqubits, nblocks, r)
```
Krastanov: all of this should be unnecessary if we remove GpuTableau

ext/QuantumCliffordGPUExt/canonicalization.jl
```diff
@@ -0,0 +1,235 @@
...
+    d_mat     = gt.d_mat
+    d_phases  = gt.d_phases
+    r_host    = gt.r
+    nqubits_host = gt.nqubits
+    nblocks_host = gt.nblocks
+    r        = Int32(r_host)
+    nqubits  = Int32(nqubits_host)
+    nblocks  = Int32(nblocks_host)
+    total_blk = Int32(2) * nblocks
+    current_col = Int32(1)
```
Krastanov: are all of these Int32s actually necessary? Can we have this a bit neater, just with type parameters and less hardcoding. If there is a benchmark that actually shows different performance, then I am convinced to have them, otherwise let's stick to the prettier to read code.

ext/QuantumCliffordGPUExt/canonicalization.jl
```diff
@@ -0,0 +1,235 @@
...
+    for j in Int32(1):nqubits
+        pivot_row = (j - Int32(1)) ÷ Int32(64) + Int32(1)
+        bit = UInt64(1) << ((j - Int32(1)) % Int32(64))
```
Krastanov: and here it is a UInt64?

ext/QuantumCliffordGPUExt/canonicalization.jl
```diff
@@ -0,0 +1,235 @@
...
+    @cuda threads=threads blocks=blocks pivot_search_kernel!(
+        d_mat, Int32(pivot_row), bit, Int32(current_col), Int32(r), d_result
+    )
+    
+    result = Array(d_result)[1]
+    return result > r ? 0 : Int(result)
+end
+
+function update_kernel!(d_mat, d_phases, current_col::Int32, pivot_row::Int32, bit::UInt64,
+                        r::Int32, total_blocks::Int32, nblocks::Int32, update_phase::Bool)
```
Krastanov: a lot of hardcoded types here and elsewhere -- let's turn those into nicely named type parameters, and if necessary we can constrain those parameters later

ext/QuantumCliffordGPUExt/canonicalization.jl
```diff
@@ -0,0 +1,235 @@
+using CUDA
+using LinearAlgebra
+using QuantumClifford  
+
+const MIN_BLOCKS = 36
+const GPUTableau = QuantumClifford.Tableau{CuArray{UInt8, 1, CUDA.Mem.DeviceBuffer}, CuArray{UInt32, 2, CUDA.Mem.DeviceBuffer}}
```
Krastanov: this type might have unnecessary amount of details in it (maybe you can skip the whole `Mem...`)

Krastanov (commented):
I need to set up the cuda test runner tomorrow, but the tests seem good enough. I have not done a particularly in-depth review, but as a first CUDA implementation this provides sufficient capabilities and tests to be merged. We can do any improvements to performance as they become profiled.

And in parallel the KA.jl work will provide an alternative -- we can simplify the code base when that exists.

riteshbhirud:
@Krastanov wanted to make sure if everything is all good for this PR?
Thanks!

Krastanov:
I merged master into here to rerun the tests with the new test runner

Krastanov:
@Hamiltonian-Action , I know you are working on related functionality. For the moment I will merge this one, but we can make it private when a KA implementation becomes available.

Krastanov:
@riteshbhirud , there seem to be test failures (see the buildkite cuda runner and screenshot below). Could you investigate?

image

riteshbhirud:
@Krastanov I fixed some critical bugs introduced in the type parameter generalization version which slipped through testing possibly due to older/correct version's function definitions still in my testing terminal. In one or two places I had to replace the incorrect generic type with actual expected types. Also in some places T<:Unsigned was changed to T as T is guaranteed to be UInt64 and it was unnecessary overhead as well. 
As per my understanding this still provides the generalization we needed and also ensures accuracy and performance. 

Please let me know if any more changes are needed.

ext/QuantumCliffordGPUExt/canonicalization.jl
```diff
@@ -0,0 +1,218 @@
...
+                          r, total_blk, nblocks, phases)
+        current_col += 1
+        if current_col > r
+            break
+        end
+    end
+    return tableau
+end
+
+function canonicalize!(stab::Stabilizer{<:QuantumClifford.Tableau{<:CuArray{UInt8}, <:CuArray{T}}}; phases::Bool=true) where {T<:Unsigned}
```
Hamiltonian-Action: Overly explicit. A (relatively) recent PR has completely parametrised the phase so that UInt8 is no longer required. As a matter of fact, there is no need to even dispatch on the element type of the `CuArray` since being inside a `Tableau` already ensures their conformity. I am only noting this here but it can be applied to all such similar signatures within this PR.

That is not to say that dispatching on a particular element type is bad, but doing so should be used for some other purpose such as ensure correctness -- e.g: `l::CuVector{T}, r::CuVector{T}` necessitates that both of them have the same element type.

Krastanov:
As mentioned in other comments, in a few places you can relax the type signatures in the methods you have implemented. Having the `T` type parameter is not wrong, but it does not provide much value unless it is being used (and you do appropriately have `T(0)`  in a bunch of places as a way to construct appropriately typed zeros). Up to you whether you want all your signatures to look similar (and potentially having a few redundant T parameters) or having your signatures be no more complex than necessary -- I am happy with either option.

But in your last commit you seem to have removed an appropriate use of the type parameter T and instead have hardcoded the assumption that T will be 64 bits long. I do not think this assumption is safe. Nothing stops someone from making a Tableau in which the storage array is CuMatrix{UInt32}. Could you bring back that parameterization? Let me know if I have missed some other reason for which your approach might be preferable.

You also do not need to worry about overhead from things like `sizeof(T)` or other type-parameter arithmetics. These will get evaluated at compile time.

ext/QuantumCliffordGPUExt/canonicalization.jl
riteshbhirud: Thanks for the valuable feedback! I will surely keep this in mind when fixing the code. 
Thanks

riteshbhirud:
@Krastanov Thanks for the feedback. I fixed all specific hardcoding while ensuring accuracy in this update and also got rid of the Int8 for phases (I was not aware earlier of the recent change earlier). 

Please let me know if this looks good.

test/test_gpu_canonicalization.jl
```diff
@@ -0,0 +1,41 @@
+@testitem "GPU Canonicalization" tags=[:cuda] begin
+    using CUDA
+    using QuantumClifford
+    using Random
+    using QuantumClifford: to_cpu, to_gpu
+    
+    if CUDA.functional()
```
Hamiltonian-Action, line 7: This call is to be deprecated at some point in the future. It is fine for now but perhaps consider using `length(devices()) > 0` instead.

test/test_gpu_canonicalization.jl
```diff
@@ -0,0 +1,41 @@
...
+                cpu_stab = random_stabilizer(n)
+                gpu_stab = to_gpu(cpu_stab)
+                cpu_result = canonicalize!(copy(cpu_stab))
+                gpu_result = canonicalize!(copy(gpu_stab))
+                cpu_from_gpu = to_cpu(gpu_result)
+                @test cpu_result == cpu_from_gpu
+            end
+        end
+        
+        @testset "GPU canonicalize! performance" begin
```
Hamiltonian-Action, line 19: I reckon tests should be for correctness whereas benchmarks ought to be for performance. Reduces false-negatives that may transpire due to hardware variation.

ext/QuantumCliffordGPUExt/canonicalization.jl
```diff
@@ -0,0 +1,194 @@
+using CUDA
+using LinearAlgebra
+using QuantumClifford  
+const MIN_BLOCKS = 36
```
Hamiltonian-Action: Is there any justification for this value? It is not a clean power of two so nothing really comes to mind and the minimum block count should ideally just be one, i.e: no need to ask for more than what is genuinely needed.
Additionally, this is being defined as a constant at the module scope. Assuming that this variable is specific just to the canonicalisation family, perhaps a less generic name should be assigned so as to avoid potential conflicts? If the generic name is intentional since it will in fact be used elsewhere then the definition should be moved somewhere more visible.

ext/QuantumCliffordGPUExt/canonicalization.jl
```diff
@@ -0,0 +1,194 @@
...
+                                     col1::Integer, col2::Integer) where {T, P}
+    N = size(d_mat, 1)
+    threads = 256 
+    blocks = max(cld(N, threads), MIN_BLOCKS)
+    @cuda threads=threads blocks=blocks swap_columns_and_phases_kernel!(
+        d_mat, d_phases, col1, col2, N)
+    return
+end
+
+function swap_columns_kernel!(d_mat::CuDeviceArray{T,2},
```
Hamiltonian-Action: Code redundancy. This is not wrong per-se but using a `::Bool` or `::Val` argument/keyword-argument allows merging both definitions into one since it will be handled at runtime or compile time respectively.

ext/QuantumCliffordGPUExt/canonicalization.jl
```diff
@@ -0,0 +1,194 @@
...
+end
+
+function gpu_canonicalize!(tableau::QuantumClifford.Tableau{<:CuArray{P}, <:CuArray{T}}, phases::Bool) where {T, P}
+    d_mat = tableau.xzs
+    d_phases = tableau.phases
+    nqubits = tableau.nqubits
+    r = size(d_mat, 2)
+    nblocks = size(d_mat, 1) ÷ 2
+    total_blk = 2 * nblocks
+    bits_per_chunk = sizeof(T) * 8
```
Hamiltonian-Action: Highly pedantic but this assumes that `Cchar` is 8 bits. Using `count_zeros(zero(T))` is more portable.

test/test_gpu_canonicalization.jl
Krastanov: We do have other performance tests as well, As long as the test is not flaky, it is a reasonable way to protect against drastic regressions. We do not run GPU benchmarks on each commit, so having this test is nice.

Think of it as a basic sanity/correctness check: "if the GPU is slower than the CPU for this type of work, then there is something very wrong with the code", not as a benchmark meant to track performance with precision.

Before merge I might remove some of the particularly slow-to-run instances just to keep test times reasonable. There are a couple of ways to make big tableaux without using the particularly slow randomize operation.

riteshbhirud:
Thanks again for the comments. I just updated my code to enhance it. Please let me know if it looks good now

riteshbhirud:
> Thanks again for the comments and helping to convert my independent study project into polished piece of code. I just updated my code to enhance it. Please let me know if it looks good now

@Krastanov Please let me know if this is good to merge for now once you have a moment.

Krastanov:
this is well separated into an extension and all tests pass, so let's merge without putting it in the CHANGELOG (in order to have it tested a bit more in usage before we commit to it publicly)

Thanks @riteshbhirud , this was extremely useful addition!

@Hamiltonian-Action , I know you are working on a KA implementation in a separate extension. If these start clashing (they will on import of CUDA), just rename these methods to something with an underscore prefix, so we can keep them as tests at least. We can worry about what is the proper way to keep these around after we have had them co-exist privately for a while.

PR #523, opened 2025-06-24
Fixing incorrect phase when applying Paulis.

Hamiltonian-Action:
Please do forgive me if I happen to be mistaken, but the proper application procedure should be that $S \mapsto P S P^\dagger$ so the phases of $P$ and $P^\dagger$ cancel each other out. The current implementation behaves as if $S \mapsto P S P$, including a two-fold phase contribution from the Pauli operator.

@Krastanov, since I have stumbled across this issue twice over so far, perhaps a large scale test is in order? The quickest sanity check is to confirm that $I \mapsto I$, as required by unitarity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,10 @@
 
 # News
 
-## v0.10.0 - 2025-06-23
+## v0.10.0 - 2025-06-25
 
 - **(fix)** The gates `SQRTY`, `CXYZ`, `CZYX` were computing phases incorrectly when acting on `I` stabilizers.
+- **(fix)** Paulis with imaginary phases had their phases incorrectly tracked.
 - **(fix)** `rowdecompose` was not accounting for the phase of the input Pauli string, leading to potential errors in non-Clifford functionality.
 - Various stabilizer data structures can now be constructed out of sequences of Paulis.
 - Most canonicalization routines can now act on sequences of Paulis.


Krastanov:
Another good catch! Thanks, these are of incredible utility! To the general question, a large additions to correctness tests is always a good idea. I doubt we have any significant tests for imaginary weights as they are not really permitted when working with Clifford circuits, but that is not a reason to not fix these edge cases.

Could you add a test for these so this can be merged?

Krastanov:
There is a lot of functionality in the library, some of it related to extensions that are dependent on python artifacts that needs to be tests. There are also a lot of "random sampling" tests. All of these are necessary to avoid exactly the type of bug you just detected. If anything, we probably should be adding Supposition.jl tests.

For convenience, you can rely on CI for running the tests (still slow, but it does not block your local device) -- you can just mark a PR as a draft and push the unfinished tests and see whether the testrunners are happy with them. You can also filter what tests you run locally, either in the VS Code test runner, or by just running the tests manually in the REPL, or by using something like [TestPicker.jl](https://discourse.julialang.org/t/ann-testpicker-interactive-tool-for-testing-based-on-fzf-and-bat/128273/3).

But overall, if you have done a quick REPL check that your test roughly works, feel free to just throw it to the test runner in a draft PR -- no need to wait locally.

Lastly, there are certainly places where we can trim tests, but not by removing tests or dependencies, rather by checking for tests that take too long because they run too many repetitions. A lot of our random sampling consistency tests do unnecessarily many repetitions. We can probably save up to 20% of test time by cutting those down.

The buildkite testrunner also runs the tests of libraries that depend on QuantumClifford to make sure we do not have downstream breakages.

Krastanov:
Thanks! I moved things around a bit, renamed the tests (the vast majority of our tests are consistency/sanity checks so I wanted something more specific), and added a changelog (so that folks hunting bugs in their code would know why the bug disappeared after they updated a dependency).

PR #524, opened 2025-06-25
WARNING: DO NOT MERGE!! Draft for swapping phase to UInt32.

Hamiltonian-Action:
My condolences go out to the poor CI runners that will have to suffer through this.

### DIFF
--- a/src/QuantumClifford.jl
+++ b/src/QuantumClifford.jl
@@ -170,7 +170,7 @@ struct Tableau{Tₚᵥ<:AbstractVector{UInt32}, Tₘ<:AbstractMatrix{<:Unsigned}
 end
 
 function Tableau(phases::Tₚᵥ, nqubits::Int, xzs::Tₘ) where {Tₚᵥ <: AbstractVector{<: Unsigned}, Tₘ <: AbstractMatrix{<: Unsigned}}
-    Tableau(convert.(UInt32, phases), nqubits, xzs)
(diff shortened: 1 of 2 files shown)


Krastanov:
I am not sure I follow. The following works fine for me: `convert.((UInt,), fill(0))`

Krastanov:
This is great news!

1. The docstrings might not need to be changed at all. There is a regex pass in the doctest string comparison that is user accessible. In principle there might be a way to just tell it to canonicalize all `0x0*01` strings to just `0x01`.

2. You might already be thinking of this, but thee PR with the change to the phase type should be a separate PR which does not involve any other changes.

3. For the moment it is my belief that the default should remain UInt8 and that this should not be a breaking release at all. We can add a test runner in CI that changes that default to UInt32 for all runs. E.g. the testrunner can apply a patch to `const DefaultPhaseType = ...` to modify it before compiling the package and running all the tests.

Krastanov:
And if we go the route of (3), we can just skip the doctests and not worry about (1).

Krastanov:
> ensure compatibility with #519

The capability of having UInt32 is necessary for #519, but that being a default is not necessary, right? After all, any KernelAbstraction capabilities are there just for `Pauli{<:AbstractGPUArray}`, so no need to make the CPU side have similar constraints.

Or maybe you are saying that making Pauli parameterized by the phase type would involve more work than just changing the hardcoded type? That is valid, and as long as the benchmarks do not show big drop in performance, we can just stick to a new hardcoded phase type.

PR #525, opened 2025-06-26
fix generation of classical `LiftedCode` and correct parameters calculation

Fe-r-oz:
The classical LiftedProduct code parity checks matrices have not really been used as seed for any quantum code so this bug was not detected. 


Krastanov (changes requested):
thanks, this is great! Could you add a small test for what used to fail, just to make sure that similar failures do not get reintroduced at some point.

Fe-r-oz:
Thank you for your feedback! I have added basic tests with `tests_nowarn` that just check there is no method error calling stuff that resulted in method errors, I will check literature about properties on these classical codes. I have placed the tests in `test_ecc_throws.jl`.

The stacktrace of the errors resulting from `code_s, code_n` and the suggested fix:
```
julia> import Hecke; import QuantumClifford.ECC: LiftedCode, code_n, code_k, code_s
julia> import QuantumClifford.ECC.QECCore: parity_matrix
julia>  base_matrix = [0 0 0 0; 0 1 2 5; 0 6 3 1]; l = 3;
julia> c LiftedCode(base_matrix, l);
julia> size(zero(c.GA), 2 or 1) # currently, we use this in code_n or code_s that throw error
ERROR: MethodError: no method matching size(::GroupAlgebraElem{FqFieldElem, GroupAlgebra{FqFieldElem, FinGenAbGroup, FinGenAbGroupElem}}, ::Int64)
The function `size` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  size(::BitVector, ::Integer)
   @ Base bitarray.jl:107
  size(::Hecke.ZZRingElem_Array_Mod.ZZRingElem_Array, ::Int64)
   @ Hecke ~/.julia/packages/Hecke/RwzTY/src/Sparse/ZZRow.jl:65
  size(::Type{SIMD.Vec{N, T}}, ::Integer) where {N, T}
   @ SIMD ~/.julia/packages/SIMD/hyXY3/src/simdvec.jl:74
  ...

Stacktrace:
 [1] top-level scope
   @ REPL[30]:1

julia> size(c.repr(zero(c.GA)), 2)
3

julia> size(c.A, 2) * order(group(c.GA))
12
julia> size(c.A, 1) * order(group(c.GA)) 
9

julia> size(c.A, 2) * size(c.repr(zero(c.GA)), 2)
12
julia> size(c.A, 1) * size(c.repr(zero(c.GA)), 1)
9
```

Defined `code_k` for the classical code as $k = n - rank(H)$

PR #526, opened 2025-06-26
Parameterizing the phase type (not hardcoded to UInt8 anymore)

Hamiltonian-Action:
~~I am uncertain if I modified buildkite properly, since I am unaware where its starting directory lies. This can be solved through a recursive `grep` but this would be absolutely unacceptable if it starts at the system root directory.~~

All the pertinent type provided by `QuantumClifford` have been modified to no longer be restricted to storing their phase as a `UInt8`.

## CHECKLIST ##

### DIFF
--- a/src/QuantumClifford.jl
+++ b/src/QuantumClifford.jl
@@ -281,7 +281,6 @@ Base.hash(t::Tableau, h::UInt) = hash(t.nqubits, hash(t.phases, hash(t.xzs, h)))
 
 Base.copy(t::Tableau) = Tableau(copy(t.phases), t.nqubits, copy(t.xzs))
 
-# This is hideous but JET despises using eltype.
 function Base.zero(::Type{Tableau{PV, XZM}}, r, q) where {
     P <: Unsigned, PV <: AbstractVector{P},
     XZ <: Unsigned, XZM <: AbstractMatrix{XZ}
--- a/src/pauli_operator.jl
+++ b/src/pauli_operator.jl
@@ -164,7 +164,6 @@ end
 
 _nchunks(i::Int, ::Type{T}) where {T <: Unsigned} = 2 * ((i - 1) ÷ count_zeros(zero(T))) + 2
 
-# This is hideous but JET despises using eltype.
 function Base.zero(::Type{PauliOperator{PS, XZV}}, q) where {
     P <: Unsigned, PS <: AbstractArray{P, 0},
     XZ <: Unsigned, XZV <: AbstractVector{XZ}


.buildkite/pipeline.yml
Krastanov, line 1: Make a new run that does the UInt8 test, do not change all the existing tests (they are there to test the library in normal operation)

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 ## v0.10.0 - 2025-06-25
 
+- **(breaking?)** the phase(s) field of the various types defined within QuantumClifford has been altered from `UInt8` to a new internal type alias, which presently matches `UInt32`. Existing functions may accept or return a mixture of the two, to be further unified at some point in the future.
```
Krastanov: This is not breaking and it is an internal change. Mentioning it briefly at the bottom of the chaingelog makes sense, but it is not something deserving of a big mention.

src/QuantumClifford.jl
```diff
@@ -135,6 +135,9 @@ end
 
 const NoZeroQubit = ArgumentError("Qubit indices have to be larger than zero, but you are attempting to create a gate acting on a qubit with a non-positive index. Ensure indexing always starts from 1.")
 
+# TODO: PHASE-TYPE: Parametrise the phase eventually.
```
Krastanov: There are a lot of TODOs sprinkled throughout that would not be very clear on a reader just going through the code. It would be cleaner to remove them.

src/QuantumClifford.jl
```diff
@@ -163,23 +166,38 @@ include("pauli_operator.jl")
...
-    tab = zero(Tableau{Vector{UInt8},Matrix{T}},r,n)
+    tab = zero(Tableau{Vector{PhaseType},Matrix{T}},r,n)
     for i in eachindex(paulis)
         tab[i] = paulis[i]
     end
     tab
 end
 
+# CAUTION: These just pass things back to the existing UInt8 implementation.
+function Tableau(phases::AbstractVector{<: Unsigned}, xs::AbstractMatrix{Bool}, zs::AbstractMatrix{Bool})
```
Krastanov, line 186: I am not sure why this method exists. Who would be using it?

It seems kinda wrong? Why is the default constructor creating a `UInt8` array?

src/QuantumClifford.jl
```diff
@@ -278,12 +297,14 @@ Base.hash(t::Tableau, h::UInt) = hash(t.nqubits, hash(t.phases, hash(t.xzs, h)))
 
 Base.copy(t::Tableau) = Tableau(copy(t.phases), t.nqubits, copy(t.xzs))
 
+# TODO: PHASE-TYPE: This should be dynamic!
```
Krastanov: mentioned above too, but there isn't really a need to add this TODO at every location where `PhaseType` is used. One can just grep for `PhaseType`

src/QuantumClifford.jl
```diff
@@ -397,6 +418,17 @@ struct Stabilizer{T<:Tableau} <: AbstractStabilizer
     tab::T
 end
 
+# CAUTION: These just pass things back to the existing UInt8 implementation.
+function Stabilizer(phases::Tₚᵥ, nqubits::Int, xzs::Tₘ) where {Tₚᵥ <: AbstractVector{<: Unsigned}, Tₘ <: AbstractMatrix{<: Unsigned}}
```
Krastanov: same comment about not being sure why these exist -- the new default is UInt32, but these default constructors give a UInt8 type

src/QuantumClifford.jl
```diff
@@ -718,6 +751,7 @@ julia> prodphase(P"XXX", P"ZZZ")
 ```
 """
 @inline function prodphase(l::AbstractVector{T}, r::AbstractVector{T})::UInt8 where T<:Unsigned
+# TODO: PHASE-TYPE: Keep or change accordingly?
```
Krastanov: unless benchmarking says otherwise, these can be kept -- a compiler should be able to take care of this casting

src/ecc/decoder_pipeline.jl
```diff
@@ -251,7 +251,8 @@ function create_lookup_table(code::Stabilizer)
     for bit_to_be_flipped in 1:qubits
         for error_type in [single_x, single_y, single_z]
             # Generate e⃗
-            error = error_type(qubits, bit_to_be_flipped)::PauliOperator{Array{UInt8, 0}, Vector{UInt}}
+            # TODO: PHASE-TYPE: Parametrise the phase eventually.
+            error = error_type(qubits, bit_to_be_flipped)::PauliOperator{Array{PhaseType, 0}, Vector{UInt}}
```
Krastanov: check whether you can just erase this typeassert -- it is there to make JET happy

src/mul_leftright.jl
```diff
@@ -113,14 +113,16 @@ function mul_ordered!(r::AbstractVector{T}, l::AbstractVector{T}; phases::Val{B}
     rcnt1, rcnt2
 end
 
-function mul_left!(r::AbstractVector{T}, l::AbstractVector{T}; phases::Val{B}=Val(true))::UInt8 where {T<:Unsigned, B}
+# TODO: PHASE-TYPE: This should be dynamic!
+function mul_left!(r::AbstractVector{T}, l::AbstractVector{T}; phases::Val{B}=Val(true))::PhaseType where {T<:Unsigned, B}
     rcnt1, rcnt2 = mul_ordered!(r, l; phases=phases)
-    return UInt8((rcnt1 ⊻ (rcnt2<<1))&0x3)
+    return convert(PhaseType, (rcnt1 ⊻ (rcnt2<<1))&0x3)
```
Krastanov: is this convert necessary?

src/mul_leftright.jl
```diff
@@ -113,14 +113,16 @@ function mul_ordered!(r::AbstractVector{T}, l::AbstractVector{T}; phases::Val{B}
...
 end
 
-function mul_right!(l::AbstractVector{T}, r::AbstractVector{T}; phases::Val{B}=Val(true))::UInt8 where {T<:Unsigned, B}
+# TODO: PHASE-TYPE: This should be dynamic!
+function mul_right!(l::AbstractVector{T}, r::AbstractVector{T}; phases::Val{B}=Val(true))::PhaseType where {T<:Unsigned, B}
     rcnt1, rcnt2 = mul_ordered!(l, r; phases=phases)
-    return UInt8(((rcnt1 ⊻ (rcnt2<<1)) + rcnt1*2)&0x3) # TODO simplify
+    return convert(PhaseType, ((rcnt1 ⊻ (rcnt2<<1)) + rcnt1*2)&0x3) # TODO simplify
```
Krastanov: same, is this necessary?

src/pauli_operator.jl
```diff
@@ -39,12 +39,25 @@ julia> p[1] = (true, true); p
...
-struct PauliOperator{Tₚ<:AbstractArray{UInt8,0}, Tᵥ<:AbstractVector{<:Unsigned}} <: AbstractCliffordOperator
+struct PauliOperator{Tₚ<:AbstractArray{PhaseType,0}, Tᵥ<:AbstractVector{<:Unsigned}} <: AbstractCliffordOperator
+# TODO: PHASE-TYPE: Parametrise the phase eventually.
     phase::Tₚ
     nqubits::Int
     xz::Tᵥ
 end
 
+# CAUTION: For compatibility with existing usage. Eliminate eventually.
+function PauliOperator(phase::Tₚ, nqubits::Int, xz::Tᵥ) where {Tₚ <: AbstractArray{<: Unsigned, 0}, Tᵥ <: AbstractVector{<: Unsigned}}
```
Krastanov: I thought you said you do not plan to parameterize the phase type, rather to just hardcoded to UInt32?

src/project_trace_reset.jl
```diff
@@ -161,7 +162,7 @@ julia> state
 + Z__
 
 julia> anticom_index, result
-(0, 0x02)
+(0, 0x00000002)
```
Krastanov: check whether these can be fixed with a regex in the doctests instead

src/symbolic_cliffords.jl
```diff
@@ -132,7 +132,7 @@ Z₂ ⟼ - _X_
 Z₃ ⟼ + __Z
 
 julia> typeof(t_op)
-CliffordOperator{QuantumClifford.Tableau{Vector{UInt8}, Matrix{UInt64}}, PauliOperator{Array{UInt8, 0}, Vector{UInt64}}}
```
Krastanov: check whether these can be addressed with a regex in the doctest runner instead

.buildkite/pipeline.yml
Krastanov, line 1: Make a new run that does the UInt8 test, do not change all the existing tests (they are there to test the library in normal operation)

Krastanov:
I do not understand the point you are making about the new constructors. What is the use for the constructors that have `convert(UInt8, ...)` in them? What would be an example of a piece of code that a user would want to execute where such constructors are used?

Generally no need to respond to each comment separately, but the ones you believe you have addressed you can just mark as resolved. Feel free to just re-request a review (upper right side of the page) when you believe all issues are addressed.

Krastanov:
A segfault!? We do not do any non-Julia calls in this library and changing the phase type is unrelated to CUDA/KA, so a segfault definitely should be investigated (but of course, this is not something that you need to do, it is outside of the scope of your project.)

Do you have some MWE? Just a small script that when executed causes a segfault?

src/QuantumClifford.jl
```diff
@@ -237,13 +240,14 @@ function Base.setindex!(tab::Tableau, pauli::PauliOperator, i)
     tab
 end
 
+# TODO: Is this in any test file? This is incomplete and does not make sense.
```
Krastanov: Could you explain this TODO?

Keep changes in PRs related to a single topic. If there is a bug here, just report it as an issue.

src/sumtypes.jl
```diff
@@ -223,6 +223,7 @@ function concrete_typeparams(t::Type{NoiseOp})
     ]
 end
 
+# TODO: Slay this dragon or keep it around? What restrictions does it impose?
```
Krastanov: Leaving this type of TODOs is not the best way to keep track of things -- it gets lost and it will not be addressed. If this is something you would like to discuss, we can put it in an issue. Please remove this type of TODOs from your PRs.

The `sumtypes.jl` file is used to dynamically create sum type representation of all gates one can have in a circuit in order to be able to loop over (what would have been) heterogeneously typed arrays without boxing/allocations/slowdown.

src/pauli_operator.jl
```diff
@@ -138,9 +158,7 @@ Base.hash(p::PauliOperator, h::UInt) = hash(p.phase,hash(p.nqubits,hash(p.xz, h)
 Base.copy(p::PauliOperator) = PauliOperator(copy(p.phase),p.nqubits,copy(p.xz))
 
 function LinearAlgebra.inv(p::PauliOperator)
-  ph = p.phase[]
-  phin = xor((ph << 1) & ~(UInt8(1) << 2), ph)
-  return PauliOperator(phin, p.nqubits, copy(p.xz))
+    return PauliOperator(-p.phase, p.nqubits, copy(p.xz))
```
Krastanov: Is there a mod 4 missing here for the phase?

The old version looks pretty silly, so this is definitely an improvement.

src/pauli_operator.jl
```diff
@@ -39,21 +39,41 @@ julia> p[1] = (true, true); p
...
+    return PauliOperator(p, length(x), xz)
+end
+
+function PauliOperator(
+    phase::Unsigned, x::AbstractVector{Bool}, z::AbstractVector{Bool}
+    )
+
+    return PauliOperator(phase, BitVector(x), BitVector(z))
+
 end
```
Krastanov:
```suggestion
)
    return PauliOperator(phase, BitVector(x), BitVector(z))
end
```

Krastanov: removing some of the extra whitespace

src/pauli_operator.jl
```diff
@@ -149,15 +167,21 @@ function Base.deleteat!(p::PauliOperator, subset)
...
 
 _nchunks(i::Int,T::Type{<:Unsigned}) = 2*( (i-1) ÷ (8*sizeof(T)) + 1 )
-Base.zero(::Type{PauliOperator{Tₚ, Tᵥ}}, q) where {Tₚ,T<:Unsigned,Tᵥ<:AbstractVector{T}} = PauliOperator(zeros(UInt8), q, zeros(T, _nchunks(q,T)))
+
+function Base.zero(::Type{PauliOperator{P, XZ}}, q) where {P, XZ}
+    return PauliOperator(
+        zeros(eltype(P)), q, zeros(eltype(XZ), _nchunks(q, eltype(XZ)))
+        )
```
Krastanov:
```suggestion
    )
```

src/pauli_operator.jl
```diff
@@ -39,21 +39,41 @@ julia> p[1] = (true, true); p
...
     nqubits::Int
-    xz::Tᵥ
+    xz::XZ
 end
 
-PauliOperator(phase::UInt8, nqubits::Int, xz::Tᵥ) where Tᵥ<:AbstractVector{<:Unsigned} = PauliOperator(fill(UInt8(phase),()), nqubits, xz)
-function PauliOperator(phase::UInt8, x::BitVector, z::BitVector)
-    phase = fill(UInt8(phase),())
-    xs = reinterpret(UInt,x.chunks)::Vector{UInt}
-    zs = reinterpret(UInt,z.chunks)::Vector{UInt}
```
Krastanov, line 52: Double check that JET is not complaining when you remove type assertions. They are sprinkled throughout the codebase wherever JET has reported poor type inference in the past (some of them might not be needed in recent versions of Julia, which is a good enough reason to remove them).

src/pauli_operator.jl
```diff
@@ -39,21 +39,41 @@ julia> p[1] = (true, true); p
...
-    xzs = cat(xs, zs, dims=1)
-    PauliOperator(phase, length(x), xzs)
+function PauliOperator(
+    phase::Unsigned, nqubits::Int, xz::AbstractVector{<: Unsigned}
+    )
+
+    p = similar(xz, typeof(phase), ())
+    fill!(p, phase)
+    return PauliOperator(p, nqubits, xz)
+
```
Krastanov:
```suggestion
```

src/pauli_operator.jl
```diff
@@ -39,21 +39,41 @@ julia> p[1] = (true, true); p
...
-    phase = fill(UInt8(phase),())
-    xs = reinterpret(UInt,x.chunks)::Vector{UInt}
-    zs = reinterpret(UInt,z.chunks)::Vector{UInt}
-    xzs = cat(xs, zs, dims=1)
-    PauliOperator(phase, length(x), xzs)
+function PauliOperator(
+    phase::Unsigned, nqubits::Int, xz::AbstractVector{<: Unsigned}
+    )
+
+    p = similar(xz, typeof(phase), ())
```
Krastanov:
```suggestion
)
    p = similar(xz, typeof(phase), ())
```

src/pauli_operator.jl
```diff
@@ -39,21 +39,41 @@ julia> p[1] = (true, true); p
 + YYZ
 ```
 """
-struct PauliOperator{Tₚ<:AbstractArray{UInt8,0}, Tᵥ<:AbstractVector{<:Unsigned}} <: AbstractCliffordOperator
-    phase::Tₚ
+struct PauliOperator{
+    P <: AbstractArray{<: Unsigned, 0}, XZ <: AbstractVector{<: Unsigned}
+    } <: AbstractCliffordOperator
```
Krastanov:
```suggestion
} <: AbstractCliffordOperator
```

src/misc_ops.jl
```diff
@@ -6,8 +6,10 @@ import QuantumInterface: nsubsystems
 Particularly useful when acting on [`Register`](@ref).
 
 See also: [`apply!`](@ref), [`projectrand!`](@ref)."""
-struct PauliMeasurement{Tₚ<:AbstractArray{UInt8,0}, Tᵥ<:AbstractVector{<:Unsigned}} <: AbstractMeasurement
-    pauli::PauliOperator{Tₚ,Tᵥ}
+struct PauliMeasurement{
+    P <: AbstractArray{<: Unsigned, 0}, XZ <: AbstractVector{<: Unsigned}
+    } <: AbstractMeasurement
```
Krastanov:
```suggestion
} <: AbstractMeasurement
```

src/QuantumClifford.jl
```diff
@@ -278,20 +282,20 @@ Base.hash(t::Tableau, h::UInt) = hash(t.nqubits, hash(t.phases, hash(t.xzs, h)))
...
 Base.copy(t::Tableau) = Tableau(copy(t.phases), t.nqubits, copy(t.xzs))
 
-function Base.zero(::Type{Tableau{Tₚᵥ, Tₘ}}, r, q) where {Tₚᵥ,T<:Unsigned,Tₘ<:AbstractMatrix{T}}
-    phases = zeros(UInt8,r)
-    xzs = zeros(UInt, _nchunks(q,T), r)
-    Tableau(phases, q, xzs)::Tableau{Vector{UInt8},Matrix{UInt}}
+function Base.zero(::Type{Tableau{P, XZ}}, r, q) where {P, XZ}
+    return Tableau(
+        zeros(eltype(P), r), q, zeros(eltype(XZ), _nchunks(q, eltype(XZ)), r)
+        )
```
```suggestion
    )
```

src/QuantumClifford.jl
Krastanov: seems to be covered by tests https://app.codecov.io/gh/QuantumSavory/QuantumClifford.jl/blob/master/src%2FQuantumClifford.jl#L240

src/QuantumClifford.jl
```diff
@@ -163,24 +163,27 @@ include("pauli_operator.jl")
 """Internal Tableau type for storing a list of Pauli operators in a compact form.
 No special semantic meaning is attached to this type, it is just a convenient way to store a list of Pauli operators.
 E.g. it is not used to represent a stabilizer state, or a stabilizer group, or a Clifford circuit."""
-struct Tableau{Tₚᵥ<:AbstractVector{UInt8}, Tₘ<:AbstractMatrix{<:Unsigned}}
-    phases::Tₚᵥ
+struct Tableau{
+    P <: AbstractVector{<: Unsigned}, XZ <: AbstractMatrix{<: Unsigned}
+    }
```
```suggestion
}
```

src/QuantumClifford.jl
```diff
@@ -278,20 +282,20 @@ Base.hash(t::Tableau, h::UInt) = hash(t.nqubits, hash(t.phases, hash(t.xzs, h)))
...
 
 """Zero-out a given row of a [`Tableau`](@ref)"""
-@inline function zero!(t::Tableau,i)
-    t.xzs[:,i] .= zero(eltype(t.xzs))
-    t.phases[i] = 0x0
-    t
+@inline function zero!(t::Tableau{P, XZ}, i) where {P, XZ}
+    fill!((@view t.phases[i]), zero(eltype(P)))
+    fill!((@view t.xzs[:, i]), zero(eltype(XZ)))
+    return t
```
Krastanov, line 298: Any reason to switch to `fill!` instead of the broadcast?

Krastanov:
I modified the PR title a bit. I am enjoying the flowery language, but when this pops up in the autogenerated release notes on github it might cause confusion to external readers.

src/QuantumClifford.jl
Hamiltonian-Action: I cannot inspect the resulting code through the Julia provided macros since they are unable to handle broadcasting, so take this with a grain of salt. Any sane implementation will have `fill!` lower down to `memcpy`, but broadcasting might be handled by different logic that results in a slower code path.

Krastanov:
There seems to be a small performance degradation on Julia 1.9 due to these changes. Some operations now allocate more than they used to before this PR, when running on Julia 1.9. This is why the `Downgrade` test is failing. Might be because of the removed typeasserts, might be because of other of these changes.

For now, let's just drop support for 1.9. But in future PRs, keep the seemingly-redundant typeasserts, just to avoid this type of risk.

src/QuantumClifford.jl
Hamiltonian-Action: It would make sense if it read as `tab.phases[i] = t.phases[i]; tab.xzs[:, i] = t.xzs[:, i]` but right now it seems to be doing some assignment of `UInt8 = Array{UInt8}` -- assuming phases are UInt8 -- which is not really legal???

src/pauli_operator.jl
Krastanov: Marking as resolved because we are just dropping support for 1.9 anyway. As mentioned elsewhere, keep this type of annotations in future PRs to avoid surprises with typeinference on older julia versions (the reason they were put in to begin with)

src/QuantumClifford.jl
Krastanov: If `i` is an integer this indeed throws a ~~warning~~ error. This is for slicing where i is a list or a range.

We can have a nicer error message for the `i::Integer` case, but there is no way to be properly exhaustive about this in Julia. We can restrict `i` to be `Base.AbstractVecOrTuple` and that would probably cover most realistic cases.

src/pauli_operator.jl
Hamiltonian-Action: The old code was not restricted to output in the range [0, 3] so I just assumed any values are fine provided they are equivalent modulo 4, will add `.& 0x3`

Krastanov: The old version would return [0,3] if the input was in [0,3].

Krastanov:
Buildkite is slow as it is going through a bit of a backlog from the last day or so.

If buildkite passes (except for quantumsavory), this should be good to merge after the two TODOs are fixed.

src/sumtypes.jl
Hamiltonian-Action: The question I have is whether this will introduce any restriction on the to-be-added GPU capabilities given that the nomenclature `concrete_typeparams` seem to suggest that this forces a dependency on the `UInt8, UInt64` pairing. Since everything is now parametric, this is not necessarily the case.

src/QuantumClifford.jl
Hamiltonian-Action: ~~How about `i <: Integer && i = collect(i)`? This will now work for both integral and general collections.~~ Who would realistically construct a single row tableau, let alone perform that assignment. Will proceed with your suggestion to communicate the intention behind the code.

Edit: Better yet, just disregard this and stash it away for some systematic review of how these things are done. I've already noted that this will error out on device arrays anyway, so it should eventually be due for an overhaul.

src/sumtypes.jl
Krastanov: This is necessary to do the following operation efficiently (on a GPU or not):

```
for g in [sX(1), sCNOT(2,3), sXCX(4,5), sMZ(5), PauliMeasurement(...) ...]
    apply!(some_gpu_pauli_frame, g)
end
```

Some updates to it might be useful (basically just adding a new entry with GPU arrays), but no need to worry about that, it is an edge case of an edge case. It will not be giving silently wrong answers, just some specific optimizations for PauliFrames will start raising warnings. We can address those when it happens.

src/pauli_operator.jl
Hamiltonian-Action: Mostly there to stick to your previous suggestion of avoiding any extraneous assertions that are not necessary for dispatch and just letting the lexer/parser/~~gandalf~~compiler handle it on its own.

Krastanov: Fair point. If you find this type of extra assertions in the current codebase, almost certainly they are there to address specific failure of type inference, usually as detected by JET or by noticing unreasonable number of allocations.

There are some allocation tests in CI to verify that such allocations are not happening. There are some more modern tools too that probably would be more to your liking, like AllocCheck.jl, but I have not employed them yet in this codebase.

Krastanov:
Thanks! The `iscss` is indeed unrelated. It is not really JET that is unhappy with `eltype`, it is the compiler's type inference that fails to prove the function is type stable. If you can make a minimal example, I am sure the JET devs, the compiler folks, and all the performance zealots would be interested in having it tracked on the issue tracker. Discussing this on the JET issue tracker or the #internals slack channel might gain traction.

src/QuantumClifford.jl
Hamiltonian-Action: I did some testing between `s .= zero(eltype(s))` and `fill!(s, zero(eltype(s)))`. There is some minute difference ($\Delta = 75 \pm 50 ~ ns$) for small array sizes but that is realistically consistent with a null-hypothesis, for large arrays there is no detectable difference at the resolution of the benchmarking being performed. In any case, `fill!` is clearly the specialised tool that is designated for this task -- unlike broadcasting, it always reports zero additional allocations being performed -- and we can safely use it without having to worry about broadcasting issues and dropping 0-d arrays to scalars.

Krastanov:
go ahead and make the change, an extra CI run is fine

Krastanov:
Welp, that is a fun failure mode. I guess we need to go back to the uglier version. Could you undo the last commit?

You probably have already seen, but the failure is due to the fact that the correct operation when getting a zero of a subarray is to NOT return a subarray.

Krastanov:
Thanks!

PR #528, opened 2025-06-28
tests: more tests for mip distance solver

Fe-r-oz:
This PR aims to do a bit of additional stress testing on the mip distance runner in #439 by testing the true minimum distance. All codes parameters are taken from [here](https://github.com/QEC-pages/GB-codes). 

Only those codes that takes less than 1 minutes have been tested to less the amount of extra time CI will consume. The solver may approximates d0 (true minimum distance)  as both strict lower bound and sometimes upper bound. We may document this approximation quirk in a sentence where upper bound parameter docstring is present.

### DIFF
--- a/test/test_ecc_generalized_bicycle_codes.jl
+++ b/test/test_ecc_generalized_bicycle_codes.jl
@@ -28,4 +28,89 @@
         @test code_n(c) == 72 && code_k(c) == 10
         @test distance(c, DistanceMIPAlgorithm(solver=HiGHS, logical_qubit=1)) == 9
     end
+
+    @testset "Distance bounds for generalized bicycle codes"
+        # codes taken from https://github.com/QEC-pages/GB-codes
+        c = generalized_bicycle_codes([0, 2], [0, 1], 5)
+        @test code_n(c) == 10
+        @test distance(c, DistanceMIPAlgorithm(solver=HiGHS)) == 3
+        c = generalized_bicycle_codes([0, 3], [0, 1], 11)
+        @test code_n(c) == 22
+        @test distance(c, DistanceMIPAlgorithm(solver=HiGHS)) == 4
+        c = generalized_bicycle_codes([0, 5], [0, 1], 13)
+        @test code_n(c) == 26
+        @test distance(c, DistanceMIPAlgorithm(solver=HiGHS)) == 5
+        c = generalized_bicycle_codes([0, 4], [0, 1], 19)
+        @test code_n(c) == 38
+        @test distance(c, DistanceMIPAlgorithm(solver=HiGHS)) == 5
+        c = generalized_bicycle_codes([0, 8], [0, 1], 29)
+        @test code_n(c) == 58
+        @test distance(c, DistanceMIPAlgorithm(solver=HiGHS)) == 7
+        c = generalized_bicycle_codes([0, 8], [0, 1], 37)
+        @test code_n(c) == 74
+        @test distance(c, DistanceMIPAlgorithm(solver=HiGHS)) == 8
+        c = generalized_bicycle_codes([0, 8], [0, 1], 53)
+        @test code_n(c) == 106
+        @test distance(c, DistanceMIPAlgorithm(solver=HiGHS)) == 9
+
+        c = generalized_bicycle_codes([0, 1,  2,  3], [0,1], 5)
+        @test code_n(c) == 10
+        @test distance(c, DistanceMIPAlgorithm(solver=HiGHS)) == 3
+        c = generalized_bicycle_codes([0, 1,  2,  5], [0,1], 11)
+        @test code_n(c) == 22
+        @test distance(c, DistanceMIPAlgorithm(solver=HiGHS)) == 5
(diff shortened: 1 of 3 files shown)


test/test_ecc_generalized_bicycle_codes.jl
```diff
@@ -28,4 +28,89 @@
         @test code_n(c) == 72 && code_k(c) == 10
         @test distance(c, DistanceMIPAlgorithm(solver=HiGHS, logical_qubit=1)) == 9
     end
+
+    @testset "generalized bicycle codes from arXiv:2203.17216" begin
```
Krastanov: this here can just go as a new testitem in a file `test_slow_ecc_generalized_bicycle...` and be tagged as `:slow`

Krastanov (changes requested):
Thanks, these are very valuable tests.

We have started adding a lot of relatively long tests lately. Even when taking a single minute, these minutes start adding up. Let's start structuring these a bit differently.

We already can filter by tags in `runtests.jl`. Let's start marking any tests that take longer than a handful of seconds as "slow" in its tags. Then we can add to the if-else in runtests.jl so that in the presence of `ENV["SLOW_TESTS"]` we run **only** the tests marked as slow. That way we do not need to worry about them slowing down the test runner.

Krastanov:
forgot to mention, once this slow-tests filter is added, we can have a buildkite step added that runs only the slow tests by setting the necessary env variable

Fe-r-oz:
Thank you for your feedback!

Just a small tangent: One useful thing when attempting this PR is that we require some specialized and dedicated algorithms for calculating upper bounds and lower bounds of codes, not just the true minimum distance. There is a recent [C library](https://github.com/QEC-pages/dist-m4ri) to calculate to minimum distance which has algorithms to calculate both the upper bound and lower bound of quantum/classical codes. It could be nice to have a wrapper if that makes sense for this to check whether it works for thousands of qubits as we will eventually require that capability.  I don't know the bridge between C to Julia and vice versa.

Krastanov:
Anything done by Leonid P. Pryadko would be a useful piece of code. Usually for C code in Julia one would use BinaryBuilder to build a "shared library" or "executable". The preferred option is usually "shared library", but it seems like this particular codebase is not really designed to be used as a library. I guess then the build target would be an executable and one would need to prepare temporary files and give them to the executable as input.

Check `BinaryBuilder`, `Base.run`, and whatever tools julia provides for creating temporary files.

src/ecc/ECC.jl
```diff
@@ -109,6 +109,7 @@ end
 
 code_k(c::AbstractECC) = code_k(parity_checks(c))
 
+# TODO Use dedicated algorithms such as random information set to calculate the upper distance bound.
```
Krastanov: this todo is a bit out of place, maybe delete it

test/runtests.jl
```diff
@@ -13,6 +13,14 @@ else
...
 end
 
+if get(ENV, "SLOW_TESTS", "") == "true"
+    @info "Running only slow tests"
+    test_args = ["--tags=slow"]
+else
+    @info "Skipping slow tests (set SLOW_TESTS=true to include them)"
+    test_args = ["--tags=!slow"]
+end
+
```
Krastanov: This would not work (has a bit of a flavor of something an LLM will generate -- if you use those (they can be helpful), be careful to double check their work -- in this case this is actually a bit nonsensical as `test_args` and `--tags=` are not really anything we actually use). Check how the JET tag is done below and use the same structure.

test/test_ecc_generalized_bicycle_codes_slow.jl
```diff
@@ -0,0 +1,94 @@
+@testitem  "ECC GB Codes - slow" tags=[:slow] begin
+    using Hecke
+    using HiGHS
+    using JuMP
+    using QuantumClifford.ECC.QECCore: code_k, code_n, distance, rate
+    using QuantumClifford.ECC: generalized_bicycle_codes, code_k, code_n, DistanceMIPAlgorithm
+
+    @testset "generalized bicycle codes from arXiv:2203.17216" begin
+        # codes taken from https://github.com/QEC-pages/GB-codes
+        # TODO investigate wrapping https://github.com/QEC-pages/dist-m4ri for supporting dedicated
```
Krastanov: This TODO is a bit out of place here -- after all this is testing capabilities unrelated to this other algorithm.

Krastanov: maybe post an issue with this so we can keep track of it

Krastanov (changes requested):

Fe-r-oz:
> The preferred option is usually "shared library", but it seems like this particular codebase is not really designed to be used as a library. I guess then the build target would be an executable and one would need to prepare temporary files and give them to the executable as input.

> Check BinaryBuilder, Base.run, and whatever tools julia provides for creating temporary files.

Sounds cool! Thank you  for the explanation! I'm not familiar with this yet, but it sounds like a really interesting endeavour!

Fe-r-oz:
Thank you very much for your feedback! I followed the JET-style for the slow tests so that only slow tests run with :slow is tagged!

PR #530, opened 2025-06-29
remove unreliable `upper_bound` from mip distance solver

Fe-r-oz:
This PR aims to removes the `upper_bound` parameter from the MIP distance solver because it conflicts with the algorithm’s design. The solver introduced in [arXiv:1108.5738](https://arxiv.org/abs/1108.5738) implements an exact minimum distance calculation formulated as the binary integer program:

```math
\begin{aligned}
&\min \sum_v x_v \
&\text{s.t. } \bigoplus_{v\in f} x_v = s_f \quad \forall f \
&x_v \in {0,1}
\end{aligned}
```

Attempting to repurpose it for upper bounds computation introduced unreliable heuristics. Empirical testing revealed the `upper_bound` mode produced inconsistent results. This change restores the solver’s single purpose:  compute the true minimum distances.

TODO: For rigorous upper bounds computation, dedicated algorithms such as random information set in [dist-m4ri](https://github.com/QEC-pages/dist-m4ri) uses probabilistic methods better suited for approximations. Future work will expose proper upper-bound interfaces by investigatiing dedicated algorithms for computing upper bound.

This PR was inspired by #528.

### DIFF
--- a/ext/QuantumCliffordJuMPExt/min_distance_mixed_integer_programming.jl
+++ b/ext/QuantumCliffordJuMPExt/min_distance_mixed_integer_programming.jl
@@ -190,20 +190,11 @@ function distance(code::AbstractECC, alg::DistanceMIPAlgorithm)
         )
     end
     # Return appropriate result based on configuration
-    if alg.upper_bound
-        if alg.opt_summary
-            max_entry = argmax(x -> x.weight, values(weights))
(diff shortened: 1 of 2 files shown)


Krastanov:
thanks for the cleanup. Indeed, that upper bound branch does not seem to be useful for anything -- there is no point to return the maximum of this list if we already have access to the minimum.

PR #533, opened 2025-06-29
add: `GraphState` type

LEXUGE:
This PR targets the following project items:
1. [Type for graph state](https://github.com/orgs/QuantumSavory/projects/12?pane=issue&itemId=109344961). I plan to implement the graph backend parameterization in the next PR involving interface.
2. [Conversion between graph states and Stabilizer](https://github.com/orgs/QuantumSavory/projects/12?pane=issue&itemId=109345027)

It gives
- a cleaner API on graph states through `GraphState` type
- updated tests and docs accordingly

Help/comment is appreciated on documentation as it's my first time writing/changing them. Also I wonder why the `doctest` is disabled?

### DIFF
--- a/src/QuantumClifford.jl
+++ b/src/QuantumClifford.jl
@@ -1418,7 +1418,7 @@ include("useful_states.jl")
 #
 include("experimental/Experimental.jl")
 #
-include("graphs.jl")
+include("./graphs/graphs.jl")
 #
 include("entanglement.jl")
 #
--- a/src/graphs/graphs.jl
+++ b/src/graphs/graphs.jl
@@ -1,13 +1,88 @@
-import Graphs
+import Graphs: Graphs, nv, AbstractGraph
+
+const clifford_id1 = one(CliffordOperator, 1)
+
+# elided as it's a non-public API and might introduce unexpected behavior for downstream users
+# we didn't wrap it inside a new struct as that introduces boiler plate for show etc.
+function _apply_vop!(vops::Vector{SingleQubitOperator}, op::AbstractSingleQubitOperator; phases::Bool=true)
+    q = affectedqubits(op)[1]
+    # Convert into a single qubit operator acting on index 1
+    op_prime = SingleQubitOperator(SingleQubitOperator(op), 1)
+    # A trick that performs slightly better than directly op_prime * CliffordOperator(vops[q])
+    # as converting to `CliffordOperator` is expensive
+    vops[q] = SingleQubitOperator(op_prime * (vops[q] * clifford_id1))
+end
+
+# NOTE: Not tested yet, so commented out for now
+# function _apply_vop_right!(vops::Vector{SingleQubitOperator}, op::AbstractSingleQubitOperator; phases::Bool=true)
+#     q = affectedqubits(op)[1]
+#     # Convert into a single qubit operator acting on index 1
+#     op_prime = SingleQubitOperator(SingleQubitOperator(op), 1)
+#     vops[q] = SingleQubitOperator(vops[q] * (op_prime * clifford_id1))
+# end
+
+# TODO: Implement sanity_check like the one in Stim
+struct GraphState{G, S}
+    # Graph representing the graph state
+    graph::G
+    # Vertex Operators that need to be applied to a stabilizer state
+    # before the state is representable as a "pure" graph state
+    vops::Vector{S}
+end
+
+nqubits(g::GraphState) = nv(g.graph)
+
+"""Return the underlying graph of the graph state"""
+graph(g::GraphState) = g.graph
+
+"""Return the VOPs (Vertex Operators) of the graph state"""
+vops(g::GraphState) = g.vops
+
+function GraphState(graph::G, h_idx::Vector{Int}, ip_idx::Vector{Int}, z_idx::Vector{Int}) where G
+    vops = [SingleQubitOperator(sId1(1)) for _ in 1:nv(graph)]
+    for id in h_idx
+        _apply_vop!(vops, sHadamard(id))
+    end
+    for id in ip_idx
+        _apply_vop!(vops, sInvPhase(id))
+    end
+    for id in z_idx
+        _apply_vop!(vops, sZ(id))
+    end
+    GraphState{G, SingleQubitOperator}(graph, vops)
+end
+
+function GraphState(stab::Stabilizer)
+    GraphState(graphstate(stab)...)
+end
+
+function Stabilizer(self::GraphState{<:AbstractGraph})
+    s_raw = Stabilizer(self.graph)
+    for id in eachindex(self.vops)
+        apply!(s_raw, SingleQubitOperator(inv(self.vops[id]), id))
+    end
+    s_raw
+end
 
 """An in-place version of [`graphstate`](@ref)."""
 function graphstate!(stab::Stabilizer)
     n = nqubits(stab)
-    stab, r, s, permx, permz = canonicalize_gott!(stab)
+    # canonicalization is in-place
+    _, r, _, permx, permz = canonicalize_gott!(stab)
+    # This is equivalent to saying perm(i) = (permz ∘ permx)(i)
+    # No typo above. permz ∘ permx because in Gottesman canonicalization
+    # we first do Gaussian elimination on the X half and then on the Z half
+    # when writing tableau in the matrix format.
     perm = permx[permz]
-    h_idx = [perm[i] for i in (r+1):n] # Qubits in which X ↔ Z is needed
-    ip_idx = [perm[i] for i in 1:n if stab[i,i]==(true,true)] # Qubits for which Y → X is needed
+
+    # Swap X ↔ Z in the (r+1):n columns so we have only X,Y,I diagonal and only Z off-diagonal
+    h_idx = [perm[i] for i in (r+1):n]
+    # Swap Y → X on the diagonal so we have only X, I on the diagonal
+    ip_idx = [perm[i] for i in 1:n if stab[i,i]==(true,true)]
+    # Since ZXZ = -X, apply Z on rows with negative phase to make all phases positive.
     phase_flips = [perm[i] for i in 1:n if phases(stab)[i]!=0x0]
+
+    # canonicalized stabilizer tableau serves as an adjacency matrix
     graph = Graphs.SimpleGraphFromIterator((
         Graphs.SimpleEdge(perm[i],perm[j])
         for i in 1:n, j in 1:n
--- a/test/test_graphs.jl
+++ b/test/test_graphs.jl
@@ -1,11 +1,36 @@
 @testitem "Graph states" begin
     using Graphs
     using Random
+    import QuantumClifford: GraphState
 
     test_sizes = [1,2,10,63,64,65,127,128,129] # Including sizes that would test off-by-one errors in the bit encoding.
 
-    for n in [1,test_sizes...]
-        s = random_stabilizer(n)
+    @testset "Graph State conversion (New API)" begin
+        for n in test_sizes
+            s = random_stabilizer(n)
+            @test canonicalize!(s) == canonicalize!(Stabilizer(GraphState(s)))
+        end
+    end
+
+    @testset "Graph State conversion (Old API)" begin
+        for n in test_sizes
+            s = random_stabilizer(n)
+            g, h_idx, ip_idx, z_idx = graphstate(s)
+            gates = graph_gatesequence(h_idx, ip_idx, z_idx)
+            gate = graph_gate(h_idx, ip_idx, z_idx, nqubits(s))
+            c0 = one(CliffordOperator,nqubits(s))
+            for gate in vcat(gates...) apply!(Stabilizer(tab(c0)), gate) end # TODO this wrapping in a Stabilizer hack is ugly; clean it up
+            @test c0==gate
+            s1 = copy(s)
+            for gate in vcat(gates...) apply!(s1, gate) end
+            @test canonicalize!(apply!(copy(s),c0)) == canonicalize!(s1) == canonicalize!(Stabilizer(g))
+        end
+        # one more check, done manually, to ensure we are testing for cases
+        # where canonicalize_gott! is causing index permutation
+        # (a few of the random cases already cover that)
+        s = S"- _XZ
+            + _ZX
+            + ZXZ"
         g, h_idx, ip_idx, z_idx = graphstate(s)
         gates = graph_gatesequence(h_idx, ip_idx, z_idx)
         gate = graph_gate(h_idx, ip_idx, z_idx, nqubits(s))
@@ -16,19 +41,4 @@
         for gate in vcat(gates...) apply!(s1, gate) end
         @test canonicalize!(apply!(copy(s),c0)) == canonicalize!(s1) == canonicalize!(Stabilizer(g))
     end
-    # one more check, done manually, to ensure we are testing for cases
-    # where canonicalize_gott! is causing index permutation
-    # (a few of the random cases already cover that)
-    s = S"- _XZ
-          + _ZX
-          + ZXZ"
-    g, h_idx, ip_idx, z_idx = graphstate(s)
-    gates = graph_gatesequence(h_idx, ip_idx, z_idx)
-    gate = graph_gate(h_idx, ip_idx, z_idx, nqubits(s))
-    c0 = one(CliffordOperator,nqubits(s))
-    for gate in vcat(gates...) apply!(Stabilizer(tab(c0)), gate) end # TODO this wrapping in a Stabilizer hack is ugly; clean it up
-    @test c0==gate
-    s1 = copy(s)
-    for gate in vcat(gates...) apply!(s1, gate) end
-    @test canonicalize!(apply!(copy(s),c0)) == canonicalize!(s1) == canonicalize!(Stabilizer(g))
 end
--- a/src/symbolic_cliffords.jl
+++ b/src/symbolic_cliffords.jl
@@ -201,6 +201,9 @@ SingleQubitOperator(p::sInvSQRTX)           = SingleQubitOperator(p.q, true , fa
 SingleQubitOperator(p::sSQRTY)              = SingleQubitOperator(p.q, false, true , true , false, true , false)
 SingleQubitOperator(p::sInvSQRTY)           = SingleQubitOperator(p.q, false, true , true , false, false, true)
 SingleQubitOperator(o::SingleQubitOperator) = o
+function SingleQubitOperator(o::SingleQubitOperator, qubit::Int)
+    return SingleQubitOperator(qubit, o.xx, o.xz, o.zx, o.zz, o.px, o.pz)
+end
 function SingleQubitOperator(op::CliffordOperator, qubit)
     nqubits(op)==1 || throw(DimensionMismatch("You are trying to convert a multiqubit `CliffordOperator` into a symbolic `SingleQubitOperator`."))
     SingleQubitOperator(qubit,tab(op)[1,1]...,tab(op)[2,1]...,(~).(iszero.(tab(op).phases))...)


docs/src/graphs.md
```diff
@@ -12,17 +11,17 @@ using QuantumClifford # hide
 ghz(4)
 ```
 
-It can be converted to a graph state with [`graphstate`](@ref)
+It can be converted to a graph state with [`GraphState`](@ref)
 
 ```julia
-graphstate(ghz(4))[1]
+GraphState(ghz(4)).graph
```
Krastanov: Struct fields are usually considered private. This should probably read `Graph(GraphState(...))`. But a few comments below might also change this consideration

docs/src/graphs.md
```diff
@@ -12,17 +11,17 @@ using QuantumClifford # hide
...
 ```
 
 ```@eval
 using Random; Random.seed!(1); using QuantumClifford, GraphMakie, CairoMakie;
 f = Figure(size=(200,200))
 a = Axis(f[1,1])
-graphplot!(a,graphstate(ghz(4))[1])
+graphplot!(a,GraphState(ghz(4)).graph)
```
Krastanov: struct fields are private

docs/src/graphs.md
```diff
@@ -41,14 +40,14 @@ julia> ghz(4)
 + _ZZ_
 + __ZZ
 
-julia> Stabilizer(Graph(ghz(4)))
+julia> Stabilizer(GraphState(ghz(4)).graph)
```
Krastanov: struct fields are private. Also, we should probably just have a method defined for Stabilizer(::GraphState) and for MixedDestabilizer(::GraphState)

docs/src/graphs.md
```diff
@@ -41,14 +40,14 @@ julia> ghz(4)
...
 + XZZZ
 + ZX__
 + Z_X_
 + Z__X
 ```
 
-There is a set of single-qubit operations that can convert any stabilizer tableau into a state representable as a graph. These transformations are performed implicitly by the `Graph` constructor when converting from a `Stabilizer`. If you need the explicit transformation you can use the [`graphstate`](@ref) function that specifies which qubits require a Hadamard, Inverse Phase, or Phase Flip gate. The [`graph_gatesequence`](@ref) or [`graph_gate`](@ref) helper functions can be used to generate the exact operations:
+There is a set of single-qubit operations that can convert any stabilizer tableau into a state representable as a graph. These transformations are performed implicitly by the `Graph` constructor when converting from a `Stabilizer`. If you need the explicit transformation you can use the `h_idx, ip_idx, z_idx` fields of [`GraphState`](@ref) struct that specifies which qubits require a Hadamard, Inverse Phase, or Phase Flip gate. The [`graph_gate_sequence`](@ref) or [`graph_gate`](@ref) helper functions can be used to generate the exact operations:
```
Krastanov: struct fields are private

docs/src/graphs.md
```diff
@@ -69,7 +68,7 @@ Z₂ ⟼ + _X__
 Z₃ ⟼ + __X_
 Z₄ ⟼ + ___X
 
-julia> canonicalize!(apply!(s,gate)) == canonicalize!(Stabilizer(g))
+julia> canonicalize!(apply!(s,gate)) == canonicalize!(Stabilizer(g.graph))
```
Krastanov: struct fields are private; we should have a constructor

src/QuantumClifford.jl
```diff
@@ -80,7 +80,7 @@ export
     bell, ghz, maximally_mixed,
     single_z, single_x, single_y,
     # Graphs
-    graphstate, graphstate!, graph_gatesequence, graph_gate,
+    GraphState, GraphState!, graph_gate_sequence, graph_gate,
```
Krastanov: `GraphState!` looks a bit weird... `!` means that it modifies in place, but CamelCase means it is a constructor.

I would suggest keeping the already existing functions and adding GraphState as a new constructor

src/graphs.jl
```diff
@@ -1,43 +1,89 @@
...
+"""
+[Graph states](https://en.wikipedia.org/wiki/Graph_state) are a special type
+of entangled stabilizer states that can be represented by a graph.
+For a graph ``G=(V,E)`` the corresponding stabilizers are ``S_v = X_v \\prod_{u ∈ neighbors(v)} Z_u``.
+Notice that such tableau are symmetric and contain only X on diagonal and Z off-diagonal.
+There is a set of single qubit gates that converts any stabilizer state to a graph state.
+"""
+# TODO: implement a graph backend interface so this constructor is generic over graph backend
+mutable struct GraphState
+    graph::AbstractGraph # underlying graph
```
Krastanov: structs should be concrete unless there is a good reason for them not to. This should be parameterized by the type of the graph

src/graphs.jl
```diff
@@ -1,43 +1,89 @@
...
+    # To transform from stabilizer state to graph state, we apply VOP in the orders of H, InvPhase, Z
+    # NOTE: Hadamard gate sequence may be swapped with InvPhase sequences. But Z gate sequence shall not be swapped with the previous two.
+    h_idx::Vector{Int} # indices for Hadamard gate to transform a given stabilizer state into graph state
+    ip_idx::Vector{Int} # indices for InvPhase gate
+    z_idx::Vector{Int} # indices for Z gate
+end
```
Krastanov: What is this struct going to be used for? It does not seem like the most useful way to manipulate a graph state. Usually one has the graph and then the local corrections at each node. Here, the local corrections are stored as a list of gates that would be very slow to modify as operations are performed.

src/graphs.jl
```diff
@@ -1,43 +1,89 @@
...
-`iphase_idx` are (different) qubits that require an inverse Phase gate (Y → X),
-and `flips_idx` are the qubits that require a phase flip (Pauli Z gate),
-after the previous two sets of gates.
+
+See also: [`graph_gate_sequence`](@ref)
+"""
+function graph_gate_sequence(self::GraphState)
+    vcat([sHadamard(i)::AbstractSingleQubitOperator for i in self.h_idx],
+         [sInvPhase(i)::AbstractSingleQubitOperator for i in self.ip_idx],
+         [sZ(i)::AbstractSingleQubitOperator for i in self.z_idx])
```
Krastanov: Any reason to do these typeasserts?

src/QuantumClifford.jl
Krastanov: The renaming of `graph_gate_sequence` introduces a breaking change with little benefit. Let's avoid those.

Krastanov (changes requested):
Thanks for the initial suggestions!

In terms of breaking changes, let's not do any renaming changes until everything else is ready. I would prefer if we undo all the changes to existing functions and just introduce the new capabilities. That would also lead to much less invasive changes to the current code and less review being needed.

I am not sure what is the plan for the newly introduced `GraphState` -- it does not seem to be convenient for efficient simulation. If it is just to encapsulate the existing functionality, let's leave this until the end. Currently we should focus on capabilities that actually permit modeling with the graph state method.

Krastanov:
I doubt that the current VOP would have anything to do with the API that you need. Just disregard them and make an independent structure to track them. A list of gates would be very inconvenient and expensive way to track evolution (you need something that has O(1) indexing wrt to qubit index, while this list is O(n) indexing as you need to search through it)

LEXUGE:
I agree VOPs should be indexed by qubit. But in any case still I think I would need a structure like `GraphState` to check it?

Are you suggesting to make it completely independent so that `GraphState` has to be constructed using the graph and `h_idx` etc returned by existing API?

If this is the understanding, then I think the current plan will be to implement `apply!` and `project!` etc on this `GraphState` and finalize API design at the very end. Presumably this will also push the interface related issue down the line.

LEXUGE:
I have updated the changes. Now,
1. existing APIs are kept
2. new struct in place with O(1) qubit indexing (just a vector)
3. put up an struct for `SingleQubitCliffordGroup`. Currently implemented with `CliffordOperator` but could use more efficient implementation (e.g. look-up table) later.
4. more comments and tests

LEXUGE:
Before
```
julia> @btime sZ * SingleQubitCliffordGroup(sZ)
  1.861 μs (137 allocations: 4.81 KiB)
X₁ ⟼ + X
Z₁ ⟼ + Z
```
After
```
julia> @btime sZ(1) * SingleQubitCliffordGroup(sZ(1))
  637.274 ns (28 allocations: 944 bytes)
SingleQubitOperator on qubit 1
X₁ ⟼ + X
Z₁ ⟼ + Z
```

Krastanov:
Thanks! Please tag me / re-request review when those changes are made/

src/QuantumClifford.jl
```diff
@@ -80,7 +80,7 @@ export
     bell, ghz, maximally_mixed,
     single_z, single_x, single_y,
     # Graphs
-    graphstate, graphstate!, graph_gatesequence, graph_gate,
+    graphstate, graphstate!, graph_gatesequence, graph_gate, GraphState, SingleQubitCliffordGroup,
```
```suggestion
    graphstate, graphstate!, graph_gatesequence, graph_gate,
```

Let's not export until we have something near finalized

test/test_graphs.jl
Krastanov, line 1: could you instead add a new file `test_graphs_...jl` -- I prefer to keep these files small, for ease of interactive debugging, as described in the SciML dev guide.

src/graphs/graphs.jl
```diff
@@ -1,13 +1,68 @@
...
+
+struct GraphState{G}
+    # Graph representing the graph state
+    graph::G
+    # Vertex Operators that need to be applied to a stabilizer state
+    # before the state is representable as a "pure" graph state
+    vops::Vector{SingleQubitCliffordGroup}
+end
+
+nqubits(g::GraphState{G}) where G = nv(g.graph)
```
Krastanov:
```suggestion
nqubits(g::GraphState) = nv(g.graph)
```

src/graphs/single_qubit_clifford.jl
```diff
@@ -0,0 +1,66 @@
+# Single qubit Clifford group
+
+# The group has a size of 24 in total
+# TODO: This could also be a place where interface is used, and we could use this stop gap implementation
+# to test future implementation of the single qubit clifford group interface.
+struct SingleQubitCliffordGroup
```
Krastanov: This does not seem to be a group, so we should think of a better name.

I am a bit confused why this is necessary though. It seems to exist just so that `GraphState.vops` can be hardcoded to `Vector{SingleQubitCliffordGroup}`. Would it not be simpler to just have `GraphState` have a type parameter specifying what the vop type is? Then we will not need the majority of methods defined here which anyway just dispatch to the same method for the inner type.

src/graphs/single_qubit_clifford.jl
```diff
@@ -0,0 +1,66 @@
...
+    else
+        return s
+    end
+end
+
+function SingleQubitCliffordGroup(s::AbstractSingleQubitOperator)
+    return SingleQubitCliffordGroup(check_single_qubit_operating_index(s))
+end
+
+Base.convert(::Type{SingleQubitCliffordGroup}, x::CliffordOperator) = SingleQubitCliffordGroup(x)
```
Krastanov: This looks like adding a lot of complexity. Why is this necessary?

src/graphs/single_qubit_clifford.jl
```diff
@@ -0,0 +1,66 @@
...
+const c_identity = one(CliffordOperator, 1)
+
+function Base.:(==)(l::SingleQubitCliffordGroup, r::SingleQubitCliffordGroup)
+    l.op == r.op
+end
+
+Base.show(io::IO, s::SingleQubitCliffordGroup) = show(io, s.op)
+
+function Base.:(*)(l::SingleQubitCliffordGroup, r::SingleQubitCliffordGroup)
+    SingleQubitCliffordGroup(l.op * r.op * c_identity)
```
Krastanov: This does not seem type stable. At line 7 you are saying op should be `SingleQubitOperator`, but `c_identity` is not of that type, rather it is `CliffordOperator`.

src/graphs/graphs.jl
```diff
@@ -1,13 +1,68 @@
...
+function GraphState(graph::G, h_idx::Vector{Int}, ip_idx::Vector{Int}, z_idx::Vector{Int}) where G
+    vops = ones(SingleQubitCliffordGroup, nv(graph))
+    for id in h_idx
+        vops[id] = sHadamard(1) * vops[id]
+    end
+    for id in ip_idx
+        vops[id] = sInvPhase(1) * vops[id]
+    end
+    for id in z_idx
+        vops[id] = sZ(1) * vops[id]
```
Krastanov: These should be `apply!` so that the operation is inplace. Otherwise each of these will be performing an allocation to create a new object.

src/graphs/graphs.jl
```diff
@@ -1,13 +1,68 @@
...
+    GraphState{G}(graph, vops)
+end
+
+function GraphState(stab::Stabilizer)
+    GraphState(graphstate(stab)...)
+end
+
+function Stabilizer(self::GraphState{<:AbstractGraph})
+    s_raw = Stabilizer(self.graph)
+    for id in 1:length(self.vops)
```
Krastanov: just to make the linter happy, this should probably be `eachindex` instead of `1:length(...)`

src/graphs/single_qubit_clifford.jl
LEXUGE: Minor comment: single qubit Clifford group should be a group with generator H and S?

LEXUGE: There are two difficulties:

1. `apply!` only allows us to multiply left, but not multiply right. However both are needed to update VOPs when doing Graph state-based simulation. So have a structure that allows us to do both will be helpful.
2. (Not really a difficulty) I only know that we can apply! Single qubit operators on a CliffordOperator. So I am not sure, as of now, what else we could use as the eltype for vops vector. And CliffordOperator is not an efficient data structure to store (and convert to).

Krastanov: Indeed, but this structure seems to represent group elements, not the group itself.

Krastanov: > apply! only allows us to multiply left, but not multiply right. However both are needed to update VOPs when doing Graph state-based simulation. So have a structure that allows us to do both will be helpful

Check #501 -- that should implement most of what you need

> (Not really a difficulty) I only know that we can apply! Single qubit operators on a CliffordOperator. So I am not sure, as of now, what else we could use as the eltype for vops vector.

I agree that only `CliffordOperator` can be easily acted upon. But my understanding is that means we should have the vops type be `CliffordOperator`, while if I am reading this correctly it is currently hardcoded to `SingleQubitOperator`.

LEXUGE: I am happy with the rest then if we have #501 merged.

Krastanov: Could you review and finish #501?

LEXUGE: Happy to. Not sure I can push to the branch though

Krastanov: if that is the case just start a new branch from it

LEXUGE: Upon a closer look, I am not entirely sure how `apply_inv!` can directly solve my issue though. Still applying from right is different from applying an inverse unless there is something I am missing.

Krastanov: Oh, I am probably the one that missed something. Could you list the operations you need in a bit more detail.

LEXUGE: Basically given Vertex Operator X, we could apply any single qubit Clifford gate A by both AX and XA

Krastanov: Yeah, we do not have a fast implementation for that yet.

Everything else we discussed stands, just use CliffordOperator for the vops. Then create this one single new (unexported, not optimized) function:

```
apply_right!(s::CliffordOperator, op)
    return apply!(copy(op), s)
end
```

We can optimize it afterwards.

Krastanov: and we can leave #501 to Sagnik, it does not seem to relate to what you are working on

Krastanov: oh, presumably you would want to act on the right with the sparse symbolic operators, in which case it is:

```
apply_right!(s::CliffordOperator, op::AbstractSingleQubitOperator)
    return apply!(copy(CliffordOperator(op,1)), s)
end
```

LEXUGE: I think this might not be the best way to implement `apply_right!`. Specifically,
```
julia> s = CliffordOperator(sZ(1), 1)
X₁ ⟼ - X
Z₁ ⟼ + Z

julia> @btime apply!(CliffordOperator(sX(1),1), s)
  943.185 ns (65 allocations: 2.28 KiB)
X₁ ⟼ + X
Z₁ ⟼ + Z
```
(P.S. I removed the `copy` as it's probably unnecessary as `CliffordOperator` already copied upon construction?)

while if we store `SingleQubitOperator` and start applying using `CliffordOperator` at the right-most (what `SingleQubitCliffordGroup` does currently)
```
julia> const clifford_id1 = one(CliffordOperator, 1)
X₁ ⟼ + X
Z₁ ⟼ + Z

julia> @btime sZ(1) * (sX(1) * clifford_id1)
  145.173 ns (16 allocations: 640 bytes)
X₁ ⟼ - X
Z₁ ⟼ - Z
```

Essentially, I think `CliffordOperator(::AbstractSingleQubitOperator)` is expensive
```
julia> @btime CliffordOperator(sX(1), 1)
  832.638 ns (64 allocations: 2.22 KiB)
X₁ ⟼ + X
Z₁ ⟼ - Z
```

Therefore, it's not a good idea to always promote operators to `CliffordOperator` before `apply!`. So I don't think directly storing `CliffordOperator` is a good initial approach for performance (though absolutely there are better ways in terms of performance than what I described above). That's why I put together this `SingleQubitCliffordGroup`.

So as it stands,
- I agree `SingleQubitCliffordGroup` seems awkward.
- But if we are to adopt the above implementation of `apply_right!`, then we should probably still have a `struct` similar to `SingleQubitCliffordGroup`?

Krastanov: I agree my suggestion is slow -- the point is to be minimal and selfcontained and we can fix the performance afterwards.

> while if we store SingleQubitOperator and start applying using CliffordOperator at the right-most (what SingleQubitCliffordGroup does currently)

I do not really understand your suggestion. The fast benchmarks you show are of left application, not of right application. And you are not storing SingleQubitOperator, you are storing CliffordOperator.

LEXUGE: Sorry I might not communicate my points clearly. 

In the current implementation, `SingleQubitCliffordGroup` does store `op::SingleQubitOperator`, not `CliffordOperator`. (Previously I did store `CliffordOperator` and the implementation was exactly as you suggested above).

And upon `apply_right!(s::SingleQubitCliffordGroup, op::AbstractSingleQubitOperator)`, it essentially does
```
s.op * (SingleQubitOperator(op) * clifford_id1)
```
where `const clifford_id1 = one(CliffordOperator, 1)`. So that implements the right multiplication using existing methods on left multiplication.

Krastanov: I am still a bit lost. The output of the operation you are doing is `CliffordOperator`, not `SingleQubitOperator`

```
julia> typeof( sZ(1)* (sX(1)*one(CliffordOperator,1)) )
CliffordOperator
```

LEXUGE: Yes, and in the current case I "translate" it back to `SingleQubitOperator` by `SingleQubitOperator(::CliffordOperator)` and store it that way.

LEXUGE: So in summary if we agree we could use the `s.op * (SingleQubitOperator(op) * clifford_id1)` trick as a _stop-gap_ solution for now, I will
- implement the `Rank1TensorProduct{S}` as a `Vector{S}`.
- implement `apply!` and `apply_right!` for `Rank1TensorProduct{SingleQubitOperator}`.
- use `Rank1TensorProduct{S}` for `vops` in `GraphState{G, S}`

Krastanov: I see, thanks for walking me through all of this!

Sounds good, stick to a parameterized vop type in GraphState, use `SingleQubitOperator` by default, and define `apply_right!(s::SingleQubitOperator, op::AbstractSingleQubitOperator) = SingleQubitOperator(op*(s*id))`. That way it should be following your idea, but with much shorter code

Krastanov: I commented before your last message showed up. You can implement `Rank1TensorProduct` or you can directly have a list of vops in your `GraphState`. I do not see big pros and cons distinguishing those two options.

LEXUGE: I think I will go with the r1tp approach for now as our implementation of `apply!(::SingleQubitOperator, ::AbstractSingleQubitOperator)` might not be wanted by other users.

Krastanov: Sounds good. But also, generally, do not worry about adding new methods of existing functions, even if they are slower-than-they-can be. Better to have a slow correct implementation than no implementation at all. Of course, if the API just does not make sense, we should avoid it, but the apply! you suggest is reasonable.

Krastanov:
The fact they are immutable and very small concrete structs means that they get stack allocated or even just never allocated. This is drastically more efficient than heap allocation (the only allocation that is actually measured by profiling tools).

So making a new one should not lead to allocations. You can even make a few new constructors for ease of use and as long as they are type stable the compiler will probably optimize away most operations related to creation.

src/graphs/graphs.jl
```diff
@@ -184,4 +256,4 @@ function graph_gate(h_idx, ip_idx, z_idx, n)
         end
     end
     c
-end
+end
```
Krastanov:
```suggestion
end

```

src/graphs/graphs.jl
```diff
@@ -1,13 +1,85 @@
...
+    # as converting to `CliffordOperator` is expensive
+    self[q] = SingleQubitOperator(op_prime * (self[q] * clifford_id1))
+end
+
+function apply_right!(self::Vector{SingleQubitOperator}, op::AbstractSingleQubitOperator; phases::Bool=true)
+    q = affectedqubits(op)[1]
+    # Convert into a single qubit operator acting on index 1
+    op_prime = SingleQubitOperator(SingleQubitOperator(op), 1)
+    self[q] = SingleQubitOperator(self[q] * (op_prime * clifford_id1))
+end
```
Krastanov: `apply_right!` is not being tested or used yet, so let's remove it from this PR

src/graphs/graphs.jl
```diff
@@ -1,13 +1,85 @@
...
+const clifford_id1 = one(CliffordOperator, 1)
+
+function apply!(self::Vector{SingleQubitOperator}, op::AbstractSingleQubitOperator; phases::Bool=true)
+    q = affectedqubits(op)[1]
+    # Convert into a single qubit operator acting on index 1
+    op_prime = SingleQubitOperator(SingleQubitOperator(op), 1)
+    # A trick that performs slightly better than directly op_prime * CliffordOperator(self[q])
+    # as converting to `CliffordOperator` is expensive
+    self[q] = SingleQubitOperator(op_prime * (self[q] * clifford_id1))
+end
```
Krastanov: This is being used only internally in your code, it does not seem to be meant as public api, so let's (for now) rename it to `_apply_vops!`. If this ends up being something that the user is expected to use, we can change this.

src/graphs/graphs.jl
```diff
@@ -1,13 +1,85 @@
-import Graphs
+import Graphs: Graphs, nv, AbstractGraph
+
+const clifford_id1 = one(CliffordOperator, 1)
+
+function apply!(self::Vector{SingleQubitOperator}, op::AbstractSingleQubitOperator; phases::Bool=true)
```
Krastanov: The use of `self` for an argument name is a bit strange (pythonic?). Maybe we can call it `vops`?

src/graphs/graphs.jl
Krastanov: Just trying to avoid whitespace noise -- the ending of this file was modified to lack a return, which is a frequent source of noise in git diffs.

You can configure your IDE to fix issues like this automatically on save.

src/graphs/graphs.jl
LEXUGE: Is it OK to just comment it out and leave a note

Krastanov: sure

Krastanov:
Marking it with skip-changelog as nothing here is public yet

LEXUGE:
I don't seen to have permission to merge though. So could you help me merge it after CI finishes or add me to the relevant permission group :)

Krastanov:
thanks! I will try to fix up the permissions on the project later, they are a bit of a mess right now

PR #534, opened 2025-06-30
D-dimensional surface and toric codes via Oscar's chain complexes and GF2 homology

Fe-r-oz:
This PR implements the D-dimensional surface codes via chain complexes and GF2 homology. All the parity check matrices and the metachecks matrices are generated from https://arxiv.org/pdf/2408.08865. 

The CSS orthogonality between $H_Z^T*H_X = 0$ has been verified with the former matrix is $H_Z^T$ and latter is $H_X$. Since the following matrices indeed obey the commutation check,  t̶h̶e̶ n̶e̶x̶t̶ s̶t̶e̶p̶ i̶s̶ t̶o̶ c̶o̶n̶s̶t̶r̶u̶c̶t̶ Q̶E̶C̶C̶ a̶n̶d̶ d̶o̶c̶u̶m̶e̶n̶t̶a̶t̶i̶o̶n̶. I̶n̶ a̶d̶d̶i̶t̶i̶o̶n̶, g̶e̶n̶e̶r̶a̶l̶i̶z̶i̶n̶g̶ i̶t̶ t̶o̶ D̶-̶d̶i̶m̶e̶n̶s̶i̶o̶n̶s̶ w̶h̶i̶c̶h̶ i̶s̶ n̶o̶t̶ d̶o̶n̶e̶ i̶n̶ t̶h̶e̶ p̶a̶p̶e̶r̶,  a̶n̶d̶ w̶h̶e̶t̶h̶e̶r̶ w̶e̶ c̶a̶n̶ u̶s̶e̶ L̶ o̶t̶h̶e̶r̶ t̶h̶a̶n̶ `̶L̶ =̶ [̶1̶ 1̶]̶ `̶w̶h̶i̶c̶h̶ i̶s̶ u̶s̶e̶d̶ a̶s̶ t̶h̶e̶ r̶e̶p̶e̶t̶i̶t̶i̶o̶n̶ c̶o̶d̶e̶ s̶e̶e̶d̶ m̶a̶t̶r̶i̶x̶ w̶i̶l̶l̶ b̶e̶ c̶o̶n̶s̶i̶d̶e̶r̶e̶d̶.

 The PR generalizes 2D, 3D, 4D surface codes to D-dimensions and tests have been added up to 5D surface codes. We can add the `:slow` tag and add more tests as it takes time from 6D surface code onwards.

 The PR also implements the subfamilies of each of these codes:
- `[[L² + (L − 1)², 1, L]]` 2D surface code
- `[[L³ + 2L(L − 1)², 1, min(L, L²)]]` 3D surface code
- `[[6L⁴ − 12L³ + 10L² − 4L + 1, 1, L²]]` 4D surface code and so on.

In addition, the D-dimensional toric codes are also implemented.

CSS orthogonality checks:
```
julia> mod.(matrix_to_int(d_dimensional_surface_code(4,2)[2])*matrix_to_int(d_dimensional_surface_code(4,2)[1]) , 2)
20×20 Matrix{Int64}:
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0
 0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0

julia> mod.(matrix_to_int(d_dimensional_surface_code(3,2)[2])*matrix_to_int(d_dimensional_surface_code(3,2)[1]) , 2)
12×2 Matrix{Int64}:
 0  0
 0  0
 0  0
 0  0
 0  0
 0  0
 0  0
 0  0
 0  0
 0  0
 0  0
 0  0

julia> mod.(matrix_to_int(d_dimensional_surface_code(2,2)[2])*matrix_to_int(d_dimensional_surface_code(2,2)[1]) , 2)
2×2 Matrix{Int64}:
 0  0
 0  0
```

The paper used Sage for implementation and there are canonical differences between the resulting matrices due to different conventions which lead to column permutations, For Example, in the case of L =2, D = 2,

Oscar
```
julia> d_dimensional_surface_code(2,2)[2]
[1   1   0   0   1]
[0   0   1   1   1]

julia> transpose(d_dimensional_surface_code(2,2)[1])
[1   0   1   0   1]
[0   1   0   1   1]
```
which is not canonically equivalent (even after canonicalization/Gaussian elimination via echelon_form) to due to columns permutations because of the different conventions 

Sage/paper
```
1 1 1 0 0
1 0 0 1 1

1 1 0 1 0
1 0 1 0 1
```

These matrices are column permutation equivalent which can be verified using (this takes huge time for larger matrices)

```
using Combinatorics
function is_equivalent_up_to_column_permutation(H1, H2)
    n = size(H1, 2)
    F = GF(2)
    rank(matrix(F, H1)) == rank(matrix(F, H2)) || return false
    rank(vcat(matrix(F, H1), matrix(F, H2))) == rank(matrix(F, H1)) && return true
    # for all column permutations (for small n)
    for perm in permutations(1:n)
        H2_perm = H2[:, perm]
        if rank(vcat(matrix(F, H1), matrix(F, H2_perm))) == rank(matrix(F, H1))
            return true
        end
    end
    return false
end

```
🐞 Computing the total complex of a tensor product of chain complexes in` Oscar` throws error. Based on the stacktrace, no method for `rank` was defined for this `DirectSumModule` in `AbstractAlgebra`, which caused this error. For the time being, I have defined the `rank` method here and submitted [PR ](https://github.com/Nemocas/AbstractAlgebra.jl/pull/2114)to AA.

```
ERROR: MethodError: no method matching rank(::AbstractAlgebra.Generic.DirectSumModule{FqFieldElem})
The function `rank` exists, but no method is defined for this combination of argument types.
```

Acknowledgements: Thank you to @thofma for the invaluable guidance on this!


ext/QuantumCliffordOscarExt/d_dimensional_codes.jl
```diff
@@ -0,0 +1,515 @@
+struct DDimensionalCode <: AbstractCSSCode
+    """Dimension of the code (must be ≥ 2)"""
+    D::Int
+    """Size parameter determining the code family. For surface codes: number of physical qubits along each dimension."""
+    L::Int
+    """Code type, either `:Surface` or `:Toric`."""
+    type::Symbol
```
nzy1997: Consider naming these `DDimensionalSurfaceCode` and `DDimensionalToricCode instead`. This would be more user-friendly, as it eliminates the need to learn separate functions like `d_dimensional_toric_codes` and `d_dimensional_surface_codes` to construct a new code.

ext/QuantumCliffordOscarExt/d_dimensional_codes.jl
```diff
@@ -0,0 +1,515 @@
...
+Here is an example of `[[12, 1, 2]]` `3D` Surface code with `L = 2` from [Berthusen_2024](@cite).
+
+```jldoctest threeDsurface
+julia> using Oscar; using QuantumClifford; using QuantumClifford.ECC; using QECCore: parity_matrix;
+
+julia> D = 3; L = 2;
+
+julia> c = d_dimensional_surface_codes(D, L);
+
+julia> code = parity_matrix(c)
```
nzy1997: Here we sholud use `parity_check`. `parity_matrix` should return the "bool-valued" partiry matrix.

ext/QuantumCliffordOscarExt/d_dimensional_codes.jl
```diff
@@ -0,0 +1,515 @@
...
+d_dimensional_toric_codes(D::Int, L::Int) = DDimensionalCode(D, L, :Toric)
+
+function parity_matrix_xz(c::DDimensionalCode)
+    c.D >= 2 || throw(ArgumentError("Dimension must be at least 2 to construct a valid D-dimensional code."))
+    hx, hz = c.type == :Surface ? _parity_matrix_xz_surface(c) : _parity_matrix_xz_toric(c)
+    # Oscar returns pcm hx', so we transpose it to convert it back to hx. See page 11, B2 for reference.
+    return hx', hz
+end
+
+function pcms(c::DDimensionalCode)
```
nzy1997: Does this function return the parity check matrix of the code? If so, rename it to `parity_matrix`. If not, please add a docstring since it is a exported function.

ext/QuantumCliffordOscarExt/d_dimensional_codes.jl
nzy1997: Also, for functions like `code_n` defined on them both. We can define an abstract type `DDimensionalCode`. And remember to use `DocStringExtensions`.

ext/QuantumCliffordOscarExt/d_dimensional_codes.jl
Fe-r-oz: The `pcms` returns all the boundary maps associated with the chain complexes -- so it includes both parity check matrices and metacheck matrices. The convention in the paper refers to both of these types of matrices as it `pcms`, so I went this this convention. Alternative is to use `boundary_maps` instead of `pcms` instead which is a bit more clearer and provides more context for what the function returns. Indeed, I will add docstring to elaborate on this and change name to `boundary_maps`.

nzy1997:
Looks good! Thank you for the changes!

Fe-r-oz:
Thank you for the feedback!

**PR Summary**

This PR introduces D-dimensional surface and toric codes using `Oscar`'s chain complexes and `GF(2)` homology.

**Future work**

- Currently, `Oscar.homology` throws an error for repetition code chain complexes over `AbstractAlgebra.FiniteField` modules. Once the fix I submitted to `Oscar` is merged, I will generalize `code_k` to compute the number of logical qubits via the dimension of the `Oscar.homology` group. For `2D`, `3D`, `4D` codes, it is proven to be `1` in the reference paper.
- Implementation of **single-stage** and **two-stage** decoders using metachecks for single-shot QEC (decoding is out of scope for this PR, which focuses on code construction).
- Exploration and investigation of `5D` and higher-dimensional codes. The reference explored only up to `4D.`  Investigation of `3D` codes a bit further to get more insights into the asymmetry of `X`- and `Z`-bases.
- There is an interesting code distance result that the dimension of the QECCs over chain complexes can be calculated exactly using insights from [Künneth formula](https://en.wikipedia.org/wiki/K%C3%BCnneth_theorem). This was first proposed in [here](https://arxiv.org/abs/1810.01519) and acknowledged again the reference paper.
- > (Optional) "An equivalent method to iteratively increasing the dimension as we did here is to take the tensor product of D repetition codes at once [24](https://arxiv.org/abs/1810.01519), obtaining triple, quadruple, or higher dimensional complexes before summing and to obtain a total complex representing the D-dimensional surface/toric codes." ~ quoted from the reference paper
- Implementation of `LPCode` over chain complexes? Any benefits? That can be helpful since we can avoid generalizing it over to `l`-dimensional `Hecke.associative_algebra`s?
- (Optional) Finding out the difference in conventions of Homological Algebra implementation between `Oscar` and `Sage`.
- Implementation of other chain complexes other than repetition code chain complexes?
- [Homological Product Codes](https://arxiv.org/pdf/1311.0885)

**cc:** @Krastanov

ext/QuantumCliffordOscarExt/d_dimensional_codes.jl
Fe-r-oz: The reference paper did not explicitly construct the `DDimensionalToricCode` or derive `[[N, K, D]]` parameters for these codes. However, it suggested that the same procedure could be applied with the caveat of using "_the L×L repetition code and only the chain complex C_." This stated procedure was used for the construction of `DDimensionalToricCode`s.

Upon investigation, the procedure for calculating `N`, `K`, was restricted to `DDimensionalSurfaceCode` because it is using both chain complex (`C`) and its dual chain complex (`D`), which is required for  `DDimensionalSurfaceCode`  codes only. So, corrected `code_n(c::DDimensionalCode)` to using separate derivations of `N` for  `code_n(c::DDimensionalSurfaceCode)` and  `code_n(c::DDimensionalToricCode)` . The latter is derived using only one chain complex. **Edit**: I have created a helper function `_total_complex_toric/surface` and use it to return the appropriate complex and define `code_n(c::DDimensionalCode)` instead of separate functions.

Krastanov:
@Fe-r-oz , if you make changes after the last round of review, please wait for those changes to be reviewed as well, even if they are just additional tests or documentation. I know it slows down things a bit, but it is a useful sanity check.

PR #535, opened 2025-07-02
add a commutativity check to the LPCode constructor

Krastanov:
a warning to the user, while waiting for #511


Fe-r-oz:
Thank you for adding the check `check_allrowscommute` inside the `CSS` constructor!

PR #538, opened 2025-07-03
Move Reed Muller Codes

nzy1997:
Move `ReedMuller`, `RecursiveReedMuller`, and `QuantumReedMuller`  to `QECCore` from `QuantumClifford.ECC`.


lib/QECCore/test/codes/quantumreedmuller.jl
```diff
@@ -21,31 +17,36 @@
...
-            @test canonicalize!(parity_checks(qrm₁₅₁₃)) == canonicalize!(parity_checks(QuantumReedMuller(4)))
+            @test H == parity_matrix(CSS(parity_matrix_x(QuantumReedMuller(m)), parity_matrix_z(QuantumReedMuller(m))))
         end
     end
+
+
+    @testset "QuantumReedMuller equivalence to simple codes" begin
+        # QuantumReedMuller(3) is the Steane7 code.
+        # TODO: add function to check if two codes are equivalent, like @test canonicalize!(parity_checks(Steane7())) == parity_checks(QuantumReedMuller(3))
+        @test parity_matrix(Steane7()) == parity_matrix(QuantumReedMuller(3))[[3,2,1,6,5,4],:]
```
Krastanov: Nemo probably has rref canonicalization operation.

Krastanov: Maybe something as simple as `Nemo.rref(Nemo.matrix(Nemo.GF(2),matrix))`

Krastanov (changes requested):
Looks good! Feel free to merge after you add back the equivalency check (I think the Nemo snippet I provided should be enough)

Krastanov:
Side notes

- [ ] May we add a buldkite step that tests this library as well?
- [ ] May we switch to DocStringExtensions (see here for instance https://github.com/QuantumSavory/QuantumSavory.jl/blob/330037d7fdb5e25232bcf15c66e4de028ae700bb/src/ProtocolZoo/cutoff.jl#L1-L15)

PR #540, opened 2025-07-03
introduce Metachecks interface to `QECCore` for single shot QEC and repair syndome decoding

Fe-r-oz:
QEC codes that are constructed via chain-complex framework  utilize metachecks for: 1) single-shot QEC capabilities 2) syndrome validation and correction.

### Proposal
I have implemented the 4D surface codes and related families that generalize to D-dimensional surface codes  and Toric codes in #534. Since I am using the chain-complex framework, the boundary maps returns the both parity check matrices and metacheck matrices, so I propose that we introduce the following functions for nicer API to have access to these meta checks matrices:

- `metachecks_z`
- `metachecks_x`
- `metachecks`

This will return the the $X$ and $Z$ metacheck matrices for a D-dimensional code (in general codes constructed via chain complexes).  For standard CSS codes without explicit metachecks, we throw error that such matrices are not present.

### D-dimensional codes

In this example, we use the $4D$ surface code. For $4D$ surface code that are constructed using as chain complexes:

$C = C_4 → C_3 → C_2 → C_1 → C_0$

with boundary maps $∂_{i}$ satisfying $∂_{i+1}∂_{i} = 0$, we have:

- **Z-metachecks**: $M_Z = ∂_4$ with $H_Z = ∂_3^T$
- **X-metachecks**: $M_X = ∂_1$ with $H_X = ∂_2$
- **Constraints**: $M_Z H_Z = 0$ and $M_X H_X = 0$

Valid syndromes must satisfy $s ∈ ker(M)$ (the metacheck condition).

### Reference

> Due to the constraints of the boundary maps, $∂_{i+1}∂_{i} = M_Z H_Z = M_X H_X = 0$. Hence valid $Z(X)$ syndromes are in $ker M_Z (ker M_X )$. In the presence of a measurement error affecting a syndrome, we will get a nonzero metasyndrome which we can use to correct the syndrome before attempting to decode. This technique of “repair syndrome” decoding was used to show competitive thresholds for the 4D surface code.

- [ ] We recently started enforcing formatting checks. If formatting issues are reported in the new code you have written, please correct them


Krastanov:
Give me some time before merging new functionality to look through it in the future. Exporting new function names commits us to long term support, and that needs to be gauged pretty carefuly. As long as things are not exported, things can be laxer.

This looks good, but could you give some more documentation for what metachecks are and why would anyone want to use them. Currently the user would be a bit lost if they see this function.

PR #544, opened 2025-07-06
Add support for sparse measurements

J-C-Q:
## Motivation
Measurements of `PauliOperators` on an `n` qubit state are currently either performed by checking the commutation relation between each row and the whole `n` qubit `PauliOperator`, or for single qubit measurements using the specialized functions `projectX!`, ... This specialized algorithm is great. However, since stabilizer measurements usually contain more than one qubit, a version optimized for more generalized sparse `PauliOperators` could be useful. 

While the code for single qubit measurement can (more or less) easily be generalized to two qubits, the memory layout of the tableau makes this sparse checking less efficient.

## Implementation
This PR tries to find a balance between efficiently using the memory layout of the tableau and sparsely computing commutation.
This is done by only computing the `xor` (in the `comm` function) for `UInt64`s in the tableau row that contain the sparsely selected qubits. This results in sparse scenarios where the qubits are close together (meaning packed into the same `UInt64`) to be very fast. For more uniformly distributed qubit selections, the speed-up is not as significant, or even non-existent. Nevertheless, for very large system sizes, measurements with a few qubits (e.g. 3 or 4) qubits are still faster this way. Especially, two-qubit parity checks.

However, right now, this requires computing the indices of the `UInt64`s in which the qubits are packed, which does allocate. Maybe there is a better way to do this.

## TODO
There is still some cleanup to be done. As well as documentation and testing. However, I thought I'd share this idea already to get feedback if you are even interested in integrating this. I'm very open to any criticism and ideas to improve my code, as well as happy to work on this more. 

I could also try to implement a two qubit parity measurement specialized function (analogous to the singe qubit functions), that would probably beat this one, if that is something you are interested in.

## Benchmark
I did some rudimentary benchmarking. Attached you can find a plot of the results for a system of `2048` qubits. The time is in seconds. 

The base case represents the `n` qubit `PauliOperator`, measured using the existing `project!` function.
```julia
n = 2048
k = 2
m = MixedDestabilizer(one(Stabilizer, n))
p = embed(n,rand(1:n,k),random_pauli(k))
@belapsed  QuantumClifford.project!($m, $p; keep_result=false,phases=false)
```
The best case represents the results for densely packed spares measurements.
```julia
n = 2048
k = 2
m = MixedDestabilizer(one(Stabilizer, n))
qubits = collect(1:k)
local_pauli = random_pauli(k)
@belapsed  QuantumClifford.project_sparse!($m, $qubits, $local_pauli;keep_result=false,phases=false)
```
The worst case for uniformly (maybe not the best way of doing this) distributed measurements.
```julia
n = 2048
k = 2
m = MixedDestabilizer(one(Stabilizer, n))
qubits = ceil.(Int,collect(range(1,n,k)))
local_pauli = random_pauli(k)
@belapsed  QuantumClifford.project_sparse!($m, $qubits, $local_pauli;keep_result=false,phases=false)
```
I'm not sure what causes the jumps in the worst case.
fig

### DIFF
--- a/src/project_trace_reset.jl
+++ b/src/project_trace_reset.jl
@@ -439,6 +439,129 @@ function _project!(d::MixedDestabilizer,pauli::PauliOperator;keep_result::Val{Bk
 end
 
 """
+A faster special-case version of [`project!`](@ref) for sparse PauliOperators.
+
+This function is faster when the PauliOperator is sparse, i.e. it has only a small number of (not identity) elements, compared to the number of system qubits. Especially when the measured qubits are close to each other.
+
+For single qubit measurements, use [`projectX!`](@ref), [`projectY!`](@ref), or [`projectZ!`](@ref), since they are faster.
+
+See also: [`project!`](@ref), [`projectX!`](@ref), [`projectY!`](@ref), [`projectZ!`](@ref).
+"""
+function project_sparse!(state,qubits::Vector{<:Integer}, local_pauli::PauliOperator;keep_result::Bool=true,phases::Bool=true)
+    @valbooldispatch _project_sparse!(state,qubits,local_pauli;keep_result=Val(keep_result),phases=Val(phases)) keep_result phases
+end
+function _project_sparse!(d::MixedDestabilizer,qubits::Vector{<:Integer},local_pauli::PauliOperator;keep_result::Val{Bkr}=Val(true),phases::Val{Bp}=Val(true)) where {Bkr, Bp}
+    anticommutes = 0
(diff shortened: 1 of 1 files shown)


Krastanov:
This is interesting. Do you have a use case you are developing this for, or just an interesting optimization?

I would suggest introducing a new `comm` method that takes an extra argument `qubits`, and then the indexing calculation happens on the fly in the loop.

Overall though, I think any SIMD speedup gets removed here, so in many situations I would expect the naive dense approach to be faster.

PR #545, opened 2025-07-06
add more documentation about metachecks

Fe-r-oz:
This PR aims to furnish the documentation about metachecks introduced in #540.

The single-shot QEC and metachecks are mentioned in great detail in https://arxiv.org/pdf/2206.03122, page 5. I used the insights  mentioned in this paper to write the documentation of metachecks along with polish from another paper.

To illustrate metachecks, this paper used a classical repetition code’s Tanner graph with a metacheck matrix `m₁:` 
> **Edit^2**: This approach is used in the **single-stage decoding**.  The other decoding approach that utilize metachecks matrices is referred as two-stage decoding aka repair syndrome decoding.

```mermaid
graph TD
    subgraph Check Nodes
        direction LR
        f1[f1]
        f2[f2]
        f3[f3]
        m1[m1]
    end
    subgraph Varaible Nodes
        direction LR
        e1((e1))
        e2((e2))
        e3((e3))
        se1((se1))
        se2((se2))
        se3((se3))
    end

    style m1 fill:#FFCCCC,stroke:#FF0000,stroke-width:2px,color:#FF0000
    style se1 fill:#CCCCFF,stroke:#0000FF,stroke-width:2px,color:#0000FF
    style se2 fill:#CCCCFF,stroke:#0000FF,stroke-width:2px,color:#0000FF
    style se3 fill:#CCCCFF,stroke:#0000FF,stroke-width:2px,color:#0000FF

    f1 --- e1
    f1 --- e3
    f2 --- e1
    f2 --- e2
    f3 --- e2
    f3 --- e3

    f1 --- se1
    f2 --- se2
    f3 --- se3

    m1 --- se1
    m1 --- se2
    m1 --- se3

    linkStyle 6 stroke:#0000FF,stroke-width:2px
    linkStyle 7 stroke:#0000FF,stroke-width:2px
    linkStyle 8 stroke:#0000FF,stroke-width:2px

    linkStyle 9 stroke:#FF0000,stroke-width:2px
    linkStyle 10 stroke:#FF0000,stroke-width:2px
    linkStyle 11 stroke:#FF0000,stroke-width:2px 
```

The metacheck `m₁ = f₁ + f₂ + f₃` imposes a parity constraint on the syndrome bits `se₁, se₂, se₃`. If `m₁(se₁, se₂, se₃) ≠ 0`, this means that a measurement error has occurred in one of the syndrome bits themselves. Such measurement errors are usually detected by repeating stabilizer measurements, e.g. O(d) for surface code of distance d. 

In contrast, single-shot QEC uses additional parity constraints like` m₁` to detect measurement faults immediately, without requiring multiple rounds. These additional parity constrained are callled metachecks  aka "checks of checks" because they validate the measured syndrome bits. The higher dimensional codes, such as ones constructed via chain complex framework, esp have metachecks matrices along with parity checks matrices. In documentation, we demonstrate this using 4D surface code which has 5-term chain complex representation.

Help/suggestions in polishing the documentation and making it reader friendly is much appreciated. 

**Edit**

### Reference

Interestingly, not all codes that support single-shot QEC require metachecks. The paper also used 4D surface code when providing an example of metachecks

> Some single-shot codes have linear dependencies amongst the check operators, leading to syndromes becoming codewords of a classical linear code (called a metacode) that can be used for syndrome repair. These linear dependencies are not a requirement for a code to be single-shot (indeed quantum expander codes are single-shot and confined, but can have full rank check matrices), however a metacode can nevertheless be useful when decoding. We can construct a code that has syndromes encoded in a metacode using a chain complex with length at least **three** (to obtain a metacode for either X or Z syndromes) (e.g. 3D surface code), or length **four** if we would like a metacode for both X and Z syndromes. (e.g. 4D surface code). 

### DIFF
--- a/docs/src/references.bib
+++ b/docs/src/references.bib
@@ -1131,3 +1131,14 @@
   year={1996},
   publisher={APS}
 }
+
+@article{quintavalle2021single,
+  title={Single-shot error correction of three-dimensional homological product codes},
+  author={Quintavalle, Armanda O and Vasmer, Michael and Roffe, Joschka and Campbell, Earl T},
+  journal={PRX Quantum},
+  volume={2},
+  number={2},
+  pages={020340},
+  year={2021},
+  publisher={APS}
+}


Fe-r-oz:
This PR is ready for review! Thank you! I hope this is reader friendly documentation. Please feel free to modify as per your convenience.

I have added points about the origin of metachecks, single-shot codes and use of metachecks  in decoding from https://arxiv.org/pdf/2206.03122 (This paper  is all about single-shot decoding for the D-dimensional codes and other single-shot codes). It proposed two types of decoders or decoding strategies using metachecks.

The actual paper I am following is about construction of these d-dimensional codes, although it uses single-shot decoding as well. How to use metachecks in decoding (two-stage decoding and single-stage decoding) would be separate PR in the future.

Krastanov:
Wow, that is very detailed! I need to be careful when I say the reader might not understand something or you would end up writing a textbook. Thanks, it looks great.

I have not read through it in detail yet. I will try to review later this week.

Fe-r-oz:
Hi Stefan,  

Would you mind reviewing the documentation updates when you have a moment? Please feel free to modify it as needed.  

I have not added additional formal statements from **Section III** yet, though they are described more thoroughly in this   [reference](https://arxiv.org/abs/2009.11790). I plan to review it soon and add additional details if required in another PR.

I’ve also attached some additional cherry-picked notes regarding single shot QEC that might be helpful to improve this documentation if required.

> The three main strategies for tackling noisy measurements are: repeated measurements on the code; performing measurement driven error-correction on a cluster state; or using a single-shot code and decoder.

> The single-shot approach has the advantage of no additional time cost or cluster state generation cost and provides a resilience against time correlated noise.

> In single-shot error correction, some residual error persists after each round of error correction, but this residual error is kept small and does not rapidly accumulate. However, only a special class of codes support single shot error correction, but exactly which codes and why is not  yet fully understood.

> Bomb´ın coined the phrase single-shot error correction and remarked that it “is related to self-correction and confinement phenomena in the corresponding quantum Hamiltonian model.”

> We introduce a general concept of confinement for quantum codes, which roughly stipulates qubit errors cannot grow without triggering more measurement syndromes.

> Moreover, the concept of confinement is closely related to soundness but it is weaker and so able to encompass more families of codes, such as the expander codes which are confined but not sound. Roughly speaking, a code has good confinement if small qubit errors produce small measurement syndromes; this differs from good soundness which entails that small syndromes can be produced by small errors.

> Our main motivation to introduce the concept of confinement and the Shadow decoder was to find a feature of codes
able to encompass all known examples of single-shot codes.

> Campbell introduced the notion of soundness and showed that this property is a sufficient condition for codes to show
single-shot properties in the adversarial setting. Nonetheless, Fawzi et al. showed that expander codes have a singleshot threshold for local stochastic noise, even though they do not have the soundness property. As already said though, expander codes do have confinement. It is proved in another literature that that their confinement function is linear and call this property robustness. Confinement, in other words, fills the gap leaved by the concept of soundness.

lib/QECCore/src/interface.jl
```diff
@@ -120,10 +120,31 @@ abstract type AbstractDistanceAlg end
 """
     metacheck_matrix_x(c::AbstractCSSCode)
 
-Returns the `X`-metacheck matrix (``\\partial_{Mx}`` boundary map in
-[chain complex](https://en.wikipedia.org/wiki/Chain_complex) notation) for a CSS code.
+Returns the `X`-metacheck matrix (``M_X = \\partial_{i-1}``) for a CSS code defined
+by a chain complex of length `l ≥ 4`, where qubits are placed on `i`-cells with
+``1 < i < l−1``. This matrix acts on `X`-syndromes (measurement outcomes of `Z`-type
+stabilizers) and enforces the constraint ``M_Xs_X = 0``, ensuring syndromes are valid
```
Krastanov:
```suggestion
stabilizers) and enforces the constraint ``M_Xs_X = 0``, ensuring syndromes ``s_X`` are valid
```

Krastanov: also specify "Z-type stabilizers (as provided by the H_WHAT matrix, implemented in `parity_checks_WHAT`)" and make sure we do not have super inconsistent indexing here

lib/QECCore/src/interface.jl
```diff
@@ -120,10 +120,31 @@ abstract type AbstractDistanceAlg end
...
+\\begin{aligned}
+C_4 \\xrightarrow{\\partial_4} C_3 \\xrightarrow{\\partial_3} C_2 \\xrightarrow{\\partial_2} C_1 \\xrightarrow{\\partial_1} C_0
+\\end{aligned}
+```
+
+the metacheck matrix ``M_X = \\partial_1`` satisfies the following:
+
+- Acts on `X`-syndromes: ``s_X \\in C_1``.
+- It enforces ``M_Xs_X = 0``, i.e. only valid syndromes lie in ``\\ker M_X``.
+- It satisfies the boundary condition ``M_XH_X = 0`` (i.e., ``\\partial_1 \\partial_2 = 0``).
```
Krastanov: Here just double check that `H_X` is indeed the one you want and remind again folks which `parity_checks_WHAT` gives that matrix.

PR #547, opened 2025-07-09
DO NOT MERGE test rocm buildkite

Krastanov:


Krastanov:
@Hamiltonian-Action , check this PR for how to set up the CI for the rocm runner. The nvidia runner will need a bit more time as the debian trixie nvidia packages are broken right now.

Modify the buildkite file in your branch as you see here and update your tests accordingly.

Krastanov:
The `queue: rocm` line in the buildkite config needs to stay the same, but you can change all other occurrences of ROCM to whatever is most appropriate / consistent with current convention in the codebase. 

> NVIDIA is notoriously difficult to deal with when it comes to their Linux support, not surprised in the slightest.

That is funny, everyone usually complains about ROCM being a pain while praising how well CUDA works on linux. But in my case the issue just happened to be that the LAN driver required a newer version of debian, and the debian folks have not finished packaging CUDA for the upcoming release.

PR #557, opened 2025-07-20
Added good_state validation for VerifyOp and included test for it

brohan-byte:
- Added a guard in the applywstatus! for VerifyOp which checks if the Stabilizer passed in for good_state is actually complete stabilizer (has no degrees of freedom). 

- Included tests for VerifyOp's applywstatus! to check if this functionality works


Krastanov:
Thank you!

PR #558, opened 2025-07-20
Add table and utils necessary for graph state simulator

LEXUGE:
This PR adds
- function that decomposes single qubit Clifford operator into specific generators (needed for CPHASE simulation)
- function that applies CPHASE to isolated qubits while satisfying constraints (needed for CPHASE simulation)
- precompiled tables generated using functions above

It will be part of the "bigger" effort to add CPHASE support to the graph state simulator.

Checklist

### DIFF
--- a/src/graphs/utils.jl
+++ b/src/graphs/utils.jl
@@ -1,0 +1,146 @@
+# Code and tables useful to graph state simulation.
+
+# A brute-force, dynamical-programming based approach on
+# finding minimal representation (decomposition) of any group elements in terms of generators provided.
+#
+# We don't need to canonicalize anything because it's single qubit operator and representation is unique.
+function decompose(e::AbstractSingleQubitOperator, generators::Vector{<:AbstractSingleQubitOperator})
+    e = CliffordOperator(e, 1)
+
+    # Initialize the table with generators. Table shouldn't include identity `e`
+    # because composing with identity gives nothing new.
+    table = Dict{CliffordOperator,Vector{AbstractSingleQubitOperator}}()
+    for g in generators
+        push!(table, CliffordOperator(g, 1) => [g])
+    end
+
+    updated = true
+    while updated
+        updated = false
+        k = copy(keys(table))
+        for a in k
+            for b in k
+                # compose a new element and see if it's already in the table
+                if (a * b) in k
+                    if length(table[a]) + length(table[b]) < length(table[a*b])
+                        table[a*b] = append!(copy(table[a]), table[b])
+                        updated = true
+                    end
+                else
+                    # Don't store identity in the table
+                    if a * b != e
+                        table[a*b] = append!(copy(table[a]), table[b])
+                        updated = true
+                    end
+                end
+            end
+        end
+    end
+
+    # transform the key to SingleQubitOperator so it's easier to lookup
+    table_prime = Dict{SingleQubitOperator,Vector{AbstractSingleQubitOperator}}()
+    for (k, v) in table
+        table_prime[SingleQubitOperator(k)] = v
+    end
+    return table_prime
+end
+
+# The subgroup of single qubit Clifford group that commutes with CPHASE.
+const Z_COMMUTATION_SUBGROUP = [CliffordOperator(x, 1) for x in [sId1(1), sZ(1), sPhase(1), sInvPhase(1)]]
+
+# Stabilizers needed to convert between equivalent states to satisfy the commutation constraint (see cphase_isolated)
+# NOTE: Luckily, we don't need the whole stabilizer subgroup which is quite big, but only a selected few.
+const PAULI_STABILIZERS = Dict(
+    # CPHASE |++>
+    true => [
+        (sId1(1), sId1(1)), (sX(1), sZ(1)), (sZ(1), sX(1)), (sY(1), sY(1)),
+        (CliffordOperator(sHadamard) * CliffordOperator(sZ), CliffordOperator(sZ) * CliffordOperator(sHadamard)),
+        (CliffordOperator(sZ) * CliffordOperator(sHadamard), CliffordOperator(sHadamard) * CliffordOperator(sZ)),
+        #  (CliffordOperator(sHadamard), CliffordOperator(sHadamard))
+    ],
+    # |++>
+    false => [
+        (sId1(1), sId1(1)),
+        (sX(1), sId1(1)),
+        (sId1(1), sX(1)),
+        (sX(1), sX(1)),
+        #  (CliffordOperator(sHadamard), CliffordOperator(sHadamard))
+    ]
+)
+
+function gen_graph_state(connected::Bool, U1::SingleQubitOperator, U2::SingleQubitOperator)
+    if connected
+        g = GraphState(S"XZ ZX")
+    else
+        g = GraphState(S"XI IX")
+    end
+    vops(g)[1], vops(g)[2] = U1, U2
+    return g
+end
+
+# Convert the graphstate to stabilizer state and perform gate there.
+# The resulting state satisfy the following constraint:
+# - Let U1, U2 be VOPs of the initial state, U1', U2' be the VOPs of the state after CPHASE
+# - We guarantee U1' ∈ Z_COMMUTATION_SUBGROUP if U1 ∈ Z_COMMUTATION_SUBGROUP. Same for U2, U2'.
+function apply_cphase_isolated(connected::Bool, U1::SingleQubitOperator, U2::SingleQubitOperator)
+    g_init = gen_graph_state(connected, U1, U2)
+    stab = Stabilizer(copy(g_init))
+    apply!(stab, sCPHASE(1, 2))
+    g = GraphState(stab)
+
+    found = false
+    for s in PAULI_STABILIZERS[connected]
+        satisfied = true
+        # Compute the VOPs, using statbilizer, that give us an equivalent state
+        U_prime = [vops(g)[i] * (s[i] * one(CliffordOperator, 1)) for i in 1:2]
+        for qubit_idx in 1:2
+            # Check if constraint is satisfied
+            if (CliffordOperator(vops(g_init)[qubit_idx], 1) in Z_COMMUTATION_SUBGROUP) && !(U_prime[qubit_idx] in Z_COMMUTATION_SUBGROUP)
+                satisfied = false
+                break
+            end
+        end
+        if satisfied
+            vops(g)[1], vops(g)[2] = SingleQubitOperator(U_prime[1]), SingleQubitOperator(U_prime[2])
+            # check again it's indeed equivalent
+            if canonicalize!(stab) == canonicalize!(Stabilizer(g))
+                found = true
+                break
+            else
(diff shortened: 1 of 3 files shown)


Krastanov (changes requested):
Could you explain a bit more why this is necessary, where it fits in, etc? Example user-facing API for instance? Currently I do not really get a sense of why this functionality is necessary.

In particular, if this is part of how the inner loop of a simulator would work, I would be a bit worried. Dictionary lookups are n*log(n), which in terms of asymptotic behavior is not bad, but they require a lot of pointer chasing (as they are not dense structures) that leads to a very significant slowdown (by a constant factor).

src/graphs/utils.jl
```diff
@@ -0,0 +1,146 @@
+# Code and tables useful to graph state simulation.
+
+# A brute-force, dynamical-programming based approach on
+# finding minimal representation (decomposition) of any group elements in terms of generators provided.
+#
+# We don't need to canonicalize anything because it's single qubit operator and representation is unique.
+function decompose(e::AbstractSingleQubitOperator, generators::Vector{<:AbstractSingleQubitOperator})
```
Krastanov: This should probably be a docstring, not a comment. Could you make the name a bit more descriptive? We have a lot of things that are called `decompose`

src/graphs/utils.jl
```diff
@@ -0,0 +1,146 @@
...
+    end
+    return table_prime
+end
+
+# The subgroup of single qubit Clifford group that commutes with CPHASE.
+const Z_COMMUTATION_SUBGROUP = [CliffordOperator(x, 1) for x in [sId1(1), sZ(1), sPhase(1), sInvPhase(1)]]
+
+# Stabilizers needed to convert between equivalent states to satisfy the commutation constraint (see cphase_isolated)
+# NOTE: Luckily, we don't need the whole stabilizer subgroup which is quite big, but only a selected few.
+const PAULI_STABILIZERS = Dict(
```
Krastanov: docstrings, not comments

src/graphs/utils.jl
```diff
@@ -0,0 +1,146 @@
...
+            table[(true, U1, U2)] = apply_cphase_isolated(true, U1, U2)
+            table[(false, U1, U2)] = apply_cphase_isolated(false, U1, U2)
+        end
+    end
+    return table
+end
+
+# Tables below are precompiled
+const IP_SQRTX_DECOMPOSITION_TABLE = decompose(sId1(1), [sInvPhase(1), sSQRTX(1)])
+const ISOLATED_CPHASE_TABLE = gen_isolated_cphase_table()
```
Krastanov: docstring that explains what each table is for

src/graphs/utils.jl
```diff
@@ -0,0 +1,146 @@
...
+    end
+    vops(g)[1], vops(g)[2] = U1, U2
+    return g
+end
+
+# Convert the graphstate to stabilizer state and perform gate there.
+# The resulting state satisfy the following constraint:
+# - Let U1, U2 be VOPs of the initial state, U1', U2' be the VOPs of the state after CPHASE
+# - We guarantee U1' ∈ Z_COMMUTATION_SUBGROUP if U1 ∈ Z_COMMUTATION_SUBGROUP. Same for U2, U2'.
+function apply_cphase_isolated(connected::Bool, U1::SingleQubitOperator, U2::SingleQubitOperator)
```
Krastanov: - Docstring, not comment
- Does this exist solely for testing purposes?

LEXUGE:
They are needed because the simulation scheme is essentially:
1. try to remove VOPs that is non-commuting with CPHASE out using local graph complementation
2. if not possible then try to apply the isolated lookup table

So (1) will need us to decompose VOPs into generators where local graph operations can remove them one-by-one.
(2) needs a lookup table.

Regarding performance, are you worried because dictionaries are keyed by `SingleQubitOperator`? I think since we only have 24 of them, we could just make the decomposition table an array and follow a convention on how to number each (as done in the paper). This is gonna be the same for the CPHASE lookup table. So that should eliminate performance concern?

However, isn't hashmap lookup supposed to be O(1) on avg? I don't quite follow the n log n part.

src/graphs/utils.jl
LEXUGE: `gen_isolated_cphase_table` calls this to compute the result for each entries in the lookup table.

src/graphs/utils.jl
LEXUGE: also realized this should not be called `PAULI_STABILIZERS` as `Hadamard` is not pauli.

src/graphs/utils.jl
Krastanov: noted!

LEXUGE:
@Krastanov 
The default `hash` function only relies on `objectId` through reflection which performs poor in this nano-second critical situation compared to hashing all fields. I have made a small change that now hashes the object by using field only.

The performance is now only 2x slower compared to array-based optimal solution (~12ns):
```julia
julia> @btime QuantumClifford.ISOLATED_CPHASE_TABLE[(true, SingleQubitOperator(sHadamard(1)), SingleQubitOperator(sX(1)))]
  26.536 ns (0 allocations: 0 bytes)

julia> @btime QuantumClifford.IP_SQRTX_DECOMPOSITION_TABLE[SingleQubitOperator(sX(1))]
  13.864 ns (0 allocations: 0 bytes)
```

Krastanov:
Sounds good, no need to worry about further performance optimizations. I agree it is more important to implement the functionality before worrying about performance, I just want to make sure it requires just a refactor, not a complete rewrite. For that purpose, could you update/expand the documentation explaining what each component is and why it is necessary -- I do not feel I have a clear view of how this is going to be used yet

LEXUGE:
Sure, in that sense I think it might be better to have the full PR ready. And we could review that way. Otherwise I might be repeating the paper.

LEXUGE:
```julia
julia> s = random_stabilizer(10)
- X_XZ_Y_Z_X
+ __ZYYZZYZ_
+ X_XYXX__XX
- ZZXYX_Y_Y_
- ZXY_XZYZ_Y
- XZYZZZ_Z_X
- YXYYXZ_XXY
+ YZXXXX_ZYY
+ Z_ZYZ_Z_XZ
+ ___Z__XZYX

julia> canonicalize!(Stabilizer(apply!(QuantumClifford.GraphState(copy(s)), sCPHASE(5,7)))) == canonicalize!(apply!(copy(s), sCPHASE(5, 7)))
true

julia> canonicalize!(Stabilizer(apply!(QuantumClifford.GraphState(copy(s)), sCPHASE(1,7)))) == canonicalize!(apply!(copy(s), sCPHASE(1, 7)))
true
```

Got the CPHASE implementation working. Quite rough though, but should help to illustrate the usage etc.

LEXUGE:
```julia
julia> @benchmark apply!(g, gate) setup=begin g=QuantumClifford.GraphState(random_stabilizer(100)); gate=sCPHASE(3,7); end
BenchmarkTools.Trial: 112 samples with 815 evaluations per sample.
 Range (min … max):  158.643 ns … 105.853 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):      65.257 μs               ┊ GC (median):    0.00%
 Time  (mean ± σ):    53.387 μs ±  31.072 μs  ┊ GC (mean ± σ):  0.05% ± 0.26%

  █                                                              
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▃▁▁▃▁▁▃▃▄▄▃▅▃▃▅▅▆▄▄▆▆▄▃▃▃▁▄▄▃▁▁▃▁▁▁▁▃ ▃
  159 ns           Histogram: frequency by time          102 μs <

 Memory estimate: 248 bytes, allocs estimate: 2.

julia> @benchmark apply!(s, gate) setup=begin s=random_stabilizer(100); gate=sCPHASE(3,7); end
BenchmarkTools.Trial: 6228 samples with 267 evaluations per sample.
 Range (min … max):  290.397 ns … 638.434 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     300.341 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):   304.077 ns ±  17.696 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

       ▃▇█▄                                                      
  ▂▂▂▃▅█████▆▅▆▅▅▄▃▃▄▃▃▃▂▂▃▂▂▂▂▂▂▂▂▂▂▂▂▂▁▂▂▂▂▂▂▁▁▂▂▁▂▁▁▁▂▂▁▁▁▁▂ ▃
  290 ns           Histogram: frequency by time          364 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

This is the current performance profile: for favorable cases (which not `all random_stabilizer`s are), we are leading the stabilizer by 2x

LEXUGE:
@Krastanov I will update after some cleaning my draft and tests. And we could have a zoom call to discuss question?

LEXUGE:
@Krastanov  I have updated the docstring and comments, and also fixed a bug in the implementation (added a test for that).

Krastanov:
This looks great! I made some very minor clarifications to the docs as I was reading through them ans some formatting changes. Two things that come to mind that we should address in future PRs

- There is a lot of new internal functionality you have created. I believe we should move this to a submodule. No need to do that in the current PR or in your upcoming PR on measurements, but we should probably have it done afterwards. Just a single commit wrapping your current graphs.jl in `module GraphSim ... end` (or another appropriate module name) and fixing up the imports/exports.

- You have done a lot of implicit benchmarking here, and there is a lot of conversion between CliffordOperator / SingleQubitOperator / others. Currently I it is not clear to me (1) what are all the options that were tried (2) what better options have not been tried because they would require unrelated development (3) what new APIs/methods/etc needs to be developed to make things clearer. Some dev documentation (e.g. in a src/graphs/README.md) would be valuable.

Could you plan to make these two additions in your future rounds of work?

Krastanov:
The windows failure is a known issue with the python install for stim. Unrelated to this work

LEXUGE:
@Krastanov Thanks for spotting the bug. I have updated the test and fixed it.

LEXUGE:
Add a minor performance improvement commit. Now it's allocation-free!

```julia
julia> @benchmark apply!(g, sCPHASE(2, 4)) setup=begin g = GraphState(random_stabilizer(100)) end
BenchmarkTools.Trial: 67 samples with 944 evaluations per sample.
 Range (min … max):   96.700 ns … 429.127 μs  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     795.769 ns               ┊ GC (median):    0.00%
 Time  (mean ± σ):    79.241 μs ± 134.651 μs  ┊ GC (mean ± σ):  0.03% ± 0.05%

  █                                                              
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▅▅▆▅▁▁▁▅▁▁▁▅▁▁▁▁▁▁▁▁▅▁▅▁▆▅▁▅▅▁▁▁▁▁▁▅▅▁▅▁▁▁▁▁▆ ▁
  96.7 ns       Histogram: log(frequency) by time        419 μs <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

(I will document these)

PR #559, opened 2025-07-21
DO NOT MERGE: investigating an Oscar/OpenCL crash

Krastanov:
see #519


Krastanov:
@Hamiltonian-Action , do you see anything obvious missing here compared to the environment in which you were seeing crashes?

PR #564, opened 2025-07-25
D-dimensional and Double Homological Product codes via Oscar's Homological Algebra

Fe-r-oz:
Building on #534, this PR implements the `D`-dimensional Homological Product codes presented in [arxiv:2009.11790](https://arxiv.org/pdf/2009.11790). Using three classical parity check matrices, the paper presents 3D quantum code via homological product. The 3D homological product codes mentioned in reference are more general and supports (3,4)-classical LDPC codes as well.

-  2D Homological product codes from quasi-cyclic classical codes of [arxiv: /2407.18490](https://arxiv.org/pdf/2407.18490) (Table III)
- 3D Homological product of [arxiv: 2009.11790](https://arxiv.org/pdf/2009.11790)
- Homological product of classical codes.
 
In addition, this PR implements Double Homological Product codes from [arxiv:1805.09271](https://arxiv.org/pdf/1805.09271) (Table I)


ext/QuantumCliffordOscarExt/homological_product_codes.jl
```diff
@@ -0,0 +1,318 @@
...
+    # δ̌1 = [δ̃₀ ⊗ I | I ⊗ δ̃₋₁ᵀ]
+    δ̌₁ = hcat(
+        kronecker_product(δ̃₀, identity_matrix(R, ñ₋₁)),
+        kronecker_product(identity_matrix(R, ñ₋₁), transpose(δ̃₋₁))
+    )
+    @assert iszero(δ̌₁*δ̌₀)
+    return fq_to_int(δ̌₋₂), fq_to_int(δ̌₋₁), fq_to_int(δ̌₀), fq_to_int(δ̌₁)
+end
+
+function parity_matrix_xz(c::DoubleHomologicalProductCode)
```
nzy1997: The `parity_matrix_xz` function isn't always necessary for all CSS codes. Here you can directly use:
```julia
parity_matrix_x(c::DoubleHomologicalProductCode) = boundary_maps(c)[2]
```

ext/QuantumCliffordOscarExt/homological_product_codes.jl
```diff
@@ -0,0 +1,318 @@
+"""Convert a polynomial parity-check matrix or generator check matrix to binary quasi-cyclic form."""
+function _circulant_matrix_from_quasi_cyclic_polynomial_matrix(H::MatSpaceElem, l::Int)
```
nzy1997: Consider placing this function after the code structure definitions.

ext/QuantumCliffordOscarExt/QuantumCliffordOscarExt.jl
```diff
@@ -25,11 +26,12 @@ import QECCore: AbstractECC, CSS, RepCode, AbstractCSSCode,
     metacheck_matrix_x, metacheck_matrix_z, metacheck_matrix
 
 # exported from extension so that Documenter.jl sees them when autogenerating API lists
-export twobga_from_direct_product, twobga_from_fp_group, DDimensionalSurfaceCode, DDimensionalToricCode, boundary_maps
+export twobga_from_direct_product, twobga_from_fp_group, DDimensionalSurfaceCode, DDimensionalToricCode, boundary_maps, quasi_cyclic_code
```
nzy1997: Where is this `quasi_cyclic_code` function defined or used?

Fe-r-oz: The `quasi_cyclic_code` was part of the earlier API; it was simply a wrapper function to convert a polynomial or generator polynomial matrix into a circulant matrix. I've renamed it to `_circulant_matrix_from_quasi_cyclic_polynomial_matrix` to better reflect its purpose for internal use. Now the polynomial matrix can be passed directly to `HomologicalProductCode`, eliminating the manual conversion step of creating `H = quasi_cyclic_code(H_poly, l)` before passing it to the code. Table III of the [reference](https://arxiv.org/pdf/2407.18490) utilize the polynomial parity check matrix as input to the Homological product code.

nzy1997 (commented):
The code looks good! Any idea why the CI failed? Maybe try rerunning it?

test/test_ecc_base.jl
```diff
@@ -241,6 +241,7 @@ p₃ = one(R) + x
 # Add some codes that require Oscar, hence do not work on Windows
 
 test_twobga_codes = []
+test_homological_product_codes = []
```
Krastanov, line 244: Third place with the same issue @Fe-r-oz @nzy1997 The codes added here are not actually tested at all. See for instance how test_twobga_codes is used elsewhere in the code base. This comes from discussion in #594

Fe-r-oz: Thank you for your feedback. You are right that the new codes should be integrated into the main test suite in `test_ecc_base.jl` to match the pattern used by all other codes including `test_twobga_codes` and ensure they are tested with the full spectrum of checks which includes syndrome measurements, and decoding.  I will add them to the test suite in `test_ecc_base.jl`. 

For context, the stabilizer correctness properties of homological product codes in this PR were verified in the following file but I agree that integration into the main suite is the necessary next step.

https://github.com/QuantumSavory/QuantumClifford.jl/blob/11dd3e53ebd2f2cb5e0cd1653839c0ea84cb5c1a/test/test_ecc_d_dimensional_codes.jl#L133
https://github.com/QuantumSavory/QuantumClifford.jl/blob/11dd3e53ebd2f2cb5e0cd1653839c0ea84cb5c1a/test/test_ecc_d_dimensional_codes.jl#L221

PR #567, opened 2025-07-29
Bugfix: improper tessellation of the ndrange in KAExt mul_leftright -- Benchmark: Tableau-Pauli

Hamiltonian-Action:
I wanted Tableau-Pauli benchmark results to compare against for the canonicalisation implementation and I discovered a subtle bug that had been hidden away due to relying on a large default batch size. This patch resolves the issue, enhances the randomised testing in order to catch similar bugs should they arise in the future, as well as adding Tableau-Pauli multiplication to the benchmark suite.
In short, `unsafe_indices` is silly and induces a discrepancy between the physical thread count and the logical thread count[1] but it is nevertheless obligatory in order to ensure that the reductions are valid. Hate to give it to them but NVIDIA once again proves that they have the least painful API to work with for no shortage of incredibly solid reasons.

Note: There seems to be some an unintended change to `src/randoms.jl` as well? I guess it must have been poorly indented and my git hooks may have inadvertently modified it for consistency. 

[1] As an example, imagine `workgroupsize = (256, 1), ndrange = (511, 1)`. The number of spawned threads will be `512` and the excess straggler behaves perfectly normally just as any other, i.e: this is no different from `workgroupsize = (256, 1), ndrange = (512, 1)`, yet querying the ndrange from inside the kernel will still actually read `511`.

### DIFF
--- a/test/KernelAbstractions/implementation/definitions.jl
+++ b/test/KernelAbstractions/implementation/definitions.jl
@@ -8,6 +8,6 @@ const max_rows = 1024
 # Keep it reasonable so that local testing remains accessible.
 const round_count = 16
 # Correctness should be independent of parameter values.
-# The omission of the const specifier is intentional, overriden in OpenCL.
+# The omission of the const specifier is intentional, overridden in OpenCL.
 block_sizes = rand(1:256, round_count)
 const batch_sizes = rand(1:256, round_count)


Krastanov:
The CUDA tests have not executed yet because of issues with the CUDA runner. It is restarted now and there is extra logging added to it to try to figure out why it is repeatedly crashing.

Let's keep the OpenCL tests but if they continue being an issue we can just stop blocking merges on failures of that particular CI. Nonetheless, keeping it around would be a useful way to track how the ecosystem is evolving.

I will merge this once the CUDA tests pass.

Krastanov:
Thanks for investigating the issue, this improvement is much appreciated. The rigor of your test suite makes reviewing much easier.

PR #569, opened 2025-08-02
applybranches for symbolic single qubit reset measurement ops (sMRZ, sMRX, sMRY)

brohan-byte:
-> Created a new abstract type called AbstractResetMeasurements (a subtype of AbstractOperation) and assigned sMRZ, sMRX and sMRY to it (they were previously a subtype of AbstractOperation). 

-> modified allops.md to reflect this change

-> wrote an applybranches function for ops of the new abstract type AbstractResetMeasurements(which includes sMRZ, sMRX, sMRY)

-> defined a helper constant for applybranches that maps each reset measurement function (such as sMRZ) to the specific functions it needs, so applybranches looks up the appropriate measurement, correction, and projector based on the input reset measurement op.


docs/src/allops.md
```diff
@@ -56,6 +56,11 @@ td > code {
 |` │   ├─ sMY                            `|❌ | n²| [`projectY!`](@ref)    |
 |` │   └─ sMZ                            `|❌ | n²| [`projectZ!`](@ref)    |
 |` │                                     `|  |   |                        |
+|` ├─ AbstractResetMeasurement           `|  |   |                        |
```
Krastanov, line 59: just mark the last column as well, so that a reader can get a sense of how they are implemented

src/classical_register.jl
```diff
@@ -166,6 +166,39 @@ function applybranches(s::Register, op::AbstractMeasurement; max_order=1)
     new_branches
 end
 
+
+
+#helper constant for the AbstractResetMeasurement applybranches 
+const reset_mappings = Dict(
```
Krastanov: let's redo this in a way that is guaranteed to happen at compilation time

test/test_noisycircuits.jl
```diff
@@ -224,6 +224,41 @@
...
+            op2 = sMRX(1,1)
+            verify2 = VerifyOp(state2, [1])
+            pet2 = petrajectories(register2, [op2,verify2])
+            @test pet2[false_success_stat] == 1
+            @test pet2[true_success_stat] == 0
+            @test pet2[failure_stat] == 0
+
+            #checks probabilstic case to see if the phase of measurement anticommuting stabilizer is the same in both branches
+            ghzState = S"XXX ZZI IZZ"
+            reg = Register(ghzState, [0])
```
Krastanov:
```suggestion
            ghz_state = S"XXX ZZI IZZ"
            reg = Register(ghz_state, [0])
```

src/classical_register.jl
```diff
@@ -166,6 +166,39 @@ function applybranches(s::Register, op::AbstractMeasurement; max_order=1)
...
+    sMRZ => [sMZ, sX, projectZ!],
+    sMRX => [sMX, sZ, projectX!],
+    sMRY => [sMY, sZ, projectY!]
+)
```
Krastanov: `_reset_mappings(::Type{sMRZ}) = (sMZ, sX, projectZ!)`

src/classical_register.jl
```diff
@@ -166,6 +166,38 @@ function applybranches(s::Register, op::AbstractMeasurement; max_order=1)
     new_branches
 end
 
+
+
+#helper dispatch functions for the AbstractResetMeasurement applybranches 
+_reset_mappings(::Type{sMRZ}) = (sMZ, sX, projectZ!)
```
Krastanov: first one seems unused

src/classical_register.jl
```diff
@@ -166,6 +166,38 @@ function applybranches(s::Register, op::AbstractMeasurement; max_order=1)
...
+_reset_mappings(::Type{sMRX}) = (sMX, sZ, projectX!)
+_reset_mappings(::Type{sMRY}) = (sMY, sZ, projectY!)
+
+#applybranches for reset measurements such as sMRZ, sMRX and sMRY
+function applybranches(r::Register, op::AbstractResetMeasurement;max_order=1)
+    operations = _reset_mappings(typeof(op))
```
Krastanov:
```suggestion
    flipop, measurementfunc = _reset_mappings(typeof(op))
```

docs/src/allops.md
```diff
@@ -56,6 +56,11 @@ td > code {
...
+|` │   ├─ sMRX                           `|❌ | n²| [`projectX!`](@ref) , [`sZ`](@ref) , [`sMX`](@ref)                     |
+|` │   ├─ sMRY                           `|❌ | n²| [`projectY!`](@ref) , [`sZ`](@ref) , [`sMY`](@ref)                     |
+|` │   └─ sMRZ                           `|❌ | n²| [`projectZ!`](@ref) , [`sX`](@ref) , [`sMZ`](@ref)                  |
```
Krastanov: removing the sX and sMX and so on, and keeping just project looks stylistically a bit more consistent to me

src/classical_register.jl
```diff
@@ -166,6 +166,38 @@ function applybranches(s::Register, op::AbstractMeasurement; max_order=1)
...
+        phases(stabilizerview(s2.stab))[anticom] = 0x02
+        s2.bits[op.bit] = true
+        apply!(s2, flipOp(op.qubit))
+        push!(new_branches, (s1,continue_stat,1/2,0))
+        push!(new_branches, (s2,continue_stat,1/2,0))
+    else
+        r.bits[op.bit] = res==0x02
+        if(res==0x02)
+            apply!(r, flipOp(op.qubit))
+            r.bits[op.bit] = false
```
GuusAvis, line 195: 1. Didn't this bit already get set on line 192? Why is it set again here?
2. Moreover, don't the two lines (192 and 195) together assure that the bit is always false regardless of what the result is?

test/test_noisycircuits.jl
```diff
@@ -224,6 +224,41 @@
...
 
+        @testset "Symbolic Reset Measurements" begin
+            using QuantumClifford:tensor
+            #test for checking reset capabilities
+            state1 = S"-Z"
+            register1 = Register(state1, [0])
+            op1 = sMRZ(1,1)
+            verify1 = VerifyOp(state1, [1])
+            pet1 = petrajectories(register1, [op1,verify1])
+            @test pet1[false_success_stat] == 1
```
GuusAvis, line 235: It is not clear to me why we expect a false success status here, as there are no errors shouldn't it always be a true success (or true failure, depending on what we try to verify)?

test/test_noisycircuits.jl
```diff
@@ -224,6 +224,41 @@
...
+            #checks probabilstic case to see if the phase of measurement anticommuting stabilizer is the same in both branches
+            ghz_state = S"XXX ZZI IZZ"
+            reg = Register(ghz_state, [0])
+            branches = applybranches(reg, sMRZ(1,1))
+            stab1 = stabilizerview(branches[2][1].stab)
+            stab2 = stabilizerview(branches[1][1].stab)
+            canonicalize_rref!(stab1)
+            canonicalize_rref!(stab2)
+            @test phases(stab1)[end] == 0x00
+            @test phases(stab2)[end] == 0x00
```
GuusAvis, line 257: How do you know that it is the last phase that should be the same? Also, wouldn't it be more complete/less error prone to check the entire output states? I.e., check that stab1 and stab2 are the same state as `S"ZII, IZI, IIZ"` and `S"ZII, -IZI, -IIZ"`?

src/classical_register.jl
brohan-byte: You're completely right! This went completely over my head, I apologize. I'm in the process of submitting another PR to fix this

test/test_noisycircuits.jl
brohan-byte: I think the logic I went with in that test is something like:
1) the original qubit to be measured has a -1 eigenvalue
2) using sMRZ on it would both measure the value and also flip the qubit to a +1 eigenvalue state
3) since verify op works on initial and final states, and the final state is flipped(+Z) and the initial state is a (-Z) you would expect a false success.

I think since this is a bit unclear, I should probably include some comments in there. Do let me know if I've made some sort of logical error here!

test/test_noisycircuits.jl
brohan-byte: I think I hard-coded i.e. ran the outputs and saw that the last stabilizer would be the anti-commuting Z__ stabilizer and wrote my test according to that. But you're right, checking it against the states as a whole is a better way of doing this and I'll be making that change!

test/test_noisycircuits.jl
GuusAvis: I do agree with your assessment that the state should be in the +1 eigenstate after the sMRZ, and that hence the verification should "fail". But wouldn't that be a `failure_stat`? The documentation of `false_success_state` says
>Returned by [applywstatus!](https://qc.quantumsavory.org/stable/API/#QuantumClifford.applywstatus!-Tuple{Any,%20Any}) if the circuit reports a success, but it is a false positive (i.e., there was an undetected error).

I don't think we expect a false positive here, just a straight-out negative. After all, there are no errors and hence also no undetected errors. Hence I'm a bit confused.

brohan-byte: I'm actually a bit confused as to how an output gets flagged as a false_success as well. If I'm understanding how verifyop works correctly
image
if theres any stabillizer in the final state which is not equal to that in the good_state then it immediately gets flagged as a false_success.


When I run this snippet of code:
image
image
this dict seems to verify that fact. I'm actually not sure as to when the failure_stat is even used since it doesn't seem to be written into verify_op. 

If I'm not wrong(feel free to correct me here) the way the false_success_stat has been implemented is different to what the definition you've provided says. I think it just flags any mismatch between initial and final states as a false_success. 

Not sure what to fix/change here, so please let me know if you have any insight!

GuusAvis: Thanks for explaining that!

The way I now understand it is that usually `VerifyOp` should only be used to compare a state to the state that you expect to have - in that case, only an error (i.e., an unexpected thing happening) can cause a mismatch. Hence, a "true failure" would never occur.

Seen from that perspective we are just abusing `VerifyOp` a little bit in this test, as we are using it to verify that the output state is _different_ than some reference state, while it is mean to be used to check if an output state is _the same_ as a reference state. And this results in the counterintuitive status being returned.

For the test, I would suggest changing the use of `VerifyOp` to correspond with its intended use and hence look less confusing, making the test easier to understand. This would be done by using S"+Z" as the `good_state` in `VerifyOp`, and then asserting you get a true success status. Anyhow I think it is in general better to do a positive verification in tests, i.e., explicitly testing that the output is exactly equal to the expected output, rather than a negative verification, i.e., testing that the output is different from one specific unexpected output. For instance, now the test would also succeed if the reset operation implemented a Hadamard instead of an X.

@Krastanov given that me and Rohan both got a bit confused about `VerifyOp`, would it make sense if we added an explicit discussion of the return status to the documentation (of either `VerifyOp` or its `applywstatus!` method)?

Krastanov: > @Krastanov given that me and Rohan both got a bit confused about `VerifyOp`, would it make sense if we added an explicit discussion of the return status to the documentation (of either `VerifyOp` or its `applywstatus!` method)?

Any clarification on the docs would be highly valuable. And if you do not like the ergonomics behind VerifyOp, we can change its meaning and semantics to be more natural and easy to use -- that is very much within scope of this whole exercise

brohan-byte: Thank you for the valuable insights @GuusAvis and @Krastanov . I'll be compiling these fixes into a new PR and submitting it soon!

PR #561, opened 2025-07-23
apply_right

sagnikpal2004:
Adds the `apply_right!` function into QuantumClifford to apply a Clifford Operator (dense or symbolic) to the right of another clifford operator, modifying it inplace.

- [x] The code is properly formatted and commented.
- [x] Substantial new functionality is documented within the docs.
- [x] All new functionality is tested.
- [x] All of the automated tests on github pass.

Changelog:
- added `apply_right!` that applies a clifford operator to the right of another

Notes:
- The symbolic operations make use of three basic operations: `rowswap!(r1, r2)`, `mul_right_ignore_anticommute!(r1, r2)` and `phase_flip!(r1)`. There could be macros built that definte the operations from these basic ones.

Closes https://github.com/QuantumSavory/QuantumClifford.jl/issues/555
Is a dependency to PR https://github.com/QuantumSavory/QuantumClifford.jl/pull/551

Adaptation of Stim's [`prepend_`](https://github.com/quantumlib/Stim/blob/3ed7ffc7bc00e999dd950a3527a41d636b9ff37b/src/stim/stabilizers/tableau_specialized_prepend.inl)

PR #573, opened 2025-08-06
`CliffordOperator` constructor from `PauliOperator`

sagnikpal2004:
This pull request adds support for constructing a `CliffordOperator` directly from a `PauliOperator`. It also introduces corresponding tests to ensure the correctness of the new constructor.

### Changelog
* Added a constructor to `CliffordOperator` that creates a Clifford operator from a `PauliOperator`
* Added an `@inline` getter for the `phases` field of a `CliffordOperator`.

### Testing
* Added a test set to verify that applying a Pauli operator via the new constructor produces the same result as applying the Pauli operator directly.

### Notes
* This consrtuctor is required to test #561, however it may not have any other practical uses, since any applications of `PauliOperators` can be done directly. Need advice if this should simply be left in the test file, or incorporated into the library
* A `PauliOperator` constructor from a `CliffordOperator` is also possible, and I can create a PR for it if deemed valueable for the library.

Dependency of #561


test/test_cliff.jl
```diff
@@ -23,6 +23,13 @@
         @testset "Constructors" begin
             @test_throws DimensionMismatch CliffordOperator(T"X")
         end
+        @testset "Constructor from PauliOperator" begin
+            for n in test_sizes
+                l = random_clifford(n)
+                pauli = random_pauli(n)
+                @test isequal(apply!(copy(l), pauli; phases=true), apply!(l, CliffordOperator(pauli); phases=true))
```
Krastanov:
```suggestion
                @test apply!(copy(l), pauli; phases=true) == apply!(l, CliffordOperator(pauli); phases=true)
```

Krastanov (approved):
could you make the small change to the tests for legibility

could you add a mention of this to the changelog

otherwise looks good for merge

src/dense_cliffords.jl
```diff
@@ -63,6 +63,11 @@ end
 CliffordOperator(op::CliffordOperator) = op
 CliffordOperator(paulis::AbstractVector{<:PauliOperator}) = CliffordOperator(Tableau(paulis))
 CliffordOperator(destab::Destabilizer) = CliffordOperator(tab(destab))
+function CliffordOperator(pauli::PauliOperator)
+    res = one(CliffordOperator, nqubits(pauli))
+    tab(res).phases = 0x02 * comm(pauli, tab(res))
```
Krastanov: this should be an inplace modification of the existing list, not a reasignment

broadcast can be a good way to do it

Krastanov:
```suggestion
    tab(res).phases .= 0x02 .* comm(pauli, tab(res))
```

sagnikpal2004:
@Fe-r-oz Thanks, I'll definitely keep that in mind next time!

PR #436, opened 2024-11-28
Honeycomb color code via Hecke's Group Algebra

Fe-r-oz:
This PR provides the group algebraic implementation of honeycomb color codes.

PR #574, opened 2025-08-08
Generalized Toric code on twisted tori via Oscar's Laurent polynomials

Fe-r-oz:
This PR implements Generalized Toric code from [reference](https://arxiv.org/pdf/2503.03827).

API:

```jldoctest
julia> R, (x,y) = laurent_polynomial_ring(GF(2), [:x, :y]);

julia> f = 1 + x;

julia> g = 1 + y;

julia> α1 = (0, 6);

julia> α2 = (3, 3);

julia> c = GeneralizedToricCode(f, g, α1, α2);

julia> code_n(c), code_k(c)
(36, 2)

julia> R, (x,y) = laurent_polynomial_ring(GF(2), [:x, :y]);

julia> f = 1 + x + x^-1*y^-3;

julia> g = 1 + y + x^3*y^-1;

julia> α1 = (0, 66);

julia> α2 = (3, 18);

julia> c = GeneralizedToricCode(f, g, α1, α2);

julia> code_n(c), code_k(c)
(396, 8)
```

- [x] We recently started enforcing formatting checks. If formatting issues are reported in the new code you have written, please correct them


test/test_ecc_generalized_toric_codes.jl
```diff
@@ -0,0 +1,137 @@
...
+    @static if !Sys.iswindows() && Sys.ARCH == :x86_64 && VERSION >= v"1.11"
+        using Oscar
+        using QECCore
+        import HiGHS
+        import JuMP
+        using Nemo: matrix, GF
+        using QuantumClifford: stab_looks_good, stab_to_gf2
+        using QuantumClifford.ECC
+
+        @testset "Generalized Bicycle Codes from Tables 1-IV of https://arxiv.org/abs/2503.03827" begin
```
nzy1997, line 11: Great job covering these cases from the literature!

test/test_ecc_generalized_toric_codes.jl
```diff
@@ -0,0 +1,137 @@
...
+                (112, 6 , 1 + x + x^-1*y^2 , 1 + y + x^-2*y^-1, (0,  7),  (8 ,   2)),
+                (114, 4 , 1 + x + x^-3*y   , 1 + y + x^-5     , (0,  3),  (19,   1)),
+                (120, 8 , 1 + x + x^-2*y   , 1 + y + x*y^2    , (0, 10),  (6 ,   4)),
+                (124, 10, 1 + x + x^-1*y^2 , 1 + y + x^-2*y^-1, (0, 31),  (2 , -12)),
+                (126, 12, 1 + x + x^-1*y^-2, 1 + y + x*y^-1   , (0,  9),  (7 ,   3)),
+                (132, 4 , 1 + x + y^-2     , 1 + y + x^-2     , (0, 33),  (2 ,  -7)),
+                (138, 4 , 1 + x + x^-3*y   , 1 + y + x^3*y^2  , (0,  3),  (23,   2)),
+                (140, 6 , 1 + x + x^-2     , 1 + y + x^-2*y^2 , (0,  7),  (10,   1)),
+                (144, 12, 1 + x + x^-1*y^-3, 1 + y + x^3*y^-1 , (0, 12),  (6 ,   0)),
+                (144, 12, 1 + x + x^-1*y^3 , 1 + y + x^3*y^-1 , (0, 12),  (6 ,   0)),
```
nzy1997, line 50: Consider adding some equivalence checks, like this code corresponds to the [[144,12,12]] bb code and color code you mentioned in the docstring.

src/ecc/ECC.jl
```diff
@@ -40,6 +40,7 @@ export parity_checks, parity_matrix_x, parity_matrix_z, iscss,
     DDimensionalSurfaceCode, DDimensionalToricCode, boundary_maps,
     GeneralizedCirculantBivariateBicycle, GeneralizedHyperGraphProductCode,
     GeneralizedBicycleCode, ExtendedGeneralizedBicycleCode,
+    GeneralizedToricCode, max_xy_exponents,
```
nzy1997: I’m wondering if `max_xy_exponents` should be exported—do you think users will need it?

Fe-r-oz: Exporting `max_xy_exponents` would be valuable for users in the future. Its primary purpose is to enable the efficient computation of the maximum logical dimension for a generalized toric code, which is one of the central result of the [reference](https://arxiv.org/pdf/2503.03827). Instead of requiring users to compute the rank of large, sparse parity-check matrices, this method allows them to determine the code dimension algebraically.

After computing the Gröbner basis for the ideal generated by the Laurent polynomials *f* and *g*, the `max_xy_exponents` function extracts the maximum exponents for the *x* and *y* variables from the generating polynomials. This is a crucial step, as these maximum exponents define all possible independent monomials (Eq. 25 of [reference](https://arxiv.org/pdf/2503.03827)). Without this function, users would have to manually inspect the Gröbner basis results to find these values, which is probably inconvenient. The total number of independent monomials determines the maximum code dimension (which is twice this number). I have included examples in the documentation demonstrating this use case after gröbner basis computation.

There is some future work to enhance the Gröbner basis computation itself and polish the current documentation (such as implementing the algorithm in [Appendix C](https://arxiv.org/pdf/2503.03827). The Groebner basis improvements can be addressed in a subsequent PR; allowing this PR to focus on code construction. We do not export the `max_xy_exponents` yet and make sure everything fully ready w.r.t to computing the maximum logical dimension --meaning completing the future work to enhance Gröbner basis computation via algorithm and documentation.

test/test_ecc_generalized_toric_codes.jl
Fe-r-oz: The Gaussian elimination process used for canonicalization is sensitive to column permutations. As discussed in [here](https://github.com/QuantumSavory/QuantumClifford.jl/pull/436#issuecomment-3047852890) and [here](https://github.com/QuantumSavory/QuantumClifford.jl/pull/534#issue-3187838563), constructing equivalent stabilizer codes via drastically different methods and realizations can result in column permutations. This leads to distinct canonical forms, causing a naive stabilizer/matrix equivalence check to fail for logically identical codes. 

To solve this, we can always test the invariant properties that are impervious to column permutations.  I have added a new test set that demonstrates this by testing the Appendix B example. The new test verifies that the check matrices share the same essential properties, including the column weights and row weights of the `X`-and `Z`-type check matrices, and the rank of the `X`-and `Z`-type matrices, which is used in determining the number of logically encoded qubits.

Fe-r-oz: And we can test that the resulting `[[144, 12, 12]]` code has row weights `3` and column weights `6` which is the property shared across different code realisations despite the column permutations. I have added a test set for this property test as well.

Fe-r-oz:
Thank you for your suggestions! Improved test coverage and documentation.

nzy1997 (commented):
Thanks for adding the new test. As long as we've confirmed the failing tests are unrelated, this is good to merge.

Issue #532, opened 2025-06-29
Cleaner practices: Cataloguing the menagerie of inlined throw strings.


Hamiltonian-Action:
## Issue ##

This is not directly a user-facing problem (during correct usage) but rather endemic to the codebase itself. There appears to be some rather prolific usage of inlined throw strings which could benefit from additional structuring, be it enhanced formatting, centralisation into a single file, or establishing a consistent ordering -- this last one need not be considered too thoroughly if the number of potential throwing scenarios is not too large.

## Solution ##

I would propose the introduction of a dedicated file, say `throws.jl` , and shifting the various strings currently in use towards it, either gradually as different sections get touch-up work or via some dedicated effort. Furthermore, there is no need to explicitly specify where the error originates from in the string itself, since the stack trace will clearly indicate so in lieu.

## Example ##

In `QuantumClifford.jl`, the function `comm!` contains the following:
```Julia
nqubits(l) == nqubits(r) || throw(DimensionMismatch(lazy"The number of qubits of the input Pauli operator and the input tableau have to match in `comm!`"))
```
Whereas in `mul_leftright.jl`, the functions `mul_left!` and `mul_right!` contain:
```Julia
nqubits(l)==nqubits(r) || throw(DimensionMismatch("The two Pauli operators should have the same length!"))
```

<br />
<br />

Under this proposal, the following definition would be provided in the dedicated file:
```Julia
const THROW_NQUBITS::LazyString = "Unable to perform the requested operation due to encountering a mismatch between the number of qubits in the provided arguments."
```
Whilst the aforementioned functions will simply read:
```Julia
nqubits(l) == nqubits(r) || throw(DimensionMismatch(THROW_NQUBITS))
```

Krastanov:
Making a global constant string that you can reuse sounds good. No need to mark it as lazy if it does not use interpolation. And for those that use interpolation you would need a function anyway, so again no need to mark it as lazy.

niranjannagumalli:
Hey @Krastanov and @Hamiltonian-Action 
I've submitted a PR which solves this issue. Could y'all please have a look and provide me any suggestions if any?
Thanks in advance.

PR #575, opened 2025-08-09
Refactor: Consolidate Inlined Throw Strings into throws.jl

niranjannagumalli:
closes issue #532

### DIFF
--- a/ext/QuantumCliffordHeckeExt/QuantumCliffordHeckeExt.jl
+++ b/ext/QuantumCliffordHeckeExt/QuantumCliffordHeckeExt.jl
@@ -32,5 +32,6 @@ include("lifted.jl")
 include("lacross.jl")
 include("lifted_product.jl")
 include("../../src/throws.jl")
+using .Throws
 
 end # module
--- a/src/nonclifford.jl
+++ b/src/nonclifford.jl
@@ -1,4 +1,3 @@
-include("throws.jl")
 """
 $(TYPEDEF)
 
--- a/src/symbolic_cliffords.jl
+++ b/src/symbolic_cliffords.jl
@@ -1,5 +1,4 @@
 using Random: AbstractRNG, GLOBAL_RNG
-include("throws.jl")
 
 THROW_COMPACT_QUBIT_OPERATOR_SHAPE_ERROR(n, operator_with_size) =
 "Set `n=$n` as a $operator_with_size being compacted (`compact=true`) has to result in a $n x $n `CliffordOperator`."
--- a/src/throws.jl
+++ b/src/throws.jl
@@ -1,3 +1,14 @@
+module Throws
+
+export THROW_BOUNDS,
+       THROW_SIZE,
+       THROW_NQUBITS,
+       THROW_INVALID_PARAMETERS_REEDMULLER,
+       THROW_INVALID_ACTION_QUBITS,
+       THROW_NO_ZERO_QUBIT,
+       THROW_MISSING_PACKAGE,
+       THROW_CHECKS_MISSING
+
 const THROW_BOUNDS =
 "Unable to perform the requested operation due to encountering a mismatch \
 between the provided index(indices) and the pertinent size(s) of the other \
@@ -11,15 +22,8 @@ const THROW_NQUBITS =
 "Unable to perform the requested operation due to encountering a mismatch \
 between the number of qubits in the provided arguments."
 
-const THROW_INVALID_XZ_COMPONENTS = 
-"`xzcomponents` should be `:split` or `:together`"
-
-
 const THROW_INVALID_PARAMETERS_REEDMULLER =
 "Invalid parameters: r must be non-negative and r ≤ m in order to valid code."
-
-const THROW_DELFOSSE_MIN_BLOCKS =
-"The number of blocks must be at least 2 to construct a valid code."
 
 const THROW_INVALID_ACTION_QUBITS =
 "The tableau and the provided operator need to act on the same number of qubits.
@@ -29,4 +33,12 @@ const THROW_NO_ZERO_QUBIT =
 "Qubit indices have to be larger than zero, 
 but you are attempting to create a gate acting on a qubit with a non-positive index. Ensure indexing always starts from 1."
 
+THROW_MISSING_PACKAGE(func, pack) =
+"You've invoked `$func` which depends on the package `$pack` but you have not installed or imported it yet.
+Immediately after you import `$pack`, the $func will be available."
 
+THROW_CHECKS_MISSING(code) =
+"Codes of the type $code do not have separate X and Z parity checks, either because they are not a CSS code
+and thus inherently do not have separate checks, or because its separate checks are not yet implemented in this library."
+
+end
--- a/ext/QuantumCliffordJuMPExt/QuantumCliffordJuMPExt.jl
+++ b/ext/QuantumCliffordJuMPExt/QuantumCliffordJuMPExt.jl
@@ -25,5 +25,6 @@ export distance
 
 include("min_distance_mixed_integer_programming.jl")
 include("../../src/throws.jl")
+using .Throws
 
 end
--- a/ext/QuantumCliffordKAExt/QuantumCliffordKAExt.jl
+++ b/ext/QuantumCliffordKAExt/QuantumCliffordKAExt.jl
@@ -8,5 +8,7 @@ include("../../src/throws.jl")
 include("utilities.jl")
 include("mul_leftright.jl")
 
+using .Throws
+
 end
 #=============================================================================#
--- a/ext/QuantumCliffordMakieExt/QuantumCliffordMakieExt.jl
+++ b/ext/QuantumCliffordMakieExt/QuantumCliffordMakieExt.jl
@@ -1,9 +1,12 @@
 module QuantumCliffordMakieExt
 
+include("../../src/throws.jl")
+
+using .Throws
 using Makie
 using QuantumClifford
 import QuantumClifford: stabilizerplot, stabilizerplot_axis
-include("../../src/throws.jl")
+
 
 # If you want to directly use heatmap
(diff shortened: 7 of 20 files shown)


Krastanov:
I will push against parts of that last suggestion. It depends on who the consumer of the error message is.

If the consumer is someone fairly knowledgeable of the internals of the package, they can be generic. When I get an error message in this package, I care more about the stacktrace than the message itself, and do not necessarily read it.

But the greatest value of the error messages is for the newbies. The perfect error message is the one that teaches the new user how to do something in a better way. Error messages should be very detailed, specific, actionable, and suggesting of possible sources of errors. They should be nicely formatted and contain context-specific data like sizes and nicely formatted types. In julia it is also possible to split the message into the proper error message and a "hint", but that is not very applicable to this discussion. Error messages also can have custom `showerror` methods to make them even more easy to understand.

The rust compiler is a good example of a tool with state of the art error messages. The SciML ecosystem also has invested a lot of time into making their error messages more helpful to new users.

Bringing the discussion back to this specific PR: plenty of the really valuable and interesting work in making the messages more useful to the user is out of scope to this PR. Nonetheless the reorganization started here is very valuable in identifying places where error messages should be standardized and put into const strings for reuse (or more probably, into functions that generate such strings containing a lot of additional context information).

Krastanov:
> Consult the documentation for further details

If an explanation is no longer than a paragraph, it should probably be part of the error message, for the sake of making onboarding non-technical users more easily.

niranjannagumalli:
@Krastanov Thanks for the suggestions. I've made the requested changes. Could you please have a look and let me know if everything looks fine?

- [x] error messages which are used only once are not moved to throws.jl
- [x] error messages which are limited to one file are declared at the beginning of the file and not moved to throws.jl
- [x] ensure that no context dependent information is lost.

Project.toml
Krastanov, line 1: Any reason for changes to this file? Avoid unrelated changes and undo this one if it is not needed.

ext/QuantumCliffordGPUExt/utils.jl
```diff
@@ -1,6 +1,8 @@
+const THROW_INVALID_CUDA_ARG = 
+"First argument to @run_cuda should be a function call"
 macro run_cuda(call, ndrange)
     # destructure the kernel call
-    Meta.isexpr(call, :call) || throw(ArgumentError("first argument to @run_cuda should be a function call"))
+    Meta.isexpr(call, :call) || throw(ArgumentError(THROW_INVALID_CUDA_ARG))
```
Krastanov, line 5: This string seems to be used in a single location. I think it makes more sense to just keep it as a literal, the way it was already set. Moving it makes things less easy to follow without significant benefit (unlike the situation where the same string is used all the time). This comment probably applies to a lot of other places.

ext/QuantumCliffordHeckeExt/QuantumCliffordHeckeExt.jl
```diff
@@ -31,5 +31,7 @@ include("types.jl")
 include("lifted.jl")
 include("lacross.jl")
 include("lifted_product.jl")
+include("../../src/throws.jl")
+using .Throws
```
Krastanov, line 35: this does not seem to be used

ext/QuantumCliffordHeckeExt/lifted_product.jl
```diff
@@ -1,3 +1,8 @@
+const THROW_REPR_NEEDS_COMMUTATIVE_ALGEBRA = 
+"The group algebra must be commutative when using a single `repr` function,\
+which is not the case here. Please specify separate `A_repr` and `B_repr` \
+instead of a single `repr`. The default choice of\
+`A_repr=right_repr_matrix, B_repr=left_repr_matrix` is frequently sufficient."
```
Krastanov, line 5: you need a whitespace a the start of lines or before the `\`, otherwise the words get linked together.

ext/QuantumCliffordJuMPExt/min_distance_mixed_integer_programming.jl
```diff
@@ -258,7 +258,7 @@ function _minimum_distance(hx, lx, opt, opt_summary, time_limit)
         elseif termination_status(model) == MOI.TIME_LIMIT
             throw(ErrorException("Model exceeded time limit"))
         else
-            throw(ErrorException("Model failed to solve: $(termination_status(model))"))
+            throw(ErrorException("Model failed to solve :$(termination_status(model))"))
```
Krastanov, line 261: this seems wrong or maybe I am misunderstanding the change?

src/throws.jl
```diff
@@ -1,3 +1,14 @@
+module Throws
+
+export THROW_BOUNDS,
+       THROW_SIZE,
+       THROW_NQUBITS,
+       THROW_INVALID_PARAMETERS_REEDMULLER,
+       THROW_INVALID_ACTION_QUBITS,
+       THROW_NO_ZERO_QUBIT,
+       THROW_MISSING_PACKAGE,
+       THROW_CHECKS_MISSING
```
Krastanov, line 10: let's not export anything

src/throws.jl
```diff
@@ -1,3 +1,14 @@
...
+
 const THROW_BOUNDS =
 "Unable to perform the requested operation due to encountering a mismatch \
 between the provided index(indices) and the pertinent size(s) of the other \
```
Krastanov, line 14: this probably should be a function that says what indices and sizes this is about

Krastanov: and given that it is always an `ArgumentError` the function should probably return directly an `ArgumentError`, not just the string in that error.

src/throws.jl
```diff
@@ -10,3 +21,24 @@ between the pertinent size(s) of the provided arguments."
 const THROW_NQUBITS =
 "Unable to perform the requested operation due to encountering a mismatch \
 between the number of qubits in the provided arguments."
```
Krastanov, line 23: should be a function as well

src/nonclifford.jl
```diff
@@ -361,18 +363,14 @@ struct PauliChannel{T,S} <: AbstractPauliChannel
...
-            n == nqubits(p[1]) == nqubits(p[2]) || throw(ArgumentError(lazy"""
-            You are attempting to construct a `PauliChannel` but have provided Pauli operators
-            that are not all of the same size (same number of qubits).
-            Please ensure that all of the Pauli operators being provided of of the same size.
-            """))
+            n == nqubits(p[1]) == nqubits(p[2]) || throw(ArgumentError("""
+You are attempting to construct a `PauliChannel` but have provided Pauli operators
+that are not all of the same size (same number of qubits).
+Please ensure that all of the Pauli operators being provided of of the same size.
+"""))
```
Krastanov, line 373: removing the new padding in front is not needed. Tripple-quote strings remove that automatically:
image

src/nonclifford.jl
```diff
@@ -42,7 +42,9 @@ end
 function GeneralizedStabilizer(state)
     n = nqubits(state)
     md = MixedDestabilizer(state)
-    rank(md)==n || throw(ArgumentError(lazy"""Attempting to convert a `Stabilizer`-like object to `GeneralizedStabilizer` object failed, because the initial state does not represent a pure state. Currently only pure states can be used to initialize a `GeneralizedStabilizer` mixture of stabilizer states."""))
+    rank(md)==n || throw(ArgumentError("""Attempting to convert a `Stabilizer`-like object to `GeneralizedStabilizer` object failed, 
+because the initial state does not represent a pure state. 
+Currently only pure states can be used to initialize a `GeneralizedStabilizer` mixture of stabilizer states."""))
```
Krastanov, line 47: Long lines are not a problem, I have nothing against them. If you want to make the long lines into a paragraph, make sure to left-pad them so that they are neatly aligned with where they start and use `\` so that they do not print with forced newlines (the printer can figure out its own wrapping, appropriate for the size of the terminal of the user).

Example of the features of `"""` strings:
image

src/symbolic_cliffords.jl
```diff
@@ -513,21 +520,21 @@ end
...
-    sMX(q, args...) = if q<=0 throw(NoZeroQubit) else new(q,args...) end
+    sMX(q, args...) = if q<=0 throw(ArgumentError(THROW_NO_ZERO_QUBIT)) else new(q,args...) end
 end
 
 """Symbolic single qubit Y measurement. See also [`Register`](@ref), [`projectYrand!`](@ref), [`sMX`](@ref), [`sMZ`](@ref)"""
 struct sMY <: AbstractMeasurement
     qubit::Int
     bit::Int
-    sMY(q, args...) = if q<=0 throw(NoZeroQubit) else new(q,args...) end
+    sMY(q, args...) = if q<=0 throw(ArgumentError(THROW_NO_ZERO_QUBIT)) else new(q,args...) end
```
Krastanov, line 530: `NoZeroQubit` or `NO_ZERO_QUBIT` seems more readable to me than `ArgumentError(THROW_NO_ZERO_QUBIT)`. Generally, if you know what error type you will be using, you can directly store that error type in your constant, instead of storing just the string.

src/throws.jl
```diff
@@ -10,3 +21,24 @@ between the pertinent size(s) of the provided arguments."
...
+"Qubit indices have to be larger than zero, 
+but you are attempting to create a gate acting on a qubit with a non-positive index. Ensure indexing always starts from 1."
+
+THROW_MISSING_PACKAGE(func, pack) =
+"You've invoked `$func` which depends on the package `$pack` but you have not installed or imported it yet.
+Immediately after you import `$pack`, the $func will be available."
+
+THROW_CHECKS_MISSING(code) =
+"Codes of the type $code do not have separate X and Z parity checks, either because they are not a CSS code
+and thus inherently do not have separate checks, or because its separate checks are not yet implemented in this library."
```
Krastanov, line 42: These probably should not have forced newlines.

PR #576, opened 2025-08-11
Implement KernelAbstractions accelerated canonicalization routines.

Hamiltonian-Action:
Well this is a bit embarrassing, I had finished up work on `canonicalize` and `canonicalize_rref` on Friday but must have forgotten to push things out and set up this draft.
There are lots of files that have been modified in this branch but most of them are merely structural and relate to code organisation, the actual meaningful changes are found in the following:

- `default_parameters.jl`
- `enumerations.jl` and `fixed_sizes.jl`
- `mutex_configuration.jl` and `mutex_management.jl`
- `bit_manipulation.jl`
- `scan_step.jl`
- `snippets.jl` (namely, the new snippets in the catalogue)
- `reductions.jl` (namely, `reduce_lexicographic_min!`)
- `canonicalize.jl` and `canonicalize_rref.jl`

I believe the implementation to be reasonably performant, be work-efficient, as well as being quite flexible for potential extensibility. However, it is also a sinful blight upon this cruel and uncaring universe due to necessitating coding around some terrible ergonomics and missing features. Test files, benchmarks, and an implementation of `canonicalize_gott` will follow in a suitable time frame.
Side note, the implementation provided can handle both left and right multiplication as well as X-first or Z-first ordering. Is there any consistent notion for what ought to be a `::T` argument and what ought to be a `::Val{T}` parameter in terms of the public-facing API? Internally they can be whatever the implementation wishes but I would prefer giving this some structured thought so as to avoid having different functions accept different signatures -- e.g: the `mul` family takes `phases::Val{B}` whereas the `canonicalize` functions only accept `phases::Bool`.

- [ ] The code is properly formatted and commented.


Krastanov:
On a cursory skim through this, it looks very impressive. Thankfully you have proven to be very attentive to details and edge cases -- it would be simply impossible for me to truly review the entirety of this code on any reasonable timescale, but thankfully I can rely on your track record, documentation, tests, and the fact that this is well compartmentalized within an extension.

To your question about ::Bool vs Val{true}/Val{false}: The use of `Val` types guarantees that any type stable/inferred/grounded (not sure which one is the correct name) code will resolve any if/else branches at compile time. On the other hand, if I used `Bool`, I need to rely on the constant propagation heuristic of the compiler, which is not a hard guarantee of the language. In testing, using Bool led to poor performance in some inner loops, so the internal API uses Val. However, while `Val` has these nice semantic guarantees, it is a bit embarrassing to use it in user-facing APIs -- it is an implementation detail related to performance leaking into what should be a nice API. So any user-facing API has a Bool that then calls the internal Val methods.

As a rule of thumb, the Val methods exist for being called in tight loops, while the Bool methods exist to be a pleasant user-facing API.

To finally answer your question: if it is public API, use Bool. If Bool and Val compile and perform equally well use Bool. Generally avoid Val unless the compiler forces you to.

It is quite possible that many of the Val methods currently in the codebase are not even necessary anymore. It is also possible that back in the Julia 1.7 era there was already a better way to structure the code without needing the Val methods to begin with, but I just could not think of it.

ext/QuantumCliffordKAExt/utilities/mutex_management.jl
```diff
@@ -0,0 +1,87 @@
...
+end
+
+# WARNING: Atomix has an API for atomic load/store, yet support is missing.
+# WARNING: KernelAbstractions does not provide any sort of memory fence.
+
+#==============================================================================
+SINFUL IMPLEMENTATION WROUGHT ABOUT BY THE FOLLY OF MANKIND
+==============================================================================#
+
+# TODO: Overhaul this entirely once support becomes available.
```
Krastanov, line 21: Once this is merged, could you post an issue to our issue tracker explaining that this implementation can potentially be drastically simplified if Atomix/KernelAbstractions gain certain features (could you enumerate them?). If there is a corresponding issue on the Atomix and KernelAbstractions issue trackers, could you link to it? If there is no such issue on their issue trackers, could you create one?

Lastly, in that same issue, could you give a **simple** example of the type of bugs that `canonicalize` might have if these fences were not used.

Krastanov:
No worries, @Hamiltonian-Action , your contributions are very valuable. Let me know if there is anything I can help with.

Hamiltonian-Action:
At long last, mine most arduous of quests resumes and I am once more crafting miraculous/heretical solutions from whatever is available in the JuliaGPU toolbox -- `KernelAbstractions` is not even nice enough to permit allocating mutable structures even if all the internal fields are just regular variables satisfying `isbits(x)`, I had to do a double take on my previous implementation since I used an array indexed via enumeration constants in order to try and mimic the desired behaviour whilst maintaining readability.

Strictly speaking, at least from a portability/validity point of view, the new code is inferior in the sense that the previous implementation would have absolutely worked with any arbitrarily large tableau whereas the workaround I just devised can only support rows with up to 64 PiB of data -- assuming `Csize_t` aliases `UInt64`. Luckily, I highly doubt this would be an issue in any of our lifetimes and the library already imposes similar restrictions elsewhere[*].

I am yet to conclude supporting the `Destabilizer`/`MixedDestabilizer` types as I am not absolutely certain of their internal layout due to the lack of thorough documentation/assertions and the fact that the constructors are somewhat obtuse, especially the latter. @Krastanov, would I be correct in stating that:
(a) `Stabilizer` on `n` qubits should always have a tableau with `n` rows.
(b) `Destabilizer` on `n` qubits should always have a tableau with `2*n` rows.
(c) `MixedStabilizer` on `n` qubits should be the same as (a) but with only `rank` non-trivial rows.
(d) `MixedDestabilizer` on `n` qubits should be the same as (b) but with only `2*rank` non-trivial rows, namely `[1 : rank]` and `[n+1 : n + rank]`.

[*] Disclaimer: should anyone be insane enough to employ my implementation on such massive scales and end up encountering any problems with the above limitations, then they are invited to get in touch so that I may:
(a) Thoroughly assess their mental well-being.
(b) Develop and deploy a customised solution for whatever ridiculous use-case they have.

Krastanov:
Hi, @Hamiltonian-Action , thanks for looking into this again, it is a very impressive piece of work that I would love seeing merged in the library!

Before addressing your question, in terms of logistics, if you are already mostly wrapped with implementing some (or all) of the canonicalization methods for Stabilizers, let's try to get that merged without worrying about adding the other tableau types -- that would be of significant help with the review velocity.

I also have a quick question here -- is there a prescription whether one should use `fastrow` or `fastcolumn` for these calls? Warnings or even flat-out error message that refuses to proceed with unpleasant layout is perfectly reasonable here.

To your questions -- first, yeah, the library is still underdocumented. The `show` methods and the `stabilizerview`, `destabilizerview`, `logicalxview`, and `logicalzview` are probably the piece of code that are most self-explanatory about internal layout of the types you mentioned.

Here are the more concrete answers:

(a) Stabilizer on n qubits should always have a tableau with n rows.

Many of the methods working on Stabilizer are resilient to having fewer than n rows, and most of the ones that do require n rows have warnings if there is a problem. In particular, this type can be conveniently used for representing an error correcting code, so it inherently permits fewer than n qubits. We do not have a distinction in the type system between "a full-rank tableau" and an "incomplete tableau". Here are a couple of edge cases that do work fine with `canonicalize!`

```
julia> canonicalize!(S"XX XX")
+ XX
+ __

julia> canonicalize!(S"ZZZ IIZ")
+ ZZ_
+ __Z
```

If you need a square Stabilizer for your work, feel free to add a check that raises something like a dim mismatch error or a domain error. But Gaussian elimination of a non-square tableau is something that is generally useful.

(b) Destabilizer on n qubits should always have a tableau with 2*n rows.

Yes, although not particularly well enforced. The first n rows correspond to the destabilizers.

(c) MixedStabilizer on n qubits should be the same as (a) but with only rank non-trivial rows.

This one is a weird one... It exists out of a desire for completeness, but there is hardly a situation where it is a good data structure. I have been considering making it a private symbol in a future breaking release.

It exists so that one can cheaply add new rows to a tableau. Thus it is useful if one is doing projection operations. But if one is doing projection operations, they should be using a datastructure that stores destabilizers too.

So internally it is a square tableau, but there is a counter that gives you which rows are "valid". If you do a projective operation or a partial trace you can change that counter value.

(This discussion also makes Destabilizer a bit of an awkward one. Who would ever need a tableau on which they can do projective operations without needing the capability to do partial traces, which is what Destabilizer gives you).

(d) MixedDestabilizer on n qubits should be the same as (b) but with only 2*rank non-trivial rows, namely [1 : rank] and [n+1 : n + rank].

MixedDestabilizer is probably the only other type that really matters besides Stabilizer. It is 1:rank destabilizers, then rank+1:n "logical x operators", then n+1:n+rank stabilizer operators, then up to 2n are the "logical z operators".

All of these entries are important (unlike the meaningless madding in MixedStabilizer). The destabilizers and the stabilizers are clear. The others are probably also known, but for completeness I will write a short explanation:

The "logical X and Z operators" (named "X" or "Z" arbitrarily) are the operators that commute with all stabilizers. All Z operators commute between each other. Same for the X ones. But they anti commute only pairwise -- the first X operator with the first Z operator.

One naive way to construct such a tableau is to start with your fewer-than-n stabilizers, to arbitrarily pad them with valid stabilizer operators such that you have a valid full rank tableau (those are the logical Z operators). That can be achieved with a contrived version of Gaussian elimination. Then you use the standard procedure to find the destabilizers, but call the ones paired up with the logical Z operators the "logical X operators".

This tableau format is extremely valuable for working with (1) partial traces (in addition to supporting measurements). And it is extremely valuable for working with (2) error correcting codes.

Krastanov:
In terms of your code, @Hamiltonian-Action , supporting `fastrow` seems to be already the case and supporting `fastcolumn` is equivalent to supporting `Tableau{..., LinearAlgebra.Adjoint{UInt..., DeviceMatrix{UInt...}}}}`, which is probably not worth it (but if it is magically already the case, then woohoo)

Hamiltonian-Action:
> `transpose(collect(transpose(...)))`

Quite succinct. For memory footprint purposes I wonder whether it would be possible to do this in place but that's definitely far from the top of the to-do list.

> supporting `fastcolumn` is equivalent to supporting `Tableau{..., LinearAlgebra.Adjoint{UInt..., DeviceMatrix{UInt...}}}}`, which is probably not worth it (but if it is magically already the case, then woohoo)

The dispatch signature that the implementations specialise to is some combination of `PauliOperator{P, XZ}` and `Tableau{P, XZ}` wherein `{P <: AbstractGPUArray, XZ <: AbstractGPUArray}`. As such, seeing as I do not perform any funky tricks such as linearly indexing multidimensional structures, then everything should function properly regardless of whether or not `fastcolumn` has been applied without any effort whatsoever.
However, as noted above regarding device side memory transactions, having two consecutive threads access `fastcolumn(xzs)[i, j]` and `fastcolumn(xzs)[i + one(i), j]` respectively is one of the biggest performance shotguns imaginable and perhaps a warning (but not an error) should be raised in this case.
It is trivial to find an asymptotically ideal solution that is efficient on both `fastrow` and `fastcolumn` layouts via batching across both axes of the tableau but, as with countlessly many solutions I typically come up with, the JuliaGPU landscape does not provide the necessary tooling to do this, let alone safely.

Hamiltonian-Action:
Okay, these last two weeks were a headache in trying to reconcile a flexible design with maintaining the same format as the existing host-side functions -- as well as being a literal headache due to illness. After much pondering, I decided that the best course of action is to perform rank tracking in a GPU specific manner and then only translate it to the present host-side format as a post-processing step.
@Krastanov, given that the implementation being considered now only supports up to `Int` qubits and the present library already enforces that restriction via the very `PauliOperator` and `Tableau` structures themselves, would it be prudent to include some overflow checks and signals or do I have permission to disregard this as undefined behaviour and optimise around it?

Hamiltonian-Action:
Okay, so this draft got pretty stale to reasonably merge and I ended up having to do some major restructuring elsewhere to keep the code clean -- one should preferably avoid function calls with a dozen arguments.
As such, would it be better to have a series of smaller PRs that do the necessary incremental updates, for easier reviewing, that build-up to the canonicalisation?
I'll leave this open for now as a reference, up until this is well and truly concluded, but I genuinely do not envision this branch being merged.

Note: I also built a small module (roughly 200 lines) that facilitates bit-packing enumeration types, which could be of general use in the greater realm of things, should it be included as a submodule or would you prefer that I just publish it and then have it as a dependency?

Krastanov:
Smaller PRs are certainly more helpful, however:

- this PR is reasonably self-contained (it touches a ton of stuff in your extension, but very few things outside of the extension), so I will not be insisting on small PRs unless that would provide some organizational value to you. My own review of the code will be focused on looking at the implemented tests, trying it out to get a sense of the UX, reviewing the file organization from afar, and only reviewing in-depth relatively small parts of the actual low-level code.
- having a dependency on a very small bit-wrangling package sounds perfectly fine to me -- it would even (very minorly) speed up compilation thanks to compilation being easy to parallelize at the package boundary. However, could you consider pursuing hosting the package in one of the github Julia community organizations. That way we do not need to be worried about its bus factor as much. Not sure which, maybe JuliaMath. Or we can host it here in the QuantumSavory org. The goal is just to protect against bus factor issues

Hamiltonian-Action:
* I reckon that smaller PRs would be more helpful for later referencing rather than having one incredibly massive overhaul that re-does everything. Everyone obviously strives to avoid introducing bugs and/or performance regressions but there's always the possibility of missing some edge case or making some subtle unjustified assumption; this way it will at least be possible for someone in the future to localise the issue(s) with something like `git bisect` without having an aneurysm from the size of the commit that induced the problem(s).

* I doubt any compilation speed-up will be possible since the most important aspect, namely the bit-unpacking, relies entirely on `@generated` functions in order to induce the minimal possible overhead at runtime, not to mention that the enumeration types need to be embedded directly into the structure type analogously to how certain things are done in `StaticArrays.jl`. In any case, the functionality itself is logically self-contained and could be of use elsewhere so I shall polish it a bit more and then try and find some suitable parent organisation to publish under.

Hamiltonian-Action:
Several API iterations later -- they all worked, I just despised the user experience I had designed each time -- there is now finally a potential `0.1.0` release of `BitPackedInstances` in the next two days.
The good news is that it is finally done, the bad news is that `@generated` functions are incredibly restricted by world age and introduce a regression to the position-dependent code of yore.

## It just works :tm:
```julia
julia> @enum CardinalDirections begin east; north; west; south; end

julia> using BitPackedInstances

julia> x = PackedInstances(UInt, north)
```

## Woe is me
```julia
julia> using BitPackedInstances

julia> @enum CardinalDirections begin east; north; west; south; end

julia> x = PackedInstances(UInt, north)
```

Hamiltonian-Action:
@Krastanov, would it be alright to host the package under @QuantumSavory?
I definitely spent far too much time on this -- given that it started out only as a small personal-use package to sidestep an unfortunate shortcoming in the JuliaGPU ecosystem -- and ought to **really** get back on track with the canonicalisation overhaul.

Krastanov:
Hosting it in this org sounds good, especially if it is about to become a dependency.

Krastanov:
@Hamiltonian-Action , I kinda lost track on how things are going here. I am sorry I have not been very helpful -- after the Unitary Hack hackathon, my github notifications are submerged in spam, so I might have missed a bunch of communication. On a brighter note, a baby joined my family recently, so on one hand I have completely lost track of a bunch more communication, on the other hand parental leave is letting me look at some volunteer projects.

Anyway, with this long preamble, please just let me know if there is anything I can do to help with this PR and let me know if I have already forgotten about any requests that have been sent my way. Thanks again for your volunteer work on this.

Hamiltonian-Action:
First and foremost, congratulations to you and your family and I hope you're all in good health.

Truth be told, I myself am not quite so satisfied at the rate of progress as I was quite ill for most of March/April and had to spend a great deal of June/July wrangling with some of the bureaucratic conundrums, which still are not suitably resolved to date, that have arisen due to the never-ending tensions in the middle east. Nevertheless, I am still slowly pushing the envelope even whilst preoccupied with other academic commitments.

To give the most terse summary of the current status, I offer only "Abstraction Nightmare" as my response. To be more explicit, whatever implementation is produced at the end has to respect the existing status quo of being **completely valid** irrespective of the underlying data types whilst simultaneously employing any and all manners of optimisation trickery whenever presented with a suitable opportunity[1]. Rather than relying on the design pattern of having a generic solution plus a few additional dispatch methods that cover the specific cases where further optimisations are possible, I am attempting to strike a compromise of having only one implementation that relies on a consistent set of utilities that abstract away these optimisations[2]. The problems herein are obviously finding the right abstractions to employ as well as ensuring that we avoid stepping over other projects' territories [3] as to avoid terrible pains further down the line.

Please feel free to commentate on the incredibly abundant scope creep I am potentially (read: definitely) introducing into what supposedly ought to be reasonably straightforward binary field linear algebra.

[1] For example, all (?) vendors support vectorised loads/stores of 32, 64, and 128 bits in one transaction for less instruction overhead. However, this needs to be properly aligned in order to avoid invoking undefined behaviour and still needs to be broken down into smaller sections (such as 128 -> `SVector{4, UInt32}`) in order to best utilise native hardware instructions rather than their simulated counterparts, should the latter even exist.

[2] Continuing on with the above, one may consider something akin to the following:
```julia
function increment!(input::AbstractArray{<: Unsigned})
    for n in eachindex(input)
        section = make_mutable(load_data(input, n))
        for i in eachindex(section)
            section[i] += one(eltype(section))
        end
        store_data(input, n, make_immutable(section))
    end
end
```

[3] For instance, dealing with the hardware's properties should be handled by `JuliaGPU` and any reasonable attempt to do so otherwise is subject to spontaneous combustion should they ever decide to modify their own internals in an incompatible manner.

PR #579, opened 2025-08-18
Implement Tableaux randomisation and Adapt.jl extension.

Hamiltonian-Action:
The adapt extension presently only supports Paulis, Tableaux, and (Mixed)Stabilizers/(Mixed)Destabilizers. Further structures can be added as necessary since many of them do a less-than-terrific job of documenting their exact contents.

Regarding Tableaux randomisation, I am not quite certain if this implementation does a proper interpretation of what it means to apply Pauli noise to a tableau and would appreciate any remarks to that end. The functions have also not been exported since the `Tableau` type itself isn't, though the function `tab` is for some reason? In any case, this can be rectified quite trivially and the changelog shall be modified if so desired.


src/randoms.jl
```diff
@@ -60,6 +68,64 @@ random_pauli(n::Int; kwargs...) = random_pauli(GLOBAL_RNG, n; kwargs...)
...
+
+Use `nophase=false` to randomize the phase.
+Use `realphase=false` to get operators with phases including ±i.
+
+
+Optionally, a "flip" probability `p` can be provided specified,
+in which case each bit is set to I with probability `1-p` and to
+X or Y or Z with probability `p`. Useful for simulating unbiased Pauli noise.
+
+See also [`random_tableau!`](@ref)"""
```
Krastanov: There are a bunch of different ensembles from which someone might try to sample. While you are clear that this is just a random tableau, not a random tableau meant to represent a valid stabilizer state, this is subtle and it might not be obvious to a new reader.

Can we be a bit more explicit, with a docstring that includes the following caveats:

- This tableau is a list of independently sampled random Pauli operators, see `random_pauli` -- this function supports the same arguments
- This tableau does not represent a valid Stabilizer state (no commutation relationships are enforced)
- This function can be of use for testing purposes, if one need a large tableau without guaranteeing it is a "valid" tableau.

Krastanov (changes requested):
Looks good! I added a small comment about the docstring for the random tableau. Let's keep these unexported as you already have done. I will look into the rocm issue tonight. It seems when switching from the previous container to the new one, it now fails to keep different test runs independent.

src/randoms.jl
Hamiltonian-Action: Hope the new docstring communicates this more clearly.

PR #581, opened 2025-08-20
small fix to applybranches for resetmeasurement

brohan-byte:
removed a line from the applybranches which forced the recorded classical bit to be false

### DIFF
--- a/src/classical_register.jl
+++ b/src/classical_register.jl
@@ -192,7 +192,6 @@ function applybranches(r::Register, op::AbstractResetMeasurement;max_order=1)
         r.bits[op.bit] = res==0x02
         if(res==0x02)
             apply!(r, flipOp(op.qubit))
-            r.bits[op.bit] = false
         end
         push!(new_branches, (r,continue_stat,1,0))
     end


Krastanov:
`.bit` it seems is permitted to be zero, implicitly meaning `nothing` -- terrible design choice of mine, but for now let's just guard against it, we can think of a better design in the future

src/classical_register.jl
```diff
@@ -172,7 +172,14 @@ end
 _reset_mappings(::Type{sMRZ}) = (sX, projectZ!)
 _reset_mappings(::Type{sMRX}) = (sZ, projectX!)
 _reset_mappings(::Type{sMRY}) = (sZ, projectY!)
-
+#helper function to only set a bit if the given index is non-zero, since internally 'nothing' is defined to be 0
+setBit = (bit_register, bit, value) -> begin
```
Krastanov: lets make it into an actual function -- use docstrings instead of comments if waranted

Krastanov: use `_setbit` for name

Krastanov: or even `_setregbit`

test/test_noisycircuits.jl
```diff
@@ -227,34 +227,76 @@
         @testset "Symbolic Reset Measurements" begin
             using QuantumClifford:tensor
             #test for checking reset capabilities
-            state1 = S"-Z"
+            resetMeasureX = sMRX(1,1)#reset measurement ops that are used in the tests
```
Krastanov: bring back removed tests so that the diff looks more obvious (red lines need to be checked otherwise)

Krastanov: also, keep naming conventions more idiomatic -- CamelCase is for types

Krastanov (changes requested):
looks good, needs a bit of polish

test/test_noisycircuits.jl
Krastanov: see https://docs.julialang.org/en/v1/manual/style-guide/#Use-naming-conventions-consistent-with-Julia-base/-1

PR #583, opened 2025-08-23
Trivariate Tricycle code

### LINKED ISSUE #582: trivariate tricycle code
Implement TT code from https://arxiv.org/pdf/2508.08191

Fe-r-oz:
This PR implements the trivariate tricycle code and resolves #582.

API:

Here is the `[[432, 12, 12]` trivariate tricycle code from Table I from https://arxiv.org/pdf/2508.08191

```julia
julia> using Oscar; using QuantumClifford.ECC;

julia> l, m, p = 6, 6, 4;

julia> R, (x, y, z) = polynomial_ring(GF(2), [:x, :y, :z]);

julia> I = ideal(R, [x^l - 1, y^m - 1, z^p - 1]);

julia> S, _ = quo(R, I);

julia> A = S(1 + x*y*z^3 + x^3*y^4*z^2);

julia> B = S(1 + x^3*y*z^2 + x^3*y^2*z^3);

julia> C = S(1 + x^4*y^3*z^3 + x^5*z^2);

julia> c = TrivariateTricycleCode(l, m, p, A, B, C);

julia> code_n(c), code_k(c)
(432, 12)
```


nzy1997:
Is it possible to make the code work for any number of polynomials and variables? For example, bivariate bycycle code or bivariate tricycle code or Trivariate bycycle code. (No need to address this in the current PR, just thinking ahead)

Fe-r-oz:
> Is it possible to make the code work for any number of polynomials and variables? 

Thank you for your question. The generalization of these codes is a conjecture, see [5.3.2](https://arxiv.org/pdf/2410.16250) *Product of Λ ≥ 3 group algebra codes page 23*. The TT code parity check matrices aka boundary maps (2), (3), (4) of [reference ](https://arxiv.org/pdf/2508.08191) are described exactly in this other paper which are the building blocks and key insight for these codes.  Additionally, the reason why there are three polynomials in the Trivariate Tricycle case is to have a metacheck matrix $M_Z$ to enable partial single shot-decoding. The group algebra consists of direct product of three abelian groups that is isomorphic to the group generated by the permutation matrices using equation (5), (6), and (7) of [reference]( https://arxiv.org/pdf/2508.08191).

test/test_ecc_base.jl
```diff
@@ -238,9 +238,38 @@ b_gb₃ = 1 + x^6
...
+# [[108, 6, 2]] from Table IV
+ℓ₄, m₄, p₄ = 4, 3, 3
+A₄ = (1 + x^2)*(1 + x*z)
+B₄ = 1 + x^2*y^2
+C₄ = 1 + x^2*y^2*z^2
+
 # Add some codes that require Oscar, hence do not work on Windows
 
 test_twobga_codes = []
+test_generalized_toric_codes = []
```
Krastanov, line 272: @Fe-r-oz @nzy1997  The codes added here are not actually tested at all. See for instance how `test_twobga_codes` is used elsewhere in the code base. This comes from discussion in https://github.com/QuantumSavory/QuantumClifford.jl/pull/594/files

Fe-r-oz: Thanks for the feedback. The stabilizer correctness and properties for these codes are verified in the isolated [test_ecc_trivariate_tricycle.jl](https://github.com/QuantumSavory/QuantumClifford.jl/blob/master/test/test_ecc_trivariate_tricycle.jl) file. You are right that the new codes should be integrated into the main test suite in `test_ecc_base.jl` to match the pattern used by all other codes including `test_twobga_codes` and ensure they are tested with the full battery of checks which includes syndrome measurements, and decoding. I will take care of adding them to the main test suite in #594.

PR #585, opened 2025-08-29
add GraphSim module

LEXUGE:
Created a new GraphSim module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 ## v0.11.0-dev
 
 - Drop support for Julia <1.12.
+- Implementation of the specialized simulation algorithms for graph states (faster for large sparse tableaux).
+- **(breaking)** moving all graph state functionality to a submodule `GraphSim`.
 - **(fix)** `canonicalize_gott!` now properly supports non-UInt64 types.
 - `ptrace` for partial traces. We already had `traceout!`, which returns tableaux with the same number of qubits. `ptrace` removes the traceout qubits besides setting their stabilizing operators to `I`.
 - Add `apply_right!` that applies a Clifford operator to the right of a dense clifford operator.


src/QuantumClifford.jl
```diff
@@ -1452,6 +1452,7 @@ include("useful_states.jl")
 include("experimental/Experimental.jl")
 #
 include("./graphs/graphs.jl")
+using .GraphSim: graphstate, graphstate!, graph_gatesequence, graph_gate
```
LEXUGE: Not sure about this bit, should I export them by `export` in GraphSim? and just do `using .GraphSim` here?

src/graphs/graphs.jl
```diff
@@ -1,4 +1,9 @@
+module GraphSim
+
 import Graphs: Graphs, nv, AbstractGraph
+using QuantumClifford
```
LEXUGE: I am not sure about this. But I think having such a "prelude" import is cleaner than manually import everything used.

Krastanov: moving to explicit imports has been the current fashion in a lot of languages (no star imports in python, ExplicitImports.jl linter in Julia, etc). It is not too important, but I do prefer the explicit list, even if it is long.

Krastanov (commented):
Currently the doctests have not finished running, so I am not sure about them.

As for the documentation itself, the failure is in resolving the references. The rules for them are a bit different. Check the various module definitions in docs/make.jl and just add an extra one the way there are a few already for other submodules.

LEXUGE:
@Krastanov I think it's now working:

- public API is exposed directly under `QuantumClifford`, not breaking existing API
- However, as benefits of the module, internal symbols won't be exposed / cause side-effects on other parts of the code.

Krastanov:
@LEXUGE , could you remind me what the status here is? Is this ready for merge after a review or do we have any outstanding questions about it?

LEXUGE:
It doesn't have anything outstanding pending on me. So should be OK to merge after a review

Krastanov:
This was embarrassingly slow to review on my end, apologies! The semester always makes it difficult to find review time.

At your convenience we can discuss next steps and how to include this in the upcoming manuscript about QuantumClifford.

PR #588, opened 2025-08-31
correctness tests for classical ECCs

### LINKED ISSUE #268: Correctness checks for classical ECCs
@Fe-r-oz , I see you have contributed quite a few different classes of classical ECCs. E.g. in:

- #254 
- #256 
- #263
- #266

These are quite valuable contributions, but it will take me some time to review them. My chief difficulty in reviewing them would be figuring out some selfconsistency checks. What I mean by this is that the following would NOT be a good way to review:

> I read the papers, then read your code and check whether your code follows the algorithms in the papers. You have already basically done the same, and me redoing it is not very valuable because then the contribution and the review would be susceptible to the exact same type of mistakes. Rereading some code is simply not a good way to review.

Instead, we need to figure out some self-correcting method. There are a few such self-correcting checks done for the quantum codes for instance. Checking that they have valid tableaux, checking ranks, verifying that encoding circuits work as expected, etc.

In order to merge all of these classical code contributions of yours, we need to figure out similar self-correctness checks. I do not have a lot of good ideas for that right now (hence this issue, so we can discuss), but a couple of options include:

- checking the rank of the binary matrices and verifying we do not have redundant rows
- checking that each column of a binary matrix has at least one non-zero entry
- finding some code database and comparing the binary matrices generated here to the matrices from the database (probably one of the most valuable options, although it also has potential complications if the database uses different bit order conventions)
- similarly to the database option, finding another piece of software that also generates these codes and running comparisons between the two

Do you know of such databases or alternative pieces of software?

Fe-r-oz:
This PR adds the automated tests for the classical ECCs by creating the `test_cecc_base.jl` that has all the classical ECCs instances which are then utilized in the `test_cecc.jl` to test the general properties of the parity check matrices of the classical error correcting codes. This resolves #268


test/test_cecc_base.jl
```diff
@@ -0,0 +1,73 @@
...
+                codeinstance = t(args...)
+                !isnothing(maxn) && code_n(codeinstance) > maxn && continue
+                push!(codeinstances, codeinstance)
+            catch e
+                @warn "Failed to create $t with args $args: $e"
+            end
+        end
+    end
+    return codeinstances
+end
```
Fe-r-oz: I will improve this so it also check with pop that all the codes are tests and simplify this similar to what is done for QECCs

test/test_cecc.jl
```diff
@@ -0,0 +1,27 @@
+@testitem "ECC" tags=[:ecc] begin
```
Krastanov: I think the tag might need to be update now that we have split the tests. I think `:ecc_base` is what we want.

Fe-r-oz: It is fixed. Also, I used the same approach (the `pop!` method) to detect that all the code in the dictionary are being instantiated! I did not added the classical `LiftedCode` to the dictionary (`code instance args`) because there is a method errors when instantiating `LiftedCode` which are being resolved at #525

PR #593, opened 2025-09-08
Add APIs for decoder pipeline in QECCore

nzy1997:


nzy1997:
@Krastanov This PR is ready for review when you have time. Thanks!

lib/QECCore/src/decoding_impl.jl
```diff
@@ -0,0 +1,189 @@
...
+"""
+    $TYPEDSIGNATURES
+
+Create a depolarization error model from a vector of error probabilities. 
+The `i`-th element of the vector is the depolarization probability of qubit `i`.
+"""
+function depolarization_error_model(pvec::Vector{Float64}, qubit_num::Int)
+    @assert length(pvec) == qubit_num "The length of the vector of error probabilities must match the number of qubits"
+    probabilities = Dict([i, i + qubit_num] => [1.0-pvec[i] pvec[i]/3; pvec[i]/3 pvec[i]/3] for i in 1:qubit_num)
+    return FactoredBitNoiseModel(2 * qubit_num, probabilities)
```
Krastanov, line 66: Conflating classical noise for bits and quantum noise for qubits seems dangerous. Maybe we should have `FactoredBitNoiseModel` and `FactoredPauliNoiseModel` and potentially ways to convert between them.

lib/QECCore/src/decoding_impl.jl
```diff
@@ -0,0 +1,189 @@
...
+Create a depolarization error model from a depolarization probability. All qubits have the same depolarization probability.
+"""
+depolarization_error_model(p::Float64, qubit_num::Int) = depolarization_error_model(fill(p, qubit_num), qubit_num)
+
+"""
+    $TYPEDSIGNATURES
+
+Check if the error model is a vector error model.
+"""
+isvector(em::FactoredBitNoiseModel) = all(length(key) == 1 for key in keys(em.probabilities))
```
Krastanov, line 81: if vector+independent error models are commonly used, maybe we should have a special type of it, that uses an array instead of a dictionary for storage, as that would provide much faster iteration

lib/QECCore/src/decoding_impl.jl
```diff
@@ -0,0 +1,189 @@
...
+    $TYPEDFIELDS
+"""
+struct DetectorModelProblem{NMT<:AbstractNoiseModel} <: AbstractDecodingProblem
+    """Probabilistic error model for the bits"""
+    error_model::NMT
+    """Parity check matrix
+    - **Matrix dimensions**: [`num_checks` × `num_bits`]
+    - **Matrix elements**: Boolean (false=0, true=1)
+    - **Operation**: Syndrome = `check_matrix` × `error_vector` (mod 2)"""
+    check_matrix::Matrix{Bool}
```
Krastanov, line 148: shouldn't this be a code, not a matrix?

Krastanov: Independently, my main worry is that there is some implicit conversion you have in mind between classical and quantum codes, that is not documented or enforced here. If someone gets a DetectorModelProblem they will not be able to tell whether it is a classical or quantum one. That can matter a lot when you decide on a decoding algorithm (in the quantum case, many of them are two independent classical decoders)

lib/QECCore/src/QECCore.jl
```diff
@@ -14,6 +14,12 @@ rate, metacheck_matrix_x, metacheck_matrix_z, metacheck_matrix, bivariate_bicycl
 generator_polynomial
 export AbstractECC, AbstractQECC, AbstractCECC, AbstractCSSCode, AbstractDistanceAlg
 
+# Decoding interfaces
+export AbstractDecoder, AbstractDecodingResult, AbstractDecodingProblem, AbstractNoiseModel, AbstractDecodingScenario, decode, sample, decoding_error_rate, AbstractSyndrome, syndrome
+
+# DetectorModelProblem
+export DetectorModelProblem, FactoredBitNoiseModel, depolarization_error_model, isvector, isindependent, IndependentVectorSampler, BitStringSamples, MatrixDecodingResult, MatrixSyndrome
```
Krastanov: let's not export all of these yet, as we would want to see how they are being used in QuantumClifford, TensorQEC, and the ecc bench wiki, before committing to public support. We might see that changes are necessary before everything can be ported over.

lib/QECCore/docs/src/decoding_pipeline.md
```diff
@@ -0,0 +1,53 @@
...
+## Overview
+
+`decoding_interface.jl` defines the core abstract interfaces for quantum error correction decoding systems, providing a modular framework for handling decoding problems of quantum codes.
+
+## Core Abstract Types
+
+### Problem Definition
+- `AbstractDecodingProblem`: Abstract base class for decoding problems
+- `AbstractNoiseModel`: Abstract base class for noise models  
+- `AbstractDecodingScenario`: Abstract base class for decoding scenarios
```
Krastanov, line 11: This does not seem to be used currently. What is the difference you have in mind between `AbstractDecodingProblem` and `AbstractDecodingScenario`.

In QuantumClifford for instance we have a PauliFrame-based setup that includes circuit noise and a commutation-check-based setup where we only model memory noise and do not simulate any circuits. Are these `Problem`s or `Scenario`s in the abstraction you are introducing. I suspect I can just use the `Problem` and we can delete the `Scenario`?

lib/QECCore/src/decoding_impl.jl
```diff
@@ -0,0 +1,189 @@
...
+    return Bool.(mod.(check_matrix * error_patterns, 2))
+end
+measure_syndrome(problem::DetectorModelProblem, error_patterns::Matrix{Bool}) = measure_syndrome(problem.check_matrix, error_patterns)
+
+struct MatrixSyndrome <: AbstractSyndrome
+    mat::Matrix{Bool}
+end
+syndrome(samples::BitStringSamples) = MatrixSyndrome(samples.check_bits)
+
+struct MatrixDecodingResult <: AbstractDecodingResult
```
Krastanov, line 211: This is not documented but it seems to be a rather important concept. Currently I am not sure what it is meant to represent.

Maybe we can also rename `mat` to a more semantically meaningful name?

lib/QECCore/src/decoding_interface.jl
```diff
@@ -0,0 +1,81 @@
...
+
+### Inputs
+- `problem::AbstractDecodingProblem`: The decoding problem to validate.
+- `syndrome::AbstractSyndrome`: The syndrome to validate.
+- `decoding_result::AbstractDecodingResult`: The decoded result to validate.
+
+### Outputs
+- `rate::Float64`: The error rate of the decoding result.
+"""
+function decoding_error_rate end
```
Krastanov, line 81: Maybe we can also have a shortcut `decoding_error_rate(problem::AbstractProblem, decoder::AbstractDecoder; samples=1000)` that does skips a bunch of the boilerplate?

lib/QECCore/src/decoding_interface.jl
```diff
@@ -0,0 +1,81 @@
...
+
+### Inputs
+- `code::AbstractCode`: The code to decode.
+- `noise_model::AbstractNoiseModel`: The noise model to use.
+- `decoding_scenario::AbstractDecodingScenario`: The decoding scenario to use.
+
+### Outputs
+- `problem::AbstractDecodingProblem`: The decoding problem.
+"""
+function decoding_problem end
```
Krastanov, line 16: this does not seem to be used, maybe we can just remove it (same discussion as in the comment I made about AbstractDecodingScenario)

lib/QECCore/docs/src/decoding_pipeline.md
```diff
@@ -0,0 +1,53 @@
...
+
+### 1. Problem Generation
+```julia
+decoding_problem(code::AbstractCode, noise_model::AbstractNoiseModel, decoding_scenario::AbstractDecodingScenario) -> AbstractDecodingProblem
+```
+Generate a decoding problem from a quantum code, noise model, and decoding scenario.
+
+### 2. Sampling
+```julia
+sample(problem::AbstractDecodingProblem, sampler::AbstractSampler) -> AbstractDecodingSamples
```
Krastanov: Maybe we can have `sample(::AbstractDecodingProblem)` that automatically chooses a sampler? Seems like that will be the case most of the time.

Krastanov (changes requested):
Apologies for the long delay, it is difficult to find review time during the semester.

This looks good! My main question is:

- do we even need `Scenario` -- it seems `Problem` already covers everything
- may we keep these unexported until TensorQEC or QuantumClifford starts using them -- that way we will have some confirmation that there are no missing features
- some questions about distinguishing bits and qubits
- some questions about documenting things

Do you think you will be able to look into this in the near future? Knowing your timeline would help me schedule some other review work I need to do. It would also help with figuring out how to approach writing about this in the upcoming manuscript about QuantumClifford

Krastanov:
If you are amenable to this, let's play a design "game" -- maybe it will help drill down to the basics of what we want without requiring each one of us to post even longer response comments to each other.

Here what I think the user facing API should look like. Could you take this and edit it to turn it into what you think it should look like. Let's iterate until we agree. And let's specifically focus on the user-facing side -- on purpose I am avoiding what is supposed to happen behind the scenes, e.g. whether PauliNoise is implemented in terms of some other type related to FactoredBitNoise or something

```
c = Toric(3,3)
s = CodeCapacityScenario(PauliNoise(0.1,0.1,0.1))
# or s = any of these
#  CodeCapacityScenario(UnbiasedNoise(0.1)),
#  NaiveSyndromeExtractionScenario(code, memory_noise=..., gate_noise=...)
#  ShorSyndromeExtractionScenario(code, memory_noise=..., gate_noise=...)
prob = DecodingProblem(c, s)
d = TableDecoder(prob)
sample_decode_evaluate(c, s, d; num_samples)
```

There are some problems with this that I see:

### 1. Warnings for inconsistent codes and noise types

```
c = RepetitionCode(3)
s = CodeCapacityScenario(PauliNoise(0.1,0.1,0.1))
prob = DecodingProblem(c, s)
# raises an error about quantum noise and classical code

# Fixed by doing:
c = CSS(RepetitionCode(3), nothing)
s = CodeCapacityScenario(PauliNoise(0.1,0.1,0.1))
prob = DecodingProblem(c, s)
```

```
c = Toric(3,3)
s = CodeCapacityScenario(BitFlipNoise(0.1))
prob = DecodingProblem(c, s)
# produces a warning that BitFlipNoise was converted to PauliNoise
```

### 2. Warning for inconsistent codes and decoders

```
c = Steane7()
s = CodeCapacityScenario(PauliNoise(0.1,0.1,0.1))
prob = DecodingProblem(c, s)
d = PyMatchingDecoder(prob)
# raises an error because PyMatching can not work with higher weight checks
```

### 3. How do we distinguish belief propagation on each classical component from belief prop on the entire code

I do not have a good idea how to do have the API distinguish these two:

- CSS code -> sample -> split syndrome in X part and Z part -> run belief prop on each part separately -> combine results
- CSS code -> sample -> run belief prop on the entire Tanner graph with the X and Z parts together

### 4. Specifying simulator (maybe this is what you meant by sampler and I just did not get it)

```
c = Toric(3,3)
s = CodeCapacityScenario(PauliNoise(0.1,0.1,0.1))
prob = DecodingProblem(c, s)
d = TableDecoder(prob)

sample_decode_evaluate(c, s, d; num_samples)
# method error because we do not have an imported simulator
# but with an error Hint saying something like
# "Install QuantumClifford or PyStim (does not exist currently) or TensorQEC"

using QuantumClifford
sample_decode_evaluate(c, s, d; num_samples) # runs fine

using TensorQEC
using PyStim

sample_decode_evaluate(c, s, d, QuantumClifford; num_samples)
sample_decode_evaluate(c, s, d, TensorQEC; num_samples)
sample_decode_evaluate(c, s, d, PyStim; num_samples)

prob2 = DetectorModelProblem(c, s)
d2 = TableDecoder(prob2)

sample_decode_evaluate(c, s, d, QuantumClifford; num_samples)
# currently this would raise an error with a hint to just use DecodingProblem or to install PyStim or TensorQEC

sample_decode_evaluate(c, s, d, TensorQEC; num_samples) # runs fine
sample_decode_evaluate(c, s, d, PyStim; num_samples) # runs fine
```

Could you just copy this comment and edit everything so that all the examples look the way you want them to look and add new examples for things that I missed.

nzy1997:
Below I’ve copied your example verbatim and edited it inline. I’ve added comments marked with “# CHANGED” where my design diverges.
```julia
c = Toric(3,3)
s = CodeCapacityScenario(PauliNoise(0.1,0.1,0.1))
# or s = any of these
#  CodeCapacityScenario(UnbiasedNoise(0.1)),
#  NaiveSyndromeExtractionScenario(memory_noise=..., gate_noise=...) # CHANGED
#  ShorSyndromeExtractionScenario(memory_noise=..., gate_noise=...) # CHANGED
prob = DecodingProblem(c, s)
d = TableDecoder(prob)
sample_decode_evaluate(prob, d; num_samples) # CHANGED: In all examples below, `sample_decode_evaluate` takes `prob` instead of `(code, scenario)`.
```

### 1. Warnings for inconsistent codes and noise types

```julia
c = RepetitionCode(3)
s = CodeCapacityScenario(PauliNoise(0.1,0.1,0.1))
prob = DecodingProblem(c, s)
# raises an error about quantum noise and classical code

# Fixed by doing:
c = CSS(RepetitionCode(3), nothing)
s = CodeCapacityScenario(PauliNoise(0.1,0.1,0.1))
prob = DecodingProblem(c, s)

# CHANGED: or maybe
c = RepetitionCode(3)
s = CodeCapacityScenario(BitFlipNoise(0.1))
prob = DecodingProblem(c, s)
```

I think the current behavior is reasonable and does not need to change:
```julia
c = Toric(3,3)
s = CodeCapacityScenario(BitFlipNoise(0.1))
prob = DecodingProblem(c, s)
# produces a warning that BitFlipNoise was converted to PauliNoise
```


### 2. Warning for inconsistent codes and decoders
I think the current behavior is reasonable and does not need to change:
```julia
c = Steane7()
s = CodeCapacityScenario(PauliNoise(0.1,0.1,0.1))
prob = DecodingProblem(c, s)
d = PyMatchingDecoder(prob)
# raises an error because PyMatching can not work with higher weight checks
```

### 3. How do we distinguish belief propagation on each classical component from belief prop on the entire code
One possible approach would be to make the default decoding problem correspond to the full code:
```julia
c = Steane7()
s = CodeCapacityScenario(PauliNoise(0.1,0.1,0.1))
prob = DecodingProblem(c, s)
```
Optionally, we could allow explicitly splitting X and Z at the problem-construction stage:
```julia
c = Steane7()
s = CodeCapacityScenario(PauliNoise(0.1,0.1,0.1); split_xz = true)
prob = DecodingProblem(c, s)
```
Here  `prob` will contain two separate small decoding subproblem `prob.probx` and `prob.probz`.

### 4. Specifying simulator


```julia
c = Toric(3,3)
s = CodeCapacityScenario(PauliNoise(0.1,0.1,0.1))
prob = DecodingProblem(c, s)
d = TableDecoder(prob)

sample_decode_evaluate(prob, d; num_samples)
# method error because we do not have an imported simulator
# but with an error Hint saying something like
# "Install QuantumClifford or PyStim (does not exist currently) or TensorQEC"

using QuantumClifford
sample_decode_evaluate(prob, d; num_samples) # runs fine

using TensorQEC
using PyStim

sample_decode_evaluate(prob, d, QuantumClifford; num_samples)
sample_decode_evaluate(prob, d, TensorQEC; num_samples)
sample_decode_evaluate(prob, d, PyStim; num_samples)

# CHANGED: `DetectorModelProblem` is an internal concrete subtype of `AbstractDecodingProblem`.
# `decoding_problem(...)` constructs different concrete problem types depending on the scenario.

s = SomeSetupThatRelatedToDetectorModelProblem()
prob2 = DecodingProblem(c, s) # DetectorModelProblem :> AbstractDecodingProblem
d2 = TableDecoder(prob2)

sample_decode_evaluate(prob, d, QuantumClifford; num_samples)
# currently this would raise an error with a hint to just use DecodingProblem or to install PyStim or TensorQEC

sample_decode_evaluate(prob, d, TensorQEC; num_samples) # runs fine
sample_decode_evaluate(prob, d, PyStim; num_samples) # runs fine
```

### 5. `DecodingProblem` or `decoding_problem`
```julia
c = Toric(3,3)
s = CodeCapacityScenario(PauliNoise(0.1,0.1,0.1))
prob = decoding_problem(c,s) # or
prob = DecodingProblem(c,s)
```
My preference is to make decoding_problem(c, s) the primary user-facing entry point.
It clearly communicates that a decoding problem is being constructed from a code and a
scenario, while allowing the function to return different concrete problem types
(e.g. detector-model-based problems) behind the scenes.

nzy1997:
Thanks for the detailed feedback and the structured design "game" — it's been very helpful for clarifying the API design. Let me address the remaining open questions.

## On DetectorModelProblem

The design philosophy behind `DetectorModelProblem` is to encapsulate the problem formulation that most mainstream decoders can handle: **a parity check matrix combined with a prior probability distribution over bit flips**.

### What can be represented as DetectorModelProblem:

| Case | Description |
|------|-------------|
| **A** | A classical code with bit-flip errors |
| **B** | A stabilizer tableau under `CodeCapacityScenario` with depolarizing errors — each qubit is represented by two classical bits with a joint prior distribution |
| **C** | A STIM file storing a detector error model |
| **D** | A STIM file storing a quantum circuit — convertible to a detector error model via STIM or TensorQEC |

### What cannot be easily represented:

- Circuits with complex error types (qubit-loss, coherent errors, erasure errors, etc.)
- Setups where the full circuit structure is essential for optimal decoding

### A Case Worth Discussing: NaiveSyndromeECCSetup

Even with realistic measurement circuits, the final input to the decoder is still a parity check matrix + prior distribution. The key question is how accurate this prior is:

- **Current QC.jl approach:** Use an approximate prior (e.g., uniform depolarizing noise). This is effectively a `DetectorModelProblem` where syndromes are sampled from the actual circuit. This is what I meant when I said the distinction is primarily in the sampler.

**Evidence from the codebase:** 

**1. Decoder construction only uses parity check matrix, not circuit/scenario info:**

```julia
# TableDecoder construction (src/ecc/decoder_pipeline.jl line 237-244):
function TableDecoder(c; error_weight=1)
    H = parity_checks(c)  # Only uses the code's parity check matrix
    # ... builds lookup table from H ...
    return TableDecoder(H, fm, n, s, k, error_weight, lookup_table)
end

# PyMatchingDecoder construction (ext/.../QuantumCliffordPyQDecodersExt.jl line 94-106):
function PyMatchingDecoder(c; weights=nothing)
    Hx = parity_matrix_x(c)  # Only uses the code's X parity checks
    Hz = parity_matrix_z(c)  # Only uses the code's Z parity checks
    H = parity_checks(c)
    # ... no circuit or noise model information passed ...
end

# BeliefPropDecoder construction (ext/.../QuantumCliffordLDPCDecodersExt.jl line 34-51):
function BeliefPropDecoder(c; errorrate=nothing, maxiter=nothing)
    Hx = parity_matrix_x(c)  # Only parity checks
    Hz = parity_matrix_z(c)
    H = parity_checks(c)
    errorrate = isnothing(errorrate) ? 0.0 : errorrate  # A scalar prior, not circuit-derived
    bpx = LDPCDecoders.BeliefPropagationDecoder(Hx, errorrate, maxiter)
    bpz = LDPCDecoders.BeliefPropagationDecoder(Hz, errorrate, maxiter)
    # ... no circuit structure passed ...
end
```

**2. Decoding only receives syndromes, regardless of setup:**

```julia
# CommutationCheckECCSetup (line 191-208):
function evaluate_decoder(d::AbstractSyndromeDecoder, setup::CommutationCheckECCSetup, ...)
    # ... sampling via comm!(syndromes, H, err) ...
    guesses = batchdecode(d, syndromes)  # decoder only sees syndromes
end

# NaiveSyndromeECCSetup (line 163-169):
function evaluate_decoder(d::AbstractSyndromeDecoder, nsamples, circuit, ...)
    syndromes = @view pfmeasurements(frames)[:, syndrome_bits]
    guesses = batchdecode(d, syndromes)  # decoder only sees syndromes
end
```

**Conclusion:** The decoder is constructed from `parity_checks(c)` only, and receives syndromes only. **No circuit structure or noise model is passed to the decoder.** The only difference between setups is how syndromes are generated.

- **Optimal approach:** Provide the decoder with full circuit structure and detailed error information. However, no existing decoder can process such information directly.

- **Long-term ideal:** Convert circuit-based scenarios to detector error models (like STIM does). This requires substantial work — implementing a circuit → detector error model pipeline — which I think should be a separate future effort. Note that in this approach, the parity check matrix in the detector error model is **no longer the code's original parity check matrix** — it becomes a more complex matrix whose size depends on both the number of QEC rounds and the code size. This is fundamentally different from the code-capacity case.

I agree with your suggestion to cache the original code and scenario in `DetectorModelProblem`:

```julia
struct DetectorModelProblem <: AbstractDecodingProblem
    parity_check_matrix
    prior_distribution  # or relevant fields for the detector model
    _code::Union{Nothing, AbstractECC}
    _scenario::Union{Nothing, AbstractDecodingScenario}
end
```

This allows QuantumClifford.jl to work directly with the cached code/scenario when beneficial, while still supporting the general detector model interface.

PR #594, opened 2025-09-08
tests: make all_testable_code_instances extension-aware

Fe-r-oz:
This PR enhances the `all_testable_code_instances` function to handle code instantiation of codes implemented in the `QuantumCliffordOscarExt`. Many advanced QECCs  such as TrivariateTricycle codes, and various homological product codes from the `QuantumCliffordOscarExt` were not instantiated by the `all_testable_code_instances` function when utilizing `code_instance_args` for the full spectrum of tests including measurements and decoding similar to what is done for `test_ecc_two_bga` codes in `test_ecc_base.jl`. 

The function now dynamically instantiates all available code types (including those from loaded extensions) and tries to instantiate them with their predefined parameters from the `code_instance_args`. If Oscar is used in CI, the `code_instance_args` is the merge of two dictionaries, namely  `oscar_code_instance_args` and `code_instance_args`. It now instantiate codes using the merged `code_instance_arg`s dictionary on available code types, then a fallback pass resolves any missing extension types from the same dictionary.

This PR original motivation was to move the Trivariate Tricycle codes tests to Oscar.jl test suite since the ideal of the multivariate polynomial ring is supported in Oscar.jl. The tests in base were not creating the quotient ring which is required to create trivariate polynomials to instantiate the TrivariateTricycleCode. Then we realized that the codes from QuantumCliffordOscarExt were not instantiated so now we have `oscar_code_instance_args` that is merged with the `code_instance_args` dictionary if Oscar CI is running.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -33,7 +33,7 @@ steps:
       - julia --project='~' -e '
         using Pkg;
         pkg"dev .";'
-  - label: "ECC Universal Tests"
+  - label: "ECC Code Properties"
     plugins:
       - JuliaCI/julia#v1:
           version: "1"
@@ -42,13 +42,72 @@ steps:
       - JuliaCI/julia-coverage#v1:
           codecov: true
     env:
-      ECC_UNIVERSAL_TESTS: true
+      ECC_CODE_PROPERTIES: true
     command:
       - echo "Julia depot path $${JULIA_DEPOT_PATH}"
       - julia --project='~' -e '
         using Pkg;
         pkg"dev .";'
-
+  - label: "ECC Encoding"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "1"
+      - QuantumSavory/julia-xvfb#v1:
+      - JuliaCI/julia-test#v1: ~
+      - JuliaCI/julia-coverage#v1:
+          codecov: true
+    env:
+      ECC_ENCODING: true
+    command:
+      - echo "Julia depot path $${JULIA_DEPOT_PATH}"
+      - julia --project='~' -e '
+        using Pkg;
+        pkg"dev .";'
+  - label: "ECC Decoding"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "1"
+      - QuantumSavory/julia-xvfb#v1:
+      - JuliaCI/julia-test#v1: ~
+      - JuliaCI/julia-coverage#v1:
+          codecov: true
+    env:
+      ECC_DECODING: true
+    command:
+      - echo "Julia depot path $${JULIA_DEPOT_PATH}"
+      - julia --project='~' -e '
+        using Pkg;
+        pkg"dev .";'
+  - label: "ECC Syndrome Measurement"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "1"
+      - QuantumSavory/julia-xvfb#v1:
+      - JuliaCI/julia-test#v1: ~
+      - JuliaCI/julia-coverage#v1:
+          codecov: true
+    env:
+      ECC_SYNDROME_MEASUREMENT: true
+    command:
+      - echo "Julia depot path $${JULIA_DEPOT_PATH}"
+      - julia --project='~' -e '
+        using Pkg;
+        pkg"dev .";'
+  - label: "ECC THROWS"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "1"
+      - QuantumSavory/julia-xvfb#v1:
+      - JuliaCI/julia-test#v1: ~
+      - JuliaCI/julia-coverage#v1:
+          codecov: true
+    env:
+      ECC_THROWS: true
+    command:
+      - echo "Julia depot path $${JULIA_DEPOT_PATH}"
+      - julia --project='~' -e '
+        using Pkg;
+        pkg"dev .";'
   - label: "ECC Bespoke Tests"
     plugins:
       - JuliaCI/julia#v1:
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -93,7 +93,6 @@ steps:
       - julia --project='~' -e '
         using Pkg;
         pkg"dev .";'
-        pkg"dev .";'
   - label: "ECC Syndrome Circuit Equivalence"
     plugins:
       - JuliaCI/julia#v1:
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -33,7 +33,7 @@ steps:
       - julia --project='~' -e '
         using Pkg;
         pkg"dev .";'
-  - label: "ECC Universal Tests "
+  - label: "ECC Universal Tests"
     plugins:
       - JuliaCI/julia#v1:
           version: "1"
@@ -42,7 +42,7 @@ steps:
       - JuliaCI/julia-coverage#v1:
           codecov: true
     env:
-      ECC_TEST_1: true
+      ECC_UNIVERSAL_TESTS: true
     command:
       - echo "Julia depot path $${JULIA_DEPOT_PATH}"
       - julia --project='~' -e '
@@ -58,7 +58,7 @@ steps:
       - JuliaCI/julia-coverage#v1:
           codecov: true
     env:
-      ECC_TEST_2: true
+      ECC_BESPOKE_TESTS: true
     command:
       - echo "Julia depot path $${JULIA_DEPOT_PATH}"
       - julia --project='~' -e '
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -78,7 +78,7 @@ steps:
       - julia --project='~' -e '
         using Pkg;
         pkg"dev .";'
-  - label: "ECC Syndrome Measurement"
+  - label: "ECC Syndrome Measurement Correctness"
     plugins:
       - JuliaCI/julia#v1:
           version: "1"
@@ -87,13 +87,29 @@ steps:
       - JuliaCI/julia-coverage#v1:
           codecov: true
     env:
-      ECC_SYNDROME_MEASUREMENT: true
+      ECC_SYNDROME_MEASUREMENT_CORRECTNESS: true
     command:
       - echo "Julia depot path $${JULIA_DEPOT_PATH}"
       - julia --project='~' -e '
         using Pkg;
         pkg"dev .";'
-  - label: "ECC THROWS"
+        pkg"dev .";'
+  - label: "ECC Syndrome Circuit Equivalence"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "1"
+      - QuantumSavory/julia-xvfb#v1:
+      - JuliaCI/julia-test#v1: ~
+      - JuliaCI/julia-coverage#v1:
+          codecov: true
+    env:
+      ECC_SYNDROME_CIRCUIT_EQUIVALENCE: true
+    command:
+      - echo "Julia depot path $${JULIA_DEPOT_PATH}"
+      - julia --project='~' -e '
+        using Pkg;
+        pkg"dev .";'
+  - label: "ECC Throws"
     plugins:
       - JuliaCI/julia#v1:
           version: "1"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -33,7 +33,7 @@ steps:
       - julia --project='~' -e '
         using Pkg;
         pkg"dev .";'
-  - label: "ECC Tests"
+  - label: "ECC Tests (Part 1)"
     plugins:
       - JuliaCI/julia#v1:
           version: "1"
@@ -42,7 +42,23 @@ steps:
       - JuliaCI/julia-coverage#v1:
           codecov: true
     env:
-      ECC_TEST: true
+      ECC_TEST_1: true
+    command:
+      - echo "Julia depot path $${JULIA_DEPOT_PATH}"
+      - julia --project='~' -e '
+        using Pkg;
+        pkg"dev .";'
+
+  - label: "ECC Tests (Part 2)"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "1"
+      - QuantumSavory/julia-xvfb#v1:
+      - JuliaCI/julia-test#v1: ~
+      - JuliaCI/julia-coverage#v1:
+          codecov: true
+    env:
+      ECC_TEST_2: true
     command:
       - echo "Julia depot path $${JULIA_DEPOT_PATH}"
       - julia --project='~' -e '
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -33,7 +33,7 @@ steps:
       - julia --project='~' -e '
         using Pkg;
         pkg"dev .";'
-  - label: "ECC Tests (Part 1)"
+  - label: "ECC Universal Tests "
     plugins:
       - JuliaCI/julia#v1:
           version: "1"
@@ -49,7 +49,7 @@ steps:
         using Pkg;
         pkg"dev .";'
 
-  - label: "ECC Tests (Part 2)"
+  - label: "ECC Bespoke Tests"
     plugins:
       - JuliaCI/julia#v1:
           version: "1"
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -49,10 +49,10 @@ testfilter = ti -> begin
         push!(exclude, :jet)
     end
 
-    if get(ENV, "ECC_TEST_1", "") == "true"
+    if get(ENV, "ECC_UNIVERSAL_TESTS", "") == "true"
         return (:ecc in ti.tags) && (:ecc_universal_checks in ti.tags)
 
-    elseif get(ENV, "ECC_TEST_2", "") == "true"
+    elseif get(ENV, "ECC_BESPOKE_TESTS", "") == "true"
         return (:ecc in ti.tags) && (:ecc_bespoke_checks in ti.tags)
     else
         push!(exclude, :ecc)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -49,9 +49,21 @@ testfilter = ti -> begin
         push!(exclude, :jet)
     end
 
-    if get(ENV, "ECC_UNIVERSAL_TESTS", "") == "true"
-        return (:ecc in ti.tags) && (:ecc_universal_checks in ti.tags)
+    if get(ENV, "ECC_CODE_PROPERTIES", "") == "true"
+        return (:ecc in ti.tags) && (:ecc_code_properties in ti.tags)
 
+    elseif get(ENV, "ECC_ENCODING", "") == "true"
+        return (:ecc in ti.tags) && (:ecc_encoding in ti.tags)
+
+    elseif get(ENV, "ECC_DECODING", "") == "true"
+        return (:ecc in ti.tags) && (:ecc_decoding in ti.tags)
+
+    elseif get(ENV, "ECC_SYNDROME_MEASUREMENT", "") == "true"
+        return (:ecc in ti.tags) && (:ecc_syndrome_measurement in ti.tags)
+
+    elseif get(ENV, "ECC_THROWS", "") == "true"
+        return (:ecc in ti.tags) && (:ecc_throws in ti.tags)
+    
     elseif get(ENV, "ECC_BESPOKE_TESTS", "") == "true"
         return (:ecc in ti.tags) && (:ecc_bespoke_checks in ti.tags)
     else
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -49,8 +49,11 @@ testfilter = ti -> begin
         push!(exclude, :jet)
     end
 
-    if get(ENV, "ECC_TEST", "") == "true"
-        return :ecc in ti.tags
+    if get(ENV, "ECC_TEST_1", "") == "true"
+        return (:ecc in ti.tags) && (:part1 in ti.tags)
+
+    elseif get(ENV, "ECC_TEST_2", "") == "true"
+        return (:ecc in ti.tags) && (:part2 in ti.tags)
     else
         push!(exclude, :ecc)
     end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -58,8 +58,11 @@ testfilter = ti -> begin
     elseif get(ENV, "ECC_DECODING", "") == "true"
         return (:ecc in ti.tags) && (:ecc_decoding in ti.tags)
 
-    elseif get(ENV, "ECC_SYNDROME_MEASUREMENT", "") == "true"
-        return (:ecc in ti.tags) && (:ecc_syndrome_measurement in ti.tags)
+    elseif get(ENV, "ECC_SYNDROME_CIRCUIT_EQUIVALENCE", "") == "true"
+        return (:ecc in ti.tags) && (:ecc_syndrome_circuit_equivalence in ti.tags)
+
+    elseif get(ENV, "ECC_SYNDROME_MEASUREMENT_CORRECTNESS", "") == "true"
+        return (:ecc in ti.tags) && (:ecc_syndrome_measurement_correctness in ti.tags)
 
     elseif get(ENV, "ECC_THROWS", "") == "true"
         return (:ecc in ti.tags) && (:ecc_throws in ti.tags)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -50,10 +50,10 @@ testfilter = ti -> begin
     end
 
     if get(ENV, "ECC_TEST_1", "") == "true"
-        return (:ecc in ti.tags) && (:part1 in ti.tags)
+        return (:ecc in ti.tags) && (:ecc_universal_checks in ti.tags)
 
     elseif get(ENV, "ECC_TEST_2", "") == "true"
-        return (:ecc in ti.tags) && (:part2 in ti.tags)
+        return (:ecc in ti.tags) && (:ecc_bespoke_checks in ti.tags)
     else
         push!(exclude, :ecc)
     end
--- a/test/test_ecc_base.jl
+++ b/test/test_ecc_base.jl
@@ -191,10 +191,10 @@ test_bb_codes = [bb1, bb2, bb3]
 # [[882, 24, 18 ≤ d ≤ 24]] from Appendix B of [panteleev2021degenerate](@cite)
 F = GF(2)
 R, x = polynomial_ring(F, "x")
-n = 7
-l = 63
-S, _ =  quo(R, x^l - 1)
-A_ghp1 = matrix(S, n, n,
+n_ghp1 = 7
+l_ghp1 = 63
+S_ghp1, _ =  quo(R, x^l_ghp1 - 1)
+A_ghp1 = matrix(S_ghp1, n_ghp1, n_ghp1,
          [x^27  0     0     0     0     1     x^54
           x^54  x^27  0     0     0     0     1
           1     x^54  x^27  0     0     0     0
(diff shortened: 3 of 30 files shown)


Krastanov:
@Fe-r-oz , could you run us through why this was not caught earlier, before the original code was merged? Mistakes like this happen, do not worry about it -- I am asking so we can structure the tests in a way in which this mistake is more difficult to make in the future.

Fe-r-oz:
> If there was something that was not working in the tests, why did the tests not fail in the previous PR?

Ah! I checked out the codes created via 

https://github.com/QuantumSavory/QuantumClifford.jl/blob/11dd3e53ebd2f2cb5e0cd1653839c0ea84cb5c1a/test/test_ecc.jl#L7

The codes only includes those created in HeckeExt. So, all the codes in OscarExt are not even instantiated when calling https://github.com/QuantumSavory/QuantumClifford.jl/blob/11dd3e53ebd2f2cb5e0cd1653839c0ea84cb5c1a/test/test_ecc_base.jl#L488

So, we should move all the Oscar codes in the Oscar.jl test suite 

https://github.com/QuantumSavory/QuantumClifford.jl/blob/11dd3e53ebd2f2cb5e0cd1653839c0ea84cb5c1a/test/test_ecc_base.jl#L289

rather than describing it via `all_testablable_code_instances()`. I will do this in the PR.

Fe-r-oz:
> I am sorry, I still do not understand. For instance, why did this line that you just removed not cause an error in the previous PR:

> :TrivariateTricycleCode => [(ℓ₁, m₁, p₁, A₁, B₁, C₁), (ℓ₂, m₂, p₂, A₂, B₂, C₂), (ℓ₃, m₃, p₃, A₃, B₃, C₃), (ℓ₄, m₄, p₄, A₄, B₄, C₄)]

The TT code is defined in` OscarExt`. The `all_testablable_code_instances()` does not instantiate `TrivariateTricycleCode` because it is defined in the Oscar.Ext. That's why there is no error because technically there QECCs in OscarExt should not be defined here. This will only instantiate the QECCs in HeckeExt. For OscarExt codes, we have a special function where those codes are declared.

Krastanov:
Ah, it is because of this line:
```
for t in subtypes(QuantumClifford.ECC.AbstractECC)
    for c in get(code_instance_args, t.name.name, [])
```

While `TrivariateTricycleCode` exists in the dictionary `code_instance_args` it does not exist in `subtypes` because the extension is not loaded.

One possible solution is to use `pop!` (or equivalent) instead of `get` and then check that we do not have any leftovers left in the dictionary. That would address potential failures like this in the future.

Krastanov:
The reason this is structured the way it is, is so that we can ensure that there are no subtypes that are missing tests. If you just create the codes by hand in the extra function that deals with oscar, that will not be tested anymore. I would suggest the following:

- make a `const oscar_code_instance_args = Dict()`
- inside of the oscar loading function populate it with entries the same way we already do for `code_instance_args`
- after `code_instance_args` is created, update it with the content of `oscar_code_instance_args`

Could you tag me when you have this, I would like to review it too, as this is a pretty crucial part of the test infrastructure.

test/test_ecc_decoder_all_setups.jl
```diff
@@ -7,9 +7,7 @@
     include("test_ecc_base.jl")
 
     @testset "table decoder, good for small codes" begin
-        codes = [
-                 all_testablable_code_instances(;maxn=10)...
-                ]
```
Krastanov, line 12: Could you undo this to make it match the code around it.

Generally avoiding stylistic changes helps keep the git blame tool more usable.

test/test_ecc_base.jl
```diff
@@ -402,128 +375,129 @@ h₄ = 1 + x + x^4
...
+                        code = t(args...)
+                        !isnothing(maxn) && nqubits(code) > maxn && continue
+                        push!(codeinstances, code)
+                    catch e
+                        @warn "Extension instance failed for $name: $(sprint(showerror, e))"
+                    end
+                end
+            else
+                @warn "Extension type $name not available"
+            end
```
Krastanov: This is drastically more complicated than the original. What is gained here?

test/test_ecc_base.jl
```diff
@@ -279,16 +228,40 @@ h₂ = 1 + x + x^2
...
+    :QuantumTannerGraphProduct => [(H1, H2),(H2, H2), (H1, H1), (H2, H1)],
+    :CyclicQuantumTannerGraphProduct => [(2,), (3,), (4,)],
+    :LaCross => [(5,h₂,true), (6,h₂,true), (8,h₂,true), (7,h₃,false)],
+    :TillichZemor => [(4,3,3), (5,4,4), (6,5,5), (7,6,6)],
+    :random_TillichZemor_code => [(6,4,3), (7,5,3), (8,6,3)],
+    :GeneralizedCirculantBivariateBicycle => [(9, 6, A1, B1)],
+    :GeneralizedHyperGraphProductCode => [(A_ghp1, b_ghp1, l_ghp1)],
+    :GeneralizedBicycleCode => [(a_gb₁,b_gb₁, l_gb₁), (a_gb₂,b_gb₂, l_gb₂)],
+    :ExtendedGeneralizedBicycleCode => [(c_gb₁,2,p_gb₁), (c_gb₂,3,p_gb₂)]
+)
```
Krastanov, line 305: Avoid moving code around unless there is a specific reason -- it makes using git blame more difficult and it makes the diff longer and more difficult to review. Could you move this back in the original location?

test/test_ecc_base.jl
```diff
@@ -191,86 +191,35 @@ test_bb_codes = [bb1, bb2, bb3]
...
+l_ghp1 = 63
+S_ghp1, _ =  quo(R, x^l_ghp1 - 1)
+A_ghp1 = matrix(S_ghp1, n_ghp1, n_ghp1,
          [x^27  0     0     0     0     1     x^54
           x^54  x^27  0     0     0     0     1
           1     x^54  x^27  0     0     0     0
           0     1     x^54  x^27  0     0     0
           0     0     1     x^54  x^27  0     0
           0     0     0     1     x^54  x^27  0
           0     0     0     0     1     x^54  x^27])
```
Krastanov, line 204: Any reason for these renamings? Same as in the other comments, these make git blame more difficult to use.

test/runtests.jl
```diff
@@ -49,8 +49,11 @@ testfilter = ti -> begin
         push!(exclude, :jet)
     end
 
-    if get(ENV, "ECC_TEST", "") == "true"
-        return :ecc in ti.tags
+    if get(ENV, "ECC_TEST_1", "") == "true"
+        return (:ecc in ti.tags) && (:part1 in ti.tags)
```
Krastanov: Instead of `:part1` we should probably use `ecc_part1`. More importantly, instead of just a numeric name, there should be some semantic meaning to what this is.

Same for part 2.

What is currently the logic of what goes in part1 and what goes in part2?

test/test_ecc_base.jl
```diff
@@ -402,128 +375,129 @@ h₄ = 1 + x + x^4
...
+    # [[241, 1, 9]] from Table I of https://arxiv.org/pdf/1805.09271
+    δ₁ = [1 1 0;
+          0 1 1]
+
+    oscar_code_instance_args[:DDimensionalSurfaceCode] = [(2, 3), (3, 2), (4, 2)]
+    oscar_code_instance_args[:DDimensionalToricCode] = [(2, 3), (3, 2), (4, 2)]
+    oscar_code_instance_args[:GeneralizedToricCode] = [(f₁, g₁, α1₁, α2₁), (f₂, g₂, α1₂, α2₂), (f₃, g₃, α1₃, α2₃), (f₄, g₄, α1₄, α2₄)]
+    oscar_code_instance_args[:HomologicalProductCode] = [([H₁,transpose(H₁)], l₁), ([δ₂,δ₂,δ₂],)]
+    oscar_code_instance_args[:DoubleHomologicalProductCode] = [(δ₁,)]
+    oscar_code_instance_args[:TrivariateTricycleCode] = [(ℓ₁, m₁, p₁, A₁, B₁, C₁), (ℓ₂, m₂, p₂, A₂, B₂, C₂), (ℓ₃, m₃, p₃, A₃, B₃, C₃), (ℓ₄, m₄, p₄, A₄, B₄, C₄)]
```
Krastanov: this would read more cleanly as
```
oscar_code_instance_args = Dict(
    :DDimension... => [(2,3), ...],
   ...
)
```

test/test_ecc_base.jl
```diff
@@ -402,128 +375,129 @@ h₄ = 1 + x + x^4
...
+        haskey(code_instance_args, name) || continue
+        for args in code_instance_args[name]
+            try
+                code = t(args...)
+                !isnothing(maxn) && nqubits(code) > maxn && continue
+                push!(codeinstances, code)
+                push!(tested, name)
+            catch e
+                @warn "Instance creation failed for $name: $(sprint(showerror, e))"
+            end
```
Krastanov: no need for a try catch here. If this fails, this is an error and should just error and crash the tests

test/test_ecc_base.jl
Krastanov: Can't you just keep the original code (which will already raise an error if there is a subtype that is not being tested) and just use `pop!` from it and check whether the dict is empty at the end? That way it is just one or two extra lines instead this much more complicated implementation.

test/test_ecc_base.jl
Fe-r-oz: I renamed the variable `l` (for lift size of matrix) because the subsequent generalized bicycle code examples also uses a variable named `l`. Since both `GeneralizedHypergraphCode` and `GeneralizedBicycleCode` use `l` as an input, the latter's value was shadowing the former, causing incorrect code parameters for the GHP case. To be safe, I used descriptive names ensures the correct input is passed to the right function. I will also review the changes to revert any unnecessary renames that do not contribute to any errors.

Krastanov: that change makes sense, it is an actual fix. Thanks for doing it, no need to revert it

test/test_ecc_base.jl
Fe-r-oz: The current reason is that since the Oscar function is defining the codes and making `oscar_code_instance_args` dictionary as constant and we are merging the dictionary with const `code_instance_args` as `merge!(code_instance_args, oscar_code_instance_args)` if  `code_instance_args` is not defined before Oscar function, it will throw error, so I moved it before the Oscar function. I will explore alternative method for this that does not require  `code_instance_args` to be moved.

Krastanov: No, you are right, this is a good reason.

Krastanov: its new position makes sense

test/test_ecc_base.jl
Fe-r-oz: Yeah; the `pop!` approach is simpler but there are some caveats -- I tried the `pop!` as you suggested which is a good sanity check. The current `all_testablable_code_instances` does not detect some other codes as well. For example, it only detects the following QECCs,


```
julia> subtypes(QuantumClifford.ECC.AbstractECC) .|> nameof
5-element Vector{Symbol}:
 :AbstractCECC
 :AbstractQECC
 :CircuitCode
 :Concat
 :LPCode
```
 So, all these codes that are of type `AbstractCSSCode` go undetected. Codes that are subtypes of `DDimensionalCode` also are not detected , e.g.  `DDimensionalSurfaceCode` and `DDimensionalToricCode` are missing from the list:

```
julia> subtypes(QuantumClifford.ECC.AbstractCSSCode) .|> nameof
22-element Vector{Symbol}:
 :CSS
 :CyclicQuantumTannerGraphProduct
 :DelfosseReichardt
 :DelfosseReichardtRepCode
 :GeneralizedCirculantBivariateBicycle
 :ColorCode
 :ExtendedGeneralizedBicycleCode
 :GeneralizedBicycleCode
 :GeneralizedHyperGraphProductCode
 :LaCross
 :DDimensionalCode
 :DoubleHomologicalProductCode
 :GeneralizedToricCode
 :HomologicalProductCode
 :TrivariateTricycleCode
 :QuantumReedMuller
 :QuantumTannerGraphProduct
 :Shor9
 :Steane7
 :Surface
 :TillichZemor
 :Toric
```

So, when I use the `pop!`, the dictionary is not empty at the end

```
ERROR: Unused code_instance_args: [:Toric, :QuantumReedMuller, :Gottesman, :Triangular488, :CyclicQuantumTannerGraphProduct, :DDimensionalSurfaceCode, :DDimensionalToricCode, :CSS, :GeneralizedCirculantBivariateBicycle, :Triangular666, :ExtendedGeneralizedBicycleCode, :DoubleHomologicalProductCode, :GeneralizedHyperGraphProductCode, :Surface, :TillichZemor, :DelfosseReichardtRepCode, :DelfosseReichardt823, :LaCross, :GeneralizedBicycleCode, :HomologicalProductCode, :random_TillichZemor_code, :DelfosseReichardt, :QuantumTannerGraphProduct, :GeneralizedToricCode, :TrivariateTricycleCode]
```

We can do something like the following with `pop!` at the end 

```
 roots = (QuantumClifford.ECC.AbstractECC, QuantumClifford.ECC.AbstractCSSCode)
 for root in roots
```

And this will get most of the QECCs, but for some that are some artificial subtype like `ColorCode` which has two subtypes, `Triangular666`, `Triangular488`, they are not detected.

```
ERROR: Unused code_instance_args: [:Triangular488, :DDimensionalSurfaceCode, :DDimensionalToricCode, :Triangular666, :random_TillichZemor_code]
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] top-level scope
   @ REPL[197]:2
```

For these undetcted subtypes, we can check the concerte subtypes for the subtypes to catch these recursively. The `random_TillichZemor_code `is a randomized function instead of struct so this can be skipped it's not a deterministic struct. It is already tested in [here](https://github.com/QuantumSavory/QuantumClifford.jl/blob/11dd3e53ebd2f2cb5e0cd1653839c0ea84cb5c1a/lib/QECCore/test/codes/tillichzemor.jl#L47)

Krastanov: Good point. 

Does the following solve it:
```
julia> concretesubtypes(T) = vcat([isconcretetype(s) ? [s] : concretesubtypes(s) for s in subtypes(T)]...)
concretesubtypes (generic function with 1 method)

julia> concretesubtypes(QECCore.AbstractQECC)
19-element Vector{Any}:
 CSS
 CyclicQuantumTannerGraphProduct
 DelfosseReichardt
 DelfosseReichardtRepCode
 GeneralizedCirculantBivariateBicycle
 Triangular488
 Triangular666
 QuantumReedMuller
 QuantumTannerGraphProduct
 Shor9
 Steane7
 Surface
 Toric
 Bitflip3
 Cleve8
 DelfosseReichardt823
 Gottesman
 Perfect5
 Phaseflip3

```

Krastanov: @nzy1997 , just FYI, this is probably an important thread to be aware of. A bunch of codes have stopped being tested during some of the restructuring this summer.

Fe-r-oz: > Does the following solve it:

Yeah; Almost -- `LPCode` is declared as AbstractECC so that's why it's not in the 19-element vector.

```
julia> c = all_testable_code_instances()
ERROR: Unused code_instance_args: [:Concat, :TillichZemor, :LPCode, :HomologicalProductCode, :CircuitCode]
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] all_testable_code_instances(; maxn::Nothing)
   @ Main ./REPL[202]:17
 [3] all_testable_code_instances()
   @ Main ./REPL[202]:1
 [4] top-level scope
   @ REPL[203]:1
```


with Oscar in environment; we now have 28 element vector excluding the above cases.

```
julia> show(concretesubtypes(QECCore.AbstractQECC))
Any[CSS, CyclicQuantumTannerGraphProduct, DelfosseReichardt, DelfosseReichardtRepCode, GeneralizedCirculantBivariateBicycle, Triangular488, Triangular666, QuantumCliffordHeckeExt.ExtendedGeneralizedBicycleCode, QuantumCliffordHeckeExt.GeneralizedBicycleCode, QuantumCliffordHeckeExt.GeneralizedHyperGraphProductCode, QuantumCliffordHeckeExt.LaCross, QuantumCliffordOscarExt.DDimensionalSurfaceCode, QuantumCliffordOscarExt.DDimensionalToricCode, QuantumCliffordOscarExt.DoubleHomologicalProductCode, QuantumCliffordOscarExt.GeneralizedToricCode, QuantumCliffordOscarExt.TrivariateTricycleCode, QuantumReedMuller, QuantumTannerGraphProduct, Shor9, Steane7, Surface, Toric, Bitflip3, Cleve8, DelfosseReichardt823, Gottesman, Perfect5, Phaseflip3]
```

I have pushed a commit that solves this as it does not lead the error by `pop!`

Fe-r-oz: Maybe better to change the type to `AbstractQECC` for the following and investigate others that was not in dictionary `[:Concat, :TillichZemor, :LPCode, :HomologicalProductCode, :CircuitCode]` when using `concretesubtypes(T)`

https://github.com/QuantumSavory/QuantumClifford.jl/blob/11dd3e53ebd2f2cb5e0cd1653839c0ea84cb5c1a/ext/QuantumCliffordHeckeExt/lifted_product.jl#L141

Krastanov: Codes that are not quantum codes should not be in this list of tests to begin with. Codes that are quantum codes should subclass from AbstractQECC -- if not that is a bug.

test/test_ecc_base.jl
```diff
@@ -402,129 +372,122 @@ h₄ = 1 + x + x^4
...
+        for t in _concrete_subtypes(root)
+            name = nameof(t)
+            if haskey(code_args, name)
+                for args in code_args[name]
+                    code = t(args...)
+                    !isnothing(maxn) && nqubits(code) > maxn && continue
+                    push!(codeinstances, code)
+                end
+                pop!(code_args, name)
+            end
```
Krastanov: I still do not follow why this needs to be changed so much. Why is it not enough to just change `get` to `pop!`?

Krastanov: Also, why are we including non-quantum codes here? Were they originally included? We have another set of tests for classical codes. I believe there are plenty of places that use this function that assume the codes it is returning are quantum, not classical.

test/test_ecc_base.jl
```diff
@@ -402,129 +372,122 @@ h₄ = 1 + x + x^4
...
         end
     end
+    !isempty(code_args) && error("Unused code_instance_args: $(collect(keys(code_args)))")
```
Krastanov: this can be just `@test isempty(...)`

test/test_ecc_base.jl
```diff
@@ -402,129 +372,122 @@ h₄ = 1 + x + x^4
...
-    :GeneralizedHyperGraphProductCode => [(A_ghp1, b_ghp1), (A_ghp2, b_ghp2)],
-    :GeneralizedBicycleCode => [(5,a_gb₁,b_gb₁), (9,a_gb₂,b_gb₂), (10,a_gb₃,b_gb₃)],
-    :ExtendedGeneralizedBicycleCode => [(c_gb₁,2,p₁), (c_gb₂,3,p₂), (c_gb₃,4,p₃)],
-    :DoubleHomologicalProductCode => [(δ₁), (δ₂)],
-    :TrivariateTricycleCode => [(ℓ₁, m₁, p₁, A₁, B₁, C₁), (ℓ₂, m₂, p₂, A₂, B₂, C₂), (ℓ₃, m₃, p₃, A₃, B₃, C₃), (ℓ₄, m₄, p₄, A₄, B₄, C₄)]
-)
+function _concrete_subtypes(T::DataType)
+    concrete = []
+    for t in subtypes(T)
+        if isempty(subtypes(t))
```
Krastanov: You probably want `isconcretetype` here.

test/test_ecc_base.jl
Fe-r-oz: Yeah; I will remove the `AbstractECC` from the roots.

Some of the QECC types are parameterized; so the following does not catch the parameterized types such as `HomologicalCode` or `TillichZemor` code. This is after I have changed some quantum codes that are subtyped as `AbstractECC` to `AbstractQECC` or `AbstractCSSCode` where makes sense, namely for `LPCode`, `CircuitCode`, and `Concat`

> julia> concretesubtypes(T) = vcat([isconcretetype(s) ? [s] : concretesubtypes(s) for s in subtypes(T)]...)

```
julia> c = all_testable_code_instances();
Test Failed at REPL[159]:15
  Expression: isempty(code_args)
   Evaluated: isempty(Dict{Symbol, Vector}(:TillichZemor => [(4, 3, 3), (5, 4, 4), (6, 5, 5), (7, 6, 6)], :HomologicalProductCode => Tuple{Vector, Vararg{Int64}}[(AbstractAlgebra.Generic.MatSpaceElem{FqPolyRingElem}[[x^2 x^2 x^2; x x^2 0], [x^2 x; x^2 x^2; x^2 0]], 3), (FqMatrix[[1 1 0; 0 1 1; 1 0 1], [1 1 0; 0 1 1; 1 0 1], [1 1 0; 0 1 1; 1 0 1]],)]))
```

test/test_ecc_generalized_bicycle_codes_slow.jl
```diff
@@ -1,4 +1,4 @@
-@testitem  "ECC GB Codes - slow" tags=[:ecc] begin
```
Krastanov, line 1: Just looking at the name of this test file, what is the relationship between `test/test_ecc_generalized_bicycle_codes.jl` and `test/test_ecc_generalized_bicycle_codes_slow.jl`?

test/test_ecc_base.jl
```diff
@@ -402,129 +372,117 @@ h₄ = 1 + x + x^4
...
-    for t in subtypes(QuantumClifford.ECC.AbstractECC)
-        for c in get(code_instance_args, t.name.name, [])
-            codeinstance = t(c...)
-            !isnothing(maxn) && nqubits(codeinstance) > maxn && continue
-            push!(codeinstances, codeinstance)
-            #@show i, t, code_n(codeinstance), code_k(codeinstance), code_s(codeinstance), code_n(codeinstance)-code_k(codeinstance)
-            i += 1
+    code_args = copy(code_instance_args)
+    roots = (QuantumClifford.ECC.AbstractCSSCode, QuantumClifford.ECC.AbstractQECC)
+    for root in roots
```
Krastanov: I still have the same question. Why does this need to be so significantly restructured? Let's plan to discuss this at one of our one-on-one meetings.

.buildkite/pipeline.yml
```diff
@@ -42,7 +42,23 @@ steps:
...
+  - label: "ECC Bespoke Tests"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: "1"
+      - QuantumSavory/julia-xvfb#v1:
+      - JuliaCI/julia-test#v1: ~
+      - JuliaCI/julia-coverage#v1:
+          codecov: true
+    env:
+      ECC_TEST_2: true
```
Krastanov: the env variables still need to be updated to match the new names

test/test_ecc_base.jl
Krastanov: Looking at this, maybe there is just a bug in how the type tree structure is set up:
```
julia> QECCore.AbstractECC |> subtypes
4-element Vector{Any}:
 AbstractCECC
 AbstractQECC
 CircuitCode
 Concat
```

Krastanov: it seems you have already fixed some of these bugs.

Fe-r-oz: Yes, we discussed this in the earlier review as well: https://github.com/QuantumSavory/QuantumClifford.jl/pull/594#discussion_r2342600487

test/test_ecc_generalized_bicycle_codes_slow.jl
Fe-r-oz: `test/test_ecc_generalized_bicycle_codes_slow` which correctness of MIP code distance algorithm of the GB codes. This was inspired by the distance of the codes given in [https://github.com/QEC-pages/GB-codes](https://github.com/QEC-pages/GB-codes) and only consider testing distance. `test/test_ecc_generalized_bicycle_codes.jl` contains tests for `GeneralizedBicycleCode`, `generalized_bicycle_codes`, and `ExtendedGeneralizedBicycleCode`. This file aims to reproduce all the given parameters mentioned in the referencs and check their stabilizer correctness and parameters.

**Edit**: Fixed the link and typos.

test/test_ecc_base.jl
Fe-r-oz: > Looking at this, maybe there is just a bug in how the type tree structure is set up:

> julia> QECCore.AbstractECC |> subtypes
> 4-element Vector{Any}:
> AbstractCECC
> AbstractQECC
> CircuitCode
>  Concat

I have fixed this tree in the previous commits before requesting review and now it correctly shows

```
julia> QECCore.AbstractECC |> subtypes
2-element Vector{Any}:
 AbstractCECC
 AbstractQECC
```

The original `all_testablable_code_instance` code 

https://github.com/QuantumSavory/QuantumClifford.jl/blob/11dd3e53ebd2f2cb5e0cd1653839c0ea84cb5c1a/test/test_ecc_base.jl#L517-L529

does not catch the concrete types, only the subtypes, so there is some modification to catch all the concrete and parameterized types and use `pop!`  at the end as validation check that the dictionary is empty and all QECCs are utilized.

The minor problem with the following after the above type tree is fixed is that 

> concretesubtypes(T) = vcat([isconcretetype(s) ? [s] : concretesubtypes(s) for s in subtypes(T)]...)

is that it does not catch some parameterized abstract types such as `HomologicalProductCode{T}` or `TillichZemor{M}` type which are `AbstractCSSCode` as shown in the following output. Editv4:  ̶I̶ ̶w̶i̶l̶l̶ ̶c̶h̶e̶c̶k̶ ̶i̶f̶ ̶a̶n̶y̶ ̶o̶t̶h̶e̶r̶ ̶t̶y̶p̶e̶ ̶i̶s̶ ̶n̶o̶t̶ ̶d̶e̶t̶e̶c̶t̶e̶d̶ ̶b̶y̶ ̶t̶h̶e̶ ̶a̶b̶o̶v̶e̶ ̶f̶u̶n̶c̶t̶i̶o̶n̶.̶  

```
julia> show(concretesubtypes(QECCore.AbstractQECC))
Any[CSS, CyclicQuantumTannerGraphProduct, DelfosseReichardt, DelfosseReichardtRepCode, GeneralizedCirculantBivariateBicycle, Triangular488, Triangular666, QuantumCliffordHeckeExt.ExtendedGeneralizedBicycleCode, QuantumCliffordHeckeExt.GeneralizedBicycleCode, QuantumCliffordHeckeExt.GeneralizedHyperGraphProductCode, QuantumCliffordHeckeExt.LPCode, QuantumCliffordHeckeExt.LaCross, QuantumCliffordOscarExt.DDimensionalSurfaceCode, QuantumCliffordOscarExt.DDimensionalToricCode, QuantumCliffordOscarExt.DoubleHomologicalProductCode, QuantumCliffordOscarExt.GeneralizedToricCode, QuantumCliffordOscarExt.TrivariateTricycleCode, QuantumReedMuller, QuantumTannerGraphProduct, Shor9, Steane7, Surface, Toric, Bitflip3, CircuitCode, Cleve8, Concat, DelfosseReichardt823, Gottesman, Perfect5, Phaseflip3]
```

Edit: Pressed entered while typing message, so edited to added proper stylistic blocks in the comments

test/test_ecc_base.jl
```diff
@@ -191,86 +191,35 @@ test_bb_codes = [bb1, bb2, bb3]
...
          [x^27  0     0     0     0     1     x^54
           x^54  x^27  0     0     0     0     1
           1     x^54  x^27  0     0     0     0
           0     1     x^54  x^27  0     0     0
           0     0     1     x^54  x^27  0     0
           0     0     0     1     x^54  x^27  0
           0     0     0     0     1     x^54  x^27])
-b_ghp1 = S(1 + x + x^6)
-
-# [[882, 48, 16]] from Appendix B of [panteleev2021degenerate](@cite)
```
Krastanov, line 207: Let's see in more details why this one is slow

test/test_ecc_base.jl
```diff
@@ -191,86 +191,35 @@ test_bb_codes = [bb1, bb2, bb3]
...
+p_gb₁ = one(R)
+l_gb₂ = 9
 a_gb₂ = 1 + x^2
 b_gb₂ = 1 + x^5
-c_gb₂ = GeneralizedBicycleCode(a_gb₂, b_gb₂, l)
-p₂ = one(R)
-l = 10
-a_gb₃ = 1 + x
-b_gb₃ = 1 + x^6
-c_gb₃ = GeneralizedBicycleCode(a_gb₃, a_gb₃, l)
```
Krastanov, line 238: same thing, let's figure out why this one is slow

test/test_ecc_base.jl
```diff
@@ -279,16 +228,37 @@ h₂ = 1 + x + x^2
 h₃ = 1 + x + x^3
 h₄ = 1 + x + x^4
 
-# Double Homological product codes
-δ₁ = [1 1 0;
-      0 1 1]
-δ₂ = [1 1 0;
-      0 1 1;
-      1 0 1]
+# Generalized Bivariate Bicycle Codes
```
Krastanov, line 252: mark these as just made randomly and bring back the other ones that were removed (and potentially mark them as taken from a paper)

test/test_ecc_base.jl
```diff
@@ -402,129 +372,117 @@ h₄ = 1 + x + x^4
...
-    append!(test_generalized_toric_codes, [gtc₁, gtc₂, gtc₃, gtc₄, gtc₅, gtc₆])
+    f₄ = 1 + x + x^-1*y^2
+    g₄ = 1 + y + x^-2*y^-1
+    α1₄ = (0,  7)
+    α2₄ = (7, 0)
+
+    # Double Homological product codes
+    # [[241, 1, 9]] from Table I of https://arxiv.org/pdf/1805.09271
+    δ₁ = [1 1 0;
+          0 1 1]
```
Krastanov: delta2 is removed, we need to figure out why it was slow

test/test_ecc_base.jl
```diff
@@ -402,129 +372,117 @@ h₄ = 1 + x + x^4
...
-    hpc₂ = HomologicalProductCode([H,transpose(H)], l)
-    # 3D Homological product code from [Quintavalle_2021](@cite)
-    μ = 2; wc = 3; wr = 4
-    c = GallagerLDPC(μ, wc, wr)
-    H = matrix(GF(2), parity_matrix(c))
-    hpc₃ = HomologicalProductCode([H,transpose(H)])
-    # 3D Homological product code from [Quintavalle_2021](@cite)
-    δ = matrix(GF(2), parity_matrix(RepCode(3)))
-    hpc₄ = HomologicalProductCode([δ,δ,δ])
-    @test iszero(mod.(metacheck_matrix_x(hpc₄)*parity_matrix_x(hpc₄), 2))
```
Krastanov, line 424: same as above, we need to figure out why these are slow and bring them back

test/test_ecc_base.jl
```diff
@@ -402,129 +372,117 @@ h₄ = 1 + x + x^4
...
+    l₁ = 3
+    H₁ = matrix(R, 2, 3, [x^2 x^2 x^2;
+                          x   x^2  0])
 
-    append!(test_homological_product_codes, [hpc₁, hpc₂, hpc₃, hpc₄])
+    # 3D Homological product code from [Quintavalle_2021](@cite)
+    δ₂ = matrix(GF(2), parity_matrix(RepCode(3)))
```
Krastanov: moving this δ₂ to where delta 1 is defined, might help with legibility

Krastanov (changes requested):
We need to bring back the tests that were removed, see comments. And we need to figure out a good way to filter them.

Fe-r-oz:
Thank you for your feedback! The update is as follows:
- the large codes are added back and generally all the universal tests (which are now executing in paralllel) take less time. 
- The parallel tests are ECC Code Properties, ECC Encoding, ECC Decoding, ECC Throws and
- ECC Syndrome Circuit Equivalence (test_ecc_syndromes.jl) and ECC Syndrome Measurement Correctness (test_ecc.jl). 
- The semantics for the two syndrome files is that the  (test_ecc_syndromes.jl) is testing the equivalence between different schemes and the other one  (test_ecc.jl) is focused on correctness properties syndrome measurement.
- The file that is taking more time (4 hours at time of writing) is `test_ecc.jl`. We can limit the number of codes in this file using `maxn` parameter to reduce the execution time.
- You are right that it was not the decoding file or code properties file that was taking more time; it is one of the syndrome file mentioned above.

Krastanov:
The windows tests are due to a python package related to a dependency failing to install.

The cuda failure happens on master, is unrelated, and is of unknown origin.

Thanks for pushing through with this one @Fe-r-oz , this is a very valuable and important improvement in testing.

PR #600, opened 2025-09-22
Changed petrajectories to handle operations with out of bounds indices

brohan-byte:
As of now if petrajectories receives an operation with indices beyond what the input register/state contains, it lets it go. 
I've added in a simple guard to ensure that it throws an error in such a case.

Since this guard runs only once, it isn't computationally expensive either.

### DIFF
--- a/src/petrajectory.jl
+++ b/src/petrajectory.jl
@@ -70,6 +70,11 @@ end
 
 See also: [`pftrajectories`](@ref), [`mctrajectories`](@ref)"""
 function petrajectories(initialstate, circuit; branch_weight=1.0, max_order=1, keepstates::Bool=false)
+    for circuit_op in circuit
+        if(maximum(affectedqubits(circuit_op)) > nqubits(initialstate))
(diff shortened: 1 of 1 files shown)


src/noise.jl
```diff
@@ -208,3 +208,11 @@ function applybranches(s::AbstractQCState, g::NoisyGate; max_order=1)
...
     return [(state, continue_stat, prob, order) for (state, prob, order) in applynoise_branches(news, g.noise, affectedqubits(g), max_order=max_order)]
 end
+
+function _sentinel_affectedqubits end
+QuantumClifford._sentinel_affectedqubits(x::Any) = QuantumClifford.affectedqubits(x)
+QuantumClifford._sentinel_affectedqubits(::QuantumClifford.NoiseOpAll) = missing
+
+function _sentinel_maximum end
+QuantumClifford._sentinel_maximum(x::Any) = maximum(x)
+QuantumClifford._sentinel_maximum(::Missing) = 0
```
Krastanov: these functions probably make more sense in the `affectedqubits` file

PR #602, opened 2025-09-23
address CUDA deprecation warning and simplify CI setup

Krastanov:
We seem to be getting this warning in the test runner:

image

I am making this draft PR without understanding of the difference or the reason for the change, just to see if things still work.


Krastanov:
welp... this is definitely a real failure
image

PR #606, opened 2025-09-28
Multivariate Multicycle codes using Koszul Complexes

### LINKED ISSUE #603: Multivariate Multicycle Codes
We introduce a novel quantum error correcting code, namely Multivariate Multicycle QECC whose subfamilies include:

- Bivariate Bicycle Codes, [High-threshold and low-overhead fault-tolerant quantum memory](https://arxiv.org/pdf/2308.07915)
- Trivariate Tricycle Codes, [Single-Shot Decoding and Fault-tolerant Gates with Trivariate Tricycle Codes](https://arxiv.org/pdf/2508.08191)
- Tetravariate Tetracycle Codes (first discovered in QuantumClifford.jl, #590)

Fe-r-oz:
We introduce a novel class of quantum CSS codes — *Multivariate Multicycle* codes — constructed from multivariate polynomial quotient ring formalism over finite fields over GF(2). Our discovery establishes that the boundary maps of these codes are governed by the combinatorial structure of *Koszul* complexes. Specifically, for a code defined by *t* polynomial relations, we demonstrate that the *k-th* boundary map is obtained by taking the **Koszul matrix** in degree *k* and replacing each variable entry with the corresponding circulant matrix derived from the code's defining relations. The **Koszul complex** provides the framework for the boundary map construction, ensuring the commutativity properties essential for the code construction. This correspondence reveals that multivariate multicycle codes possess the structure of **Koszul complexes** with circulant coefficients. This family of codes generalizes the bivariate bicycle, trivariate tricycle, and tetravariate tetracycle codes and it enables full single shot decoding in both X and Z directions.

This code generalizes the following code families and resolves #603
1. Bivariate Bicycle Codes
2. Trivariate Tricycle Codes
3. Tetravariate Tetracycle codes


docs/src/references.bib
```diff
@@ -1143,6 +1143,12 @@ @article{quintavalle2021single
   publisher={APS}
 }
 
+@book{eisenbud2013commutative,
+  title={Commutative algebra: with a view toward algebraic geometry},
+  author={Eisenbud, David},
+  volume={150},
+  year={2013},
+  publisher={Springer Science \& Business Media}
```
Krastanov: missing figure bracket for closing?

ext/QuantumCliffordOscarExt/multivariate_multicycle.jl
```diff
@@ -0,0 +1,210 @@
...
+ensuring the commutativity properties essential for the code construction. This correspondence reveals that multivariate
+multicycle codes can be constructed using the framework of **Koszul complexes**.
+
+This family of codes generalizes the bivariate bicycle, trivariate tricycle ([`TrivariateTricycleCode`](@ref)), and
+tetravariate tetracycle codes and it enables full single shot decoding in both X and Z directions, a capability that the
+[`TrivariateTricycleCode`](@ref) lacks.
+
+# Special Cases 
+
+## t = 2: Bivariate bicycle codes
```
Krastanov: add paper reference and a "see also"

ext/QuantumCliffordOscarExt/multivariate_multicycle.jl
```diff
@@ -0,0 +1,210 @@
...
+
+julia> B = S(y^5 + x^3  + x^19);
+
+julia> c = MultivariateMulticycleCode([l,m], [A,B]);
+
+julia> code_n(c), code_k(c)
+(756, 16)
+```
+
+## t = 3: Trivariate tricycle codes
```
Krastanov: add paper reference and a "see also"

ext/QuantumCliffordOscarExt/multivariate_multicycle.jl
```diff
@@ -0,0 +1,210 @@
...
+
+julia> C = S(1 + x^4*y^3*z^3 + x^5*z^2);
+
+julia> c = MultivariateMulticycleCode([l,m, p], [A, B, C]);
+
+julia> code_n(c), code_k(c)
+(432, 12)
+```
+
+## t = 4: Tetravariate Tetracycle Codes
```
Krastanov: mention that these are made up here, no paper, and when your paper is out make sure you reference them

Fe-r-oz:
> Please wait for me to merge:
>https://github.com/QuantumSavory/QuantumClifford.jl/pull/644
> a related PR for https://github.com/QuantumSavory/QuantumClifford.jl/issues/623
> a PR for fixing downgrade check
> After these are done, and you have the CHANGELOG updated, and tests on non-alpha versions of julia pass, feel free to merge.

Quick update on this: #644 and #646 (PR related to #623) have been merged. The downgrade CI is currently failing (unrelated to this PR). The tests on non-alpha CI pass, the changelog is updated, and some additional assertions and tests were added. Once the downgrade CI is fixed, this PR will proceed to merging

ext/QuantumCliffordOscarExt/multivariate_multicycle.jl
```diff
@@ -0,0 +1,454 @@
...
+julia> S, _ = quo(R, I);
+
+julia> A = S(x^3 + y^10 + y^17);
+
+julia> B = S(y^5 + x^3  + x^19);
+
+julia> c = MultivariateMulticycle([l,m], [A,B]);
+
+julia> code_n(c), code_k(c)
+[ Info: Constructing boundary maps for MultivariateMulticycle
```
Krastanov: Turn these to `debug` statements. They are much too verbose for being printed by default, but can be very useful when debugging what is happening.

test/test_ecc_trivariate_tricycle.jl
```diff
@@ -77,6 +77,44 @@
...
+        # A MM code is defined on n = 6*l*m*p*r data qubits.
+        @test code_n(c) == code_n(stab) == 6*l*m*p*r
+        @test code_k(c) == code_k(stab)
+        @test stab_looks_good(stab, remove_redundant_rows=true) == true
+        @test iszero(mod.(metacheck_matrix_z(c)*parity_matrix_z(c), 2))
+        @test iszero(mod.(metacheck_matrix_x(c)*parity_matrix_x(c), 2))
+        @test metacheck_matrix_z(c) != metacheck_matrix_x(c)
+        @test parity_matrix_z(c) != metacheck_matrix_z(c)
+        @test parity_matrix_x(c) != metacheck_matrix_x(c)
+        @test parity_matrix_x(c) != parity_matrix_z(c)
```
Krastanov: This seems to be the only place where we are testing the metacheck matrices for correctness.

Generally speaking, what properties do metacheck matrices obey that we can test. Can we, for instance, do a bunch of random checks that `M*H*error` is the zero vector?

We probably should add a new file that does metacheck tests on all codes. Something like

```
for c in all_codes
    for mcheck in [metacheck_matrix_x, metacheck_matrix_z]
        hasmethod(mcheck, typeof(c)) || continue
        H = parity_matrix_...(c)
        M = metacheck_...(c)
        @test M*H is a zero matrix
    end
end
```

Fe-r-oz: We are testing the metacheck properties of codes in bespoke tests but we should have a new file as well!

> Generally speaking, what properties do metacheck matrices obey that we can test. Can we, for instance, do a bunch of random checks that M*H*error is the zero vector?

X-Metacheck matrix satisfies $M_XH_X = 0$, which means that any valid noiseless $X$ syndrome $z_x$ (in the image of $H_X$ ) must satisfy $M_X z_x = 0$. Similarly, the $Z$-Metacheck matrix satisfies $M_ZH_Z = 0$, so $M_Z z_z = 0$ where $z_z$ is any valid noiseless $Z$ syndrome. We can test something like the following:

```
Let error be e
z_z = mod.(Hz*e, 2) // Z-syndrome
s_z = mod.(Mz * z_z, 2) // called metasyndrome 
@test all(s_z .== 0)
```

ext/QuantumCliffordOscarExt/multivariate_multicycle.jl
Fe-r-oz: Thank you!

test/test_ecc_metachecks.jl
```diff
@@ -1,27 +1,35 @@
...
         for c in all_codes
+            code_mx = (hasproperty(c, :D) && c.D < 4) || # for DDimensional Toric and surface codes, Mx exist for D = 4
+                (hasproperty(c, :boundary_maps) && length(c.boundary_maps) != 4) || # for homological product, length 4 means Mx exist
+                (hasproperty(c, :t) && c.t < 4) || # for MM code, t  = 4 means Mx exist, otherwise not
+                (hasproperty(c, :polynomials) && length(c.polynomials) != 4)
+
+            code_mz = (hasproperty(c, :D) && c.D < 3) || # for DDimensional Toric and surface codes, Mz exist for D = 3
+                (hasproperty(c, :boundary_maps) && length(c.boundary_maps) != 3) || # for homological product, length 3 means Mz exist
+                (hasproperty(c, :t) && c.t < 3) || # for MM code, 3  = 4 means Mx exist, otherwise not
+                (hasproperty(c, :polynomials) && length(c.polynomials) != 4) # for TT code, Mz exist since polynomials are 3
```
Krastanov: This seems very complicated. Why is it needed?

Krastanov: More importantly, it is very fragile and will break if we add more codes. Let's discuss other options when we meet.

test/test_ecc_metachecks.jl
```diff
@@ -0,0 +1,37 @@
...
+
+            @testset "Metacheck tests for $(typeof(c))" begin
+                if hasmethod(metacheck_matrix_x, Tuple{typeof(c)}) && !code_mx
+                    Hx = parity_matrix_x(c)
+                    Mx = metacheck_matrix_x(c)
+                    @test iszero(mod.(Mx*Hx, 2))
+                end
```
Krastanov: how about something like this
```
mx, mz = hasmetachecks(c)
if mx
    Mx
    Hz
    @test iszero(...)
end
...
```

where we need to define

```
hasmetachecks(c)::Tuple{Bool, Bool}
```

Krastanov:
I pushed some cleanup changes, make sure to pull them before making further changes. Mostly, I tried to clean up the Oscar import statement. It was very messy, and it was actually importing a bunch of things that are defined in Base, not in Oscar.

Krastanov:
I also removed `hasmethods` from the list of exported symbols. We should not export brand new things until we know they will be part of the public API, otherwise making non-breaking changes in the future becomes very difficult.

test/test_jet.jl
```diff
@@ -6,6 +6,6 @@
 
     rep = JET.report_package(QuantumClifford, target_modules=[QuantumClifford, QuantumInterface])
     @show rep
-    @test length(JET.get_reports(rep)) <= 5
+    @test length(JET.get_reports(rep)) <= 11
```
Krastanov: Do not make changes here unless you know that the new ones are (a) false positives AND (b) caused by the code you added. Otherwise we just get in the habit of not looking at JET tests. We should have a separate PR investigating why these are happening.

lib/QECCore/src/interface.jl
```diff
@@ -464,3 +464,5 @@ code C of length n over ``\\mathbb{F}_q``, g(x) satisfies:
 The input is a classical polynomial error-correcting code defined over a finite field.
 """
 function generator_polynomial end
+
+hasmetachecks(c::AbstractQECC) = (nothing, nothing)
```
Krastanov: The API you have picked does not really match the name. `hasmetachecks` is a question that should be answered true/false or maybe true/false/nothing (nothing corresponding to not knowing).

Let's rename it to `metacheck_matrices` if you want to keep the current return type.

test/test_ecc_metachecks.jl
```diff
@@ -0,0 +1,18 @@
...
+
+    include("test_ecc_base.jl")
+
+    all_codes = all_testable_code_instances()
+    function test_all_codes_metachecks(all_codes)
+        for c in all_codes
+            Mx, Mz = hasmetachecks(c)
+            !isnothing(Mx) && @test iszero(mod.(Mx*parity_matrix_x(c), 2))
+            !isnothing(Mz) && @test iszero(mod.(Mz*parity_matrix_z(c), 2))
+            isnothing(Mx) && isnothing(Mz) && @info "No metacheck matrices defined for $(typeof(c))"
```
Krastanov: This is not a useful log message. It will make the test log very noisy without contributing understanding.

Krastanov: I would prefer if it is just removed

lib/QECCore/src/interface.jl
Fe-r-oz: Oh...  I have corrected the use of `hasmetachecks(all_codes)` to return a tuple `(false, false)`, then it makes it easier to keep track of which codes only has Mx/Mz (or both or none) which corresponds to the second entry of tuple i. e. TT code has Mz so we use `(false, true)`

test/test_ecc_metachecks.jl
Fe-r-oz: removed!

Fe-r-oz:
>I pushed some cleanup changes, make sure to pull them before making further changes.

Thank you for the cleanup!

> Is there anything else that you plan to modify here?

There are no plans for further modifications to the core of the API (that is not related to metacheck new file API we are currently looking into)

test/test_ecc_metachecks.jl
```diff
@@ -0,0 +1,25 @@
...
+    include("test_ecc_base.jl")
+
+    all_codes = all_testable_code_instances()
+
+    function test_all_codes_metachecks(all_codes)
+        for c in all_codes
+            has_mx, has_mz = hasmetachecks(c)
+            if has_mz && hasmethod(metacheck_matrix_z, Tuple{typeof(c)})
+                Mz = metacheck_matrix_z(c)
+                !isempty(Mz) && @test iszero(mod.(Mz*parity_matrix_z(c), 2))
```
Krastanov: Why is the `isempty` check necessary? Where/why does this happen?

test/test_jet.jl
Krastanov: bump on this one -- please undo the change unless you can explain how it fits (a) or (b) above

test/test_ecc_metachecks.jl
Fe-r-oz: Good catch! It is now removed

Krastanov:
sigh... gosh our luck.

All these failures are happening because Ncurses_jll v6.6.0 was released 9 hours ago, which does not seem to work inside of the podman container we use for buildkite. 

https://github.com/JuliaPackaging/Yggdrasil/pull/12884

I think I have a workaround. I will push something here.

Krastanov:
I give up... I can not fix this. I will just make a compat bound to use only the older Ncurses_jll which did not have this problem.

Krastanov:
Thanks, Feroz, this is an awesome addition!

PR #614, opened 2025-10-17
implement ptrace

### LINKED ISSUE #605: Tracing out in the QuantumClifford formalism
I would naively expect that running

```
stab = S"XX
        ZZ"
stab = tensor_pow(stab, 2)
stab = MixedDestabilizer(stab)
projectXrand!(stab, 4)
traceout!(stab, 4)
projectXrand!(stab, 3)
traceout!(stab, 3)
stab

```

would yield a two-qubit stabilizer group on systems 1 and 2 (since I am measuring and tracing out single qubits). Instead, the output is

`𝒟ℯ𝓈𝓉𝒶𝒷
+ Z___
+ _X__
𝒳ₗ━━━━
+ __ZZ
+ ___Z
𝒮𝓉𝒶𝒷━━
+ XX__
+ ZZ__
𝒵ₗ━━━━
+ __X_
+ __XX`

i.e. a four-qubit stabilizer group.

It would be nice to have a separate function or potential renaming that allows me to end up with just the two-qubit system of interest.

There is one complication I can imagine---what happens when you trace out a qubit that is entangled with the rest? To prevent this, one can add a check to see if the qubit(s) is entangled with the other qubits (which is efficient).

Alternatively there are two 'reasonable' operations that one can apply to stabilizer groups that are related to partial traces, either https://qc.quantumsavory.org/stable/API/#QuantumClifford.delete_columns-Tuple{Stabilizer,%20Any} or https://qc.quantumsavory.org/stable/API/#QuantumClifford.contractor-Tuple{Stabilizer,%20Any}. These agree when a qubit/column is not entangled with the rest. 
I am not sure what is wise to do when the qubit is entangled, since then these two operations have different interpretations, and it depends on the context 'which one is the right one to use'. 

For the moment, I think adding a check would be most convenient?

This is also related to issue #604.

Krastanov:
fixes #605

`ptrace` for partial traces. We already had `traceout!`, which returns tableaux with the same number of qubits. `ptrace` removes the traceout qubits besides setting their stabilizing operators to `I`.


Krastanov:
actually, the install is a bit more difficult because of another package that also needs to be release together with QuantumClifford. Here is the command that actually works

```
using Pkg
Pkg.pkg"add QECCore#master QuantumClifford#master"
```

PR #615, opened 2025-10-20
Non-Clifford implementation of sparsification framework, sum over clifford and enhanced magic states

riteshbhirud:
This implements sparsification framework, sum over clifford and enhanced magic states
### 1. **Sparsification Framework** (Section 5.2)
- Implements Sparsification Lemma (Lemma 6) for reducing dense stabilizer decompositions
- Achieves χ_δ(ψ) ≤ 1 + ||c||₁²/δ² bound (Theorem 1)
- Controllable approximation error via δ parameter

### 2. **Enhanced Magic State Library** (Section 5.3)
- **T-gate**: Optimal decomposition with ξ(T) ≈ 1.172
- **CCZ gate**: Optimal 8-term group decomposition with ξ(CCZ) = 16/9 ≈ 1.778
- **R(θ) rotations**: Parametric decomposition with ξ(R(θ)) = (cos(θ/2) + tan(π/8)sin(θ/2))²
- Implements Lifting Lemma for converting magic states to gate decompositions

### 3. **Sum-over-Cliffords Simulation** (Section 2.3.2)
- Decomposes circuits as U = Σⱼ cⱼKⱼ where Kⱼ are Clifford operations
- Multiplicative cost scaling: O(δ⁻² ∏ᵐⱼ₌₁ ξ(Vⱼ))
- Handles arbitrary mixed Clifford/non-Clifford circuits

### 4. **Fast Metropolis Sampling** (Section 4.2)
- O(kn) complexity per sample (vs naive O(k²n³))
- Auto-tuned burn-in and thinning parameters

### DIFF
--- a/src/QuantumClifford.jl
+++ b/src/QuantumClifford.jl
@@ -95,7 +95,7 @@ export
     # petrajectories
     petrajectories, applybranches,
     # nonclifford
-    GeneralizedStabilizer, UnitaryPauliChannel, PauliChannel, pcT, pcPhase,
+    GeneralizedStabilizer, UnitaryPauliChannel, PauliChannel, pcT, pcPhase, pcRx,
     # makie plotting -- defined only when extension is loaded
     stabilizerplot, stabilizerplot_axis,
     # sum types
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Add `mul_right!` methods for inplace operations between tableaus
 - Add a `CliffordOperator` constructor that builds a dense clifford from a `PauliOperator`
 - Add a `phases` getter for `CliffordOperator`
+- A new non-Clifford operation was added as a Pauli channel, the rotation gate `pcRx`.
 - Adapt.jl can now be used to convert various types to GPU-backed storage.
 - The phase storage type can now be parameterized, instead of hardcoded to UInt8.
 - Add an extension to `QECCore` -- `QECCoreNemoExt` for accurate matrix `rank` computation
--- a/src/nonclifford.jl
+++ b/src/nonclifford.jl
@@ -812,3 +812,10 @@ function pcPhase(ϕ)
 end
 
 const pcT = pcPhase(π/4)
+
+function pcRx(θ)
+    UnitaryPauliChannel(
+        (I, X),
+        (cos(θ/2), -im*sin(θ/2))
+    )
+end
--- a/test/test_nonclifford_quantumoptics.jl
+++ b/test/test_nonclifford_quantumoptics.jl
@@ -184,36 +184,44 @@
         for n in num_qubits  # Exponential cost in this term
             for repetition in 1:num_trials
                 test_phases = [π/4, π/8, π/2, π, 2π]
+                test_gates = [pcPhase, pcRx]
                 for ϕ in test_phases
-                    stab = random_stabilizer(n)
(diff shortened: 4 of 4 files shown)


src/non_clifford_simulation.jl
```diff
@@ -0,0 +1,451 @@
...
+nqubits(::TGate) = 1
+nqubits(::CCZGate) = 3
+
+"""
+    identify_gate_type(op::AbstractOperation)
+
+Gate classification using direct isa checks.
+Identifies all Clifford operations and distinguishes parametric non-Clifford gates.
+"""
+function identify_gate_type(op::AbstractOperation)
```
Krastanov: This is not really extendable or in the style of Julia multimethods. It should just become a function/trait that anyone can define a method for.

src/non_clifford_simulation.jl
```diff
@@ -0,0 +1,451 @@
...
+    return :non_clifford
+end
+
+"""
+    extract_gate_parameters(op::AbstractOperation)
+
+Extract parameters from gates using QuantumClifford.jl's structure.
+Returns (gate_type, parameters, qubits) for both Clifford and non-Clifford gates.
+"""
+function extract_gate_parameters(op::AbstractOperation)
```
Krastanov: This is not really extendable or in the style of Julia multimethods. It should just become a function/trait that anyone can define a method for.

src/QuantumClifford.jl
```diff
@@ -95,6 +95,14 @@ export
...
     # nonclifford
     GeneralizedStabilizer, UnitaryPauliChannel, PauliChannel, pcT, pcPhase,
+    # Non-Clifford simulation - sum-over-Cliffords
+    TGate, CCZGate,
+    SparsifiedState, MagicStateDecomposition, CliffordGateDecomposition,
+    SimulationResult, QuantumSimulationResults, BitString,
+    sparsify_stabilizer_decomposition, estimate_sparsification_quality,
+    decompose_T_gate, decompose_CCZ_gate, decompose_rotation_magic_state,
+    simulate_non_clifford_circuit, sample_measurement_outcomes,
+    display_results, estimate_simulation_cost,
```
Krastanov: this is too many public things -- we should cut it down to the essentials

Krastanov (changes requested):
1. Move everything to a module called `LowRankNonClifford` or a similar name
in a subfolder

2. Keep these public:

- TGate
- CCZGate
- simulate_non_clifford_circuit
- estimate_simulation_cost
- QuantumSimulationResults

potentially will rename them

Keep consistent names and APIs with the style of use in Pauli Frame simulations, e.g. pftrajectories and getters related to extracting samples information

3. Rename or restructure

- BitString -> Just a normal `Vector{Bool}` or even just a `BitVector`
- MagicStateDecomposition -> MagicStateDecompositionCache
- remove CliffordGateDecomposition if not used (and anything else not used anymore)

4. Use existing types and multimethods instead of building an independent type system

We can introduce a `isunitary` and `isclifford` and similar methods.

5. `display_results` should be a show method (see other existing ones)

6. add to this page of the docs

https://qc.quantumsavory.org/stable/allops/

7. have a tests that relies on purely clifford circuits and compares against mctrajectories (using Register and explicit measurements)

8. have a test that relies on QuantumOpticsBase (for very few qubits so it can run efficiently)

PR #622, opened 2025-11-04
tests: centralized Oscar-required tag

Fe-r-oz:
This PR adds the centralized Oscar tag in `runtest.jl`

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -36,6 +36,7 @@ if any((CUDA_flag, ROCm_flag, OpenCL_flag))
          )
 end
 Oscar_flag && Pkg.add("Oscar")
+Oscar_compatible = !Sys.iswindows() && Sys.ARCH == :x86_64 && VERSION >= v"1.11"
 using TestItemRunner
 using QuantumClifford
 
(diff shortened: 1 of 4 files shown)


test/runtests.jl
```diff
@@ -36,6 +36,7 @@ if any((CUDA_flag, ROCm_flag, OpenCL_flag))
          )
 end
 Oscar_flag && Pkg.add("Oscar")
+Oscar_compatible = !Sys.iswindows() && Sys.ARCH == :x86_64 && VERSION >= v"1.11"
```
Krastanov, line 39: I have the impression that Oscar_flag and Oscar_compatible are the same. If that is true, you can just use the one that already exists.

test/runtests.jl
```diff
@@ -49,6 +50,10 @@ testfilter = ti -> begin
         push!(exclude, :jet)
     end
 
+    if !Oscar_compatible
+        push!(exclude, :oscar_required)
+    end
+
```
Krastanov, line 56: I do not think this is actually used for anything during :ecc tests. Maybe the various return statements below should be updated to add `&& all(!in(exclude), ti.tags)` in which case `:oscar_required` will actually be used.

PR #634, opened 2025-12-04
Add AnythingLLM chat embed to docs

Krastanov-agent:
## Summary
- add docs/anythingllm.jl to manage AnythingLLM workspace creation, document/docstring uploads, and embed config
- workspace names now include the package version and deploy target (dev/stable/tag/PR) and uploads are moved into a folder matching the workspace slug
- AnythingLLM sync only runs when Documenter’s deploy decision would deploy (same defaults as deploydocs; otherwise skipped)
- upload cleanup deletes embeds, workspaces, and uploaded docs to avoid accumulation; linkcheck remains disabled for simpler local runs
- API key is not hardcoded; must be provided via ANYTHINGLLM_API_KEY; default allowlist is empty; embed defaults to query mode
- helper functions now carry detailed docstrings for clarity

## Testing
- julia --project=docs -q -e 'include(docs/anythingllm.jl); using .AnythingLLMDocs; AnythingLLMDocs.integrate_anythingllm(QuantumClifford, Module[], joinpath(@__DIR__,docs,src))' *(skips sync locally due to deploy decision)*

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -52,7 +52,7 @@ format = Documenter.HTML(
     assets = anythingllm_assets,
 ),
 modules = doc_modules,
-warnonly = [:missing_docs],
+warnonly = [:missing_docs, :linkcheck],
 linkcheck = true,
 authors = "Stefan Krastanov",
 pages = [


Krastanov:
the initial prompt was:

Documenter.jl is the typical way for julia libraries to write and build their documentation. I want to enable an LLM chat inside of docs built with Documenter.jl.

For the chat service I will use anythingllm, served from a local instance.

The local instance of anythingllm is https://anythingllm.krastanov.org/ 
The API is at https://anythingllm.krastanov.org/api/v1/
The API docs are at https://anythingllm.krastanov.org/api/docs/ -- you should look through the docs.
The API key is QYR3732-TWCM91S-NQZS080-SYNSJ3K
At https://anythingllm.krastanov.org/api/v1/embed you can see already existing "embeds". "Embeds" is what anythingllm calls pre-configured embedable LLM chat instances.

As a test case, let's use the QuantumClifford.jl library.

In Julia, you can use `pkg"dev --local <Package Name>"` in order to get a local clone of a library. Create such a local clone for Documenter (so that you can read through its documentation) and for QuantumClifford (so that you can start implementing the LLM chat).

To build docs you can use the following from the folder of a given package:
julia --project=docs -e '
            using Pkg
            Pkg.develop(PackageSpec(path=pwd()))
            Pkg.instantiate()';
julia --project=docs docs/make.jl

Your overall task is to enable an anythingllm chat inside of the docs of QuantumClifford.jl. Here is a suggested way to do this:

1. Create a new file in the docs folder, named anythingllm.jl that will be included from make.jl. That file should implement functions to create a new workspace in anythingllm and create an embedded chat that uses the newly created workspace. The workspace should be named after the package. The new function should be called in make.jl during doc building. Make sure that any existing workspaces of the same name are first deleted (together with all files stored in those workspaces).

2. Ensure that all markdown source files in the docs are uploaded to anythingllm for embedding in vector databases for use in a RAG.

3. Ensure all public docstrings that were found are also uploaded, each as a separate file.

4. Finally, add the embeddable chat script to the doc page as a RawHTMLHeadContent in the list of assets.

You already have access to an authenticated gh. Use it for creating PRs. Use it for enabling git authentication as well.

Make a PR when you are done.

PR #643, opened 2025-12-18
add tesseract-decoder extension

Krastanov:
PyTesseractDecoder.jl is still being registered. When registered, the sources entries to Project.toml can be removed and this can be merged.

cc @Fe-r-oz


Krastanov:
The registration link for PyTesseractDecoder https://github.com/JuliaRegistries/General/pull/144670

Issue #623, opened 2025-11-04
Polish the summer additions to the ECC API


Krastanov:
@Fe-r-oz , @nzy1997 , I finally got around to reviewing the vast array of codes you have added/reviewed over the last few months. Great work! Thank you so much for it!

As the preparation of the paper is coming up, and we should also plan on having a release for the library, I wanted to go through some cleanup chores and review. Especially given that this is a breaking release, we can use it to make things a bit more consistent. There seems to be a few unidiomatic use of structs that have been merged that would need to be fixed before the next release. Posting them here to keep track and in case you have the bandwidth to address them:

## Many structs constructors are not deterministic

A structure should be a deterministic complete representation of a code instance. There should be no randomness involved. The values of a field of a struct should completely define it.

- [x] The GoppaCode has a  `seed` argument. This has to be removed as it has little to do with **what** the code instance is, rather it is implementation details on how it is implemented here.
    - [x] remove `seed`
    - [x] turn `GoppaCode(m, t; seed)` into `random_goppa_code(m, t)` that dispatches to `random_goppa_code(rng, m, t)` (no seeds, just rng instances)

- [x] it is not clear to me that `BivariateBicycleCode` and `GeneralizedCircuilantBivariateBicycleCode` should be separate structures or be named the way they are. We need to clarify their relationship.
    - [x] the docstring for BivariateBicycleCode has a reference to `GeneralizedCirculantBivariateBicycleCode` , but it is not super clear whether this should just be a different constructor (why "generalized") -- on the other hand the docstring for GeneralizedCirculantBivariateBicycleCode explains that really well! Copying that one sentence could help.
    - [x] `BivariateBicycleCode` has a weird unicode field name -- unicode is cool, but this seems like an inconsistent use as just one letter is unicode, while that particular letter also has a perfectly reasonable ascii character
    - [x] I really like the description of alternatives in BivariateBicycleCode's docstring -- very thorough and well done. We should probably make a diagram to clarify it
    - [x] one of them is in QuantumClifford.ECC and the other one is QECCore -- can we have both in the same location before the release
    - [x] the name seems a bit inconsistent for the `GeneralizedCirculantBivariatBicycle`. Why add `Circulant`? Why remove `Code`?

- [x] `GeneralizedBicycleCode` and `generalized_bicycle_codes` -- why are there both
[...]

PR #647, opened 2025-12-23
removing the `QuantumTannerGraphProduct` in favour of existing `hgp`

Fe-r-oz:
Removing the `QuantumTannerGraphProduct` in favour of existing `hgp`: Ref. #623


Krastanov:
Thanks for submitting this Feroz! I have been struggling to think through which option is better -- having both, having hgp only, or having this QTGP type only. On reading through the changes here, I think I am mostly convinced that your initial setup of having these separate was a good design decision. My main reason to say that is that `QTGP(H1, H2)` is much more informative than `CSS(hgp(H1, H2)...)` and can encode much more useful information in the type. Apologies for vacillating on this one. For the moment I will close this PR, but let me know if you disagree.

PR #660, opened 2026-01-31
2D Tile codes, a slight generalization of 2D Surface codes

### LINKED ISSUE #657: 2D Tile codes
Tile codes are generalization of surface codes that offers flexibility in terms of locality and stabilizer weight without compromising 2D locality with open boundary conditions. These codes outperform surface codes and planar versions of bivariate bicycle codes. BB codes have a modest amount of non-locality because of periodic boundary conditions of the 2D lattice. These hardware issues caused by non-locality has been recently addressed in new layout such as [here](https://quantum-journal.org/papers/q-2025-05-05-1728/ ) or other means. Nevertheless, if we want to avoid the assumption of periodic boundary conditions, tile codes present an new venue of code design with open boundary conditions.

The rates of these codes is not that high compared to some other codes. However, they offer the following benefit:

> Tile codes do not require periodicity and thus have true O(1)-locality on a planar 2D lattice. With that, tile codes address this major challenge in quantum error correction, achieving high efficiency while preserving implementability on near-term hardware.

Implementing it would be valuable for comparison with other classes of modern error correcting codes.

reference: https://arxiv.org/pdf/2504.09171

Fe-r-oz:
This PR implements 2D Tile codes from [reference](https://arxiv.org/pdf/2504.09171) and resolves #657. These codes are generalizations of 2D surfaces codes and have O(1) locality on a 2D lattice.


lib/QECCore/src/codes/quantum/tile2d.jl
```diff
@@ -0,0 +1,128 @@
...
+    """Positions of horizontal edges within the tile box."""
+    horiz::Vector{Tuple{Int,Int}}
+    """Positions of vertical edges within the tile box."""
+    vert::Vector{Tuple{Int,Int}}
+    """Number of tiles along the x-direction."""
+    Lx::Int
+    """Number of tiles along the y-direction."""
+    Ly::Int
+
+    function Tile2D(B::Int, horiz::Vector{Tuple{Int,Int}}, vert::Vector{Tuple{Int,Int}}, Lx::Int, Ly::Int)
```
Krastanov, line 43: Why is this constructor necessary? It seems to be just what the default constructor would do anyway.

lib/QECCore/test/test_codes.jl
```diff
@@ -90,4 +90,8 @@
     @testset "Circulant Bivariate Bicycle Codes" begin
         include("codes/generalized_circulant_bivariate_bicycle.jl")
     end
+
+    @testset "Tile 2D" begin
+        include("codes/tile2d.jl")
+    end
```
Krastanov: this confuses me... `testset`s should not contain `testitem`s...

Testitems get detected automatically without the need for include statements. Is this entire file nonsensical?

src/ecc/ECC.jl
```diff
@@ -44,7 +44,7 @@ export parity_checks, parity_matrix_x, parity_matrix_z, iscss,
     GeneralizedBicycle, ExtendedGeneralizedBicycle,
     HomologicalProduct, DoubleHomologicalProduct,
     GeneralizedToric, TrivariateTricycle, BivariateBicycleViaPoly,
-    MultivariateMulticycle,
+    MultivariateMulticycle, Tile2D,
```
Krastanov: if you are going to import Tile2D into quantumclifford, firt you need to release a new version of QECCore, then bump the compat bound for it in quantumclifford

Krastanov (changes requested):
This looks like a pretty useful addition, but it requires a bit more work, including updating the CHANGELOG of both libraries and making sure that QECCore is released so that QuantumClifford can have proper dependency bounds (that part probably will need a whole second PR)

lib/QECCore/test/test_codes.jl
Fe-r-oz: I did not pushed this test file, but I kept using this file haha. You are right

Krastanov: yeah, it seems that whole file is nonsensical -- we should verify that all the tests it "includes" have their own testitem and if they do we can just delete the entire file

lib/QECCore/src/codes/quantum/tile2d.jl
Fe-r-oz: Not really necessary, but cosmetic purpose. Maybe we can add some error messages that are thrown -- for example negative size of box is inputed or something? I probably had something in mind when creating this inner constructor for making sure user only enters valid parameters -- will check the reference if there is some relations on parameters, violating which should throw errors

lib/QECCore/test/test_codes.jl
Fe-r-oz: Agreed. Silly me for not flagging this file earlier. Although I have had my suspicions about this test file

src/ecc/ECC.jl
Fe-r-oz: Okay, let's just remove it from here for now

Fe-r-oz:
Sounds good. I would turn it back to non-draft at this time. I can improve the documentation to it can be easier to see what some stuff mean for users(like how to visualize the X and Z configurations)  who are reading it for very first time.

Krastanov:
@Fe-r-oz , this was not reviewed. Please do not merge things without review. If you are impatient, just bump other devs about a review, but do not skip the review.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 
 # News
 
+## v0.11.4 - dev
+
+- `2DTile` codes, a slight generalization of 2D `Surface` codes, are now implemented in `QECCore`.
```
Krastanov, line 10: This changelog entry seems to be in the wrong file. QECCore has its own changelog

Krastanov (changes requested):
The code is not added to the large list of codes in the basic checks. I thought that should raise an error, I am a bit worried that it did not. 

We should either undo this merge so the master can be clean or fix the issues ASAP. 

Please never merge without an explicit approval in a review.

Krastanov:
Also not that you merged this while buildkite checks were disabled. The majority of tests were not ran at all

Fe-r-oz:
The buildkite was recently was disabled anyways, and all the tests passed the last time the tests ran with buildkite.

Krastanov:
But the tests were not executed for the merge and there was no review. Neither one of these is acceptable -- the probability of a mistake is small, but it is not zero, and I have to deal with a lot of PRs, not just this one, so that probability does add up for me.

Fe-r-oz:
Thank you for clarifying the review and merge process. I understand the issue and will follow required review process going forward.

PR #682, opened 2026-03-04
print generalized stabilizer in lexicographic order (to stabilize doctests)

Krastanov-agent:
## Summary
- sort `GeneralizedStabilizer` `show` output deterministically
- sort branches by descending `abs(coefficient)` and then lexicographically by the rendered `Pᵢ | Pⱼ` entry
- refresh affected doctest snapshots in `src/nonclifford.jl`
- make one range-print doctest stable across Julia versions by printing explicitly in `src/grouptableaux.jl`

## Verification
- `julia +release -tauto --project=/tmp/qc-release-doctest -e 'using Documenter, QuantumClifford, QuantumInterface; ENV["LINES"]="80"; ENV["COLUMNS"]="80"; DocMeta.setdocmeta!(QuantumClifford, :DocTestSetup, :(using QuantumClifford; using QuantumClifford.ECC); recursive=true); modules=[QuantumClifford, QuantumClifford.ECC, QuantumInterface]; doctest(nothing, modules; doctestfilters=[r"(QuantumClifford\\.|)"])'`
- `julia +alpha -tauto --project=/tmp/qc-alpha-doctest -e 'using Documenter, QuantumClifford, QuantumInterface; ENV["LINES"]="80"; ENV["COLUMNS"]="80"; DocMeta.setdocmeta!(QuantumClifford, :DocTestSetup, :(using QuantumClifford; using QuantumClifford.ECC); recursive=true); modules=[QuantumClifford, QuantumClifford.ECC, QuantumInterface]; doctest(nothing, modules; doctestfilters=[r"(QuantumClifford\\.|)"])'`

Both doctest runs pass.

PR #645, opened 2025-12-22
Non-Clifford implementation of sparsification framework, sum over clifford and enhanced magic states

riteshbhirud:
This implements sparsification framework, sum over clifford and enhanced magic states
### 1. **Sparsification Framework** (Section 5.2)
- Implements Sparsification Lemma (Lemma 6) for reducing dense stabilizer decompositions
- Achieves χ_δ(ψ) ≤ 1 + ||c||₁²/δ² bound (Theorem 1)
- Controllable approximation error via δ parameter

### 2. **Enhanced Magic State Library** (Section 5.3)
- **T-gate**: Optimal decomposition with ξ(T) ≈ 1.172
- **CCZ gate**: Optimal 8-term group decomposition with ξ(CCZ) = 16/9 ≈ 1.778
- **R(θ) rotations**: Parametric decomposition with ξ(R(θ)) = (cos(θ/2) + tan(π/8)sin(θ/2))²
- Implements Lifting Lemma for converting magic states to gate decompositions

### 3. **Sum-over-Cliffords Simulation** (Section 2.3.2)
- Decomposes circuits as U = Σⱼ cⱼKⱼ where Kⱼ are Clifford operations
- Multiplicative cost scaling: O(δ⁻² ∏ᵐⱼ₌₁ ξ(Vⱼ))
- Handles arbitrary mixed Clifford/non-Clifford circuits

### 4. **Fast Metropolis Sampling** (Section 4.2)
- O(kn) complexity per sample (vs naive O(k²n³))
- Auto-tuned burn-in and thinning parameters

PR #689, opened 2026-03-05
Non-Clifford sparsification framework (rebased PR 645)

Krastanov-agent:
## Summary
- Rebased version of #645 on top of current master, resolving merge conflicts
- Adds non-Clifford simulation via low-rank stabilizer decomposition (sparsification framework, sum over Clifford, enhanced magic states)
- Moves `nonclifford.jl` to `src/lowrank/PauliChannelNonClifford.jl` as a submodule
- Adds new `src/lowrank/PureNonClifford.jl` submodule with `TGate`, `CCZGate`, `lrtrajectories` and related functionality
- Includes master's printing changes (lexicographic ordering of GeneralizedStabilizer display from #682)

## Changes from PR #645
- Resolved merge conflict in `docs/src/references.bib` (kept both master's new references and PR's nonclifford reference)
- All PR commits rebased cleanly on top of master (including the printing/doctest changes from #682)

## Test plan
- [x] PauliChannelNonClifford tests pass (485 tests)
- [x] PureNonClifford/lowrank tests pass (225 tests)
- [ ] Full CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/test/test_lowrank_purenonclifford.jl
+++ b/test/test_lowrank_purenonclifford.jl
@@ -3,7 +3,8 @@
     using QuantumClifford.PureNonClifford
     import QuantumClifford.PureNonClifford:
         get_gate_decomposition,
-        CliffordGateDecompositionCache
+        CliffordGateDecompositionCache,
+        lrcost, stabilizer_extent
 
     @testset "Pure Clifford Circuit" begin
         circuit_clifford = [sHadamard(1), sCNOT(1,2), sZ(1)]
@@ -209,7 +210,8 @@ end
     import QuantumClifford: AbstractOperation
     import QuantumClifford.PureNonClifford:
         sample_measurement_outcomes,
-        compute_outcome_frequencies
+        compute_outcome_frequencies,
+        lrcost, stabilizer_extent
 
     @testset "Measurement Sampling Tests" begin
         state = PureGeneralizedStabilizer(2, 0.2)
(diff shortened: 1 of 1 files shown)


Krastanov:
Here is the prompt I am thinking of for the final set of refactoring:

In PR 689 of QuantumClifford.jl I want to additionally refactor the API to match the rest of the QuantumClifford.jl library a bit better. I want the following changes:

1. Make `PureGeneralizedStabilizer` public and give it a (documented) constructor `PureGeneralizedStabilizer(nqubits)` that just makes an initial state of that many qubits.
2. Remove the `lrstate` function (and remove it from the exports). It is a very short function that is used only from `lrtrajectories` and can just as well be inlined in it.
3. Remove the time-tracking `total_runtime` and related variables. The user can just use the standard `@time` macro if they need this.
4. Remove the lrtrajectories(::AbstractVector) method which simply checks that all operations are AbstractOperation -- this would break potential future extensions by external libraries.
5. For the remaining lrtrajectories method, have its first argument be untyped so any iterator over operations can be given to it.
6. Rename lrtrajectories to `cbtrajectories` for "computational basis" and rename `LRTrajectoryResults` to `ComputationalBasisMeasurementTrajectoryResults`. Rename `lrmeasurements` to `cbmeasurements`.

In these changes use simple obvious code, avoid complications. The overall goal is to make the code simpler and shorter. Stop, report, and ask for directions if a step becomes too complicated.

Add appropriate new tests at each step of the refactor. Keep new tests extremely simple and easy to read, without creating mock infrastructure. Run tests in between each step.

Make frequent small commits to the same branch for well-separated steps of the refactoring process with detailed commit messages.

Make a plan and execute these changes. Above everything else, simplifying the code is most important, making it more easily matching of the rest of the APIs in the library.

PR #696, opened 2026-03-11
Add In-Place `decode!` and Remove CSS `vcat` Allocations

### LINKED ISSUE #217: In-place and batch `decode` functions
In #212 we provide a number of decoders, however the current interface has performance detriments:

- [ ] we need to have an in-place `decode!` method that does not allocate a new array each time it outputs a guess
- [ ] we need a (in-place) bach `decode_batch` method that can take advantage of vectorization and SIMD (or at least it can take advantage of multi-threading)
- [ ] the current `vcat` used in the decoders is pretty bad for performance and a good first place to try a fix

arnavk23:
Adds in-place decoding APIs for ECC decoders by introducing `decode!` and `batchdecode!` (plus decode_batch aliases), addressing allocation overhead in hot paths.
Refactors CSSTableDecoder to avoid `vcat` and write directly into preallocated output buffers, and adds optional threaded batch decoding for table-based decoders.
Includes focused decoder tests in `test_ecc_decoder_all_setups.jl` covering in-place, batch, alias, and threaded behavior.

Fixes #217


arnavk23:
@Krastanov Plan to merge reinterpret pr first, before you review this. Making it a draft for now.

PR #398, opened 2024-10-19
tests: Group Presentation for Cyclic Group Cₘₕ = Cₘ × C₂

Fe-r-oz:
This PR reproduces the Table 2 by using presentation of cyclic group instead of Direct Product of cyclic groups which was done in: https://github.com/QuantumSavory/QuantumClifford.jl/pull/392

As $C_{mh} = C_m \times C_2 = \langle x, s \mid x^m = s^2 = xsx^{-1}s^{-1} = 1 \rangle$, there are two ways to reproduce Table 2:

1) $C_m \times C_2$
2) $\langle x, s \mid x^m = s^2 = xsx^{-1}s^{-1} = 1 \rangle$

This PR implements the 2) approach. This second approach was not mentioned in the caption of Table 2 but was referenced in the text of the Appendix. Both gives the same results. 

- [x] The code is properly formatted and commented.
- [x] Substantial new functionality is documented within the docs.
- [x] All new functionality is tested.
- [x] All of the automated tests on github pass.
- [x] We recently started enforcing formatting checks. If formatting issues are reported in the new code you have written, please correct them. <small>There will be plenty of old code that is flagged as we are slowly transitioning to enforced formatting. Please do not worry about or address older formatting issues -- keep your PR just focused on your planned contribution.</small>

PR #697, opened 2026-03-14
Abelian and non-Abelian Mirror codes

### LINKED ISSUE #694: Mirror codes
[Mirror codes](https://arxiv.org/pdf/2603.05496) are a new class of QEC codes that are slight generalization of 2BGA codes.

Fe-r-oz:
[Mirror](https://arxiv.org/pdf/2603.05496) codes are now implemented. All the code instances from Table I will be added to the tests. This reference also has nice propositions/theorems etc. that could be turned into valuable tests. Also, even though these codes are non-CSS, some of these codes are equivalently *CSS via Hadamards*.


Fe-r-oz:
Turning to draft since the tests pass on buildkite

Fe-r-oz:
Turning from draft to check the specific bespoke tests on CI (particularly, distance computation)

Fe-r-oz:
Turning back to draft to work on documentation, docstrings, adding tests to ecc_base

Fe-r-oz:
Turning from draft to test on all CI tests.

ext/QuantumCliffordOscarExt/mirror.jl
```diff
@@ -0,0 +1,44 @@
+# documentation in a bit
```
Krastanov, line 1: This one does not seem quite finished. I will ask codex to make a suggestion if you can then vet it.

Krastanov:
continuing the current review march that started with #398 and #402 

This is a really neat addition -- thanks to your work this library is pretty much the biggest public resource when it comes to a repository of codes.

Here there were some more substantial issues, including missing docs. As with the previous ones, I asked codex for a first pass and you can see the 4 commits (each reviewable on its own with detailed explanation of the specific issue being fixed): https://github.com/QuantumSavory/QuantumClifford.jl/pull/795/commits

1. https://github.com/QuantumSavory/QuantumClifford.jl/pull/795/changes/b13c8e73af2a4f162def8c81d6cd1f9bb1f4e643
2. https://github.com/QuantumSavory/QuantumClifford.jl/pull/795/changes/4a6e84932980c98488d394d560819c4fc6fe00cb
3. https://github.com/QuantumSavory/QuantumClifford.jl/pull/795/changes/370fb1fbe89cd0047485313df81ebb98f3cf013f
4. https://github.com/QuantumSavory/QuantumClifford.jl/pull/795/changes/8f26cd3a88d46a89ba1b7ccd341f6b88342ba3cf

Point 1 is about mismatches between the mathematics you are implementing and the Oscar & QuantumClifford conventions we have. Here is the detailed explanation by codex:

```
The commit was needed because the original implementation encoded the right mathematical idea, but its group representation and parity-matrix API were not reliable in QuantumClifford.

  The main problems were:

  - It chose + versus * using is_abelian(G). That is incorrect: Oscar has multiplicative Abelian groups, such as cyclic_group(4). Being Abelian does not imply additive Julia syntax. The fix
    dispatches on the element representation instead.

  - Actual FinGenAbGroupElem values were rejected in some constructor paths. The commit normalizes both group elements and coordinate tuples when constructing the code.
  - Infinite groups were accepted, which could make collect(G) nonterminating.
  - Elements from another, even isomorphic, group could be silently ignored through haskey checks. They now produce a clear ArgumentError.
  - Duplicate supports and empty supports had inconsistent behavior. Duplicates are now removed, and mathematically valid empty supports are allowed.
  - parity_matrix returned Matrix{Int}. QuantumClifford and QECCore expect a Boolean binary symplectic matrix in (X | Z) order.
  - The original tests did not parse and attempted to use a CSS-only MIP distance algorithm on non-CSS Mirror codes. They also represented a very large CI workload and treated reported
    distance bounds as exact values.

  The corrected implementation explicitly applies

  - symmetric: B g⁻¹,
  - asymmetric: g⁻¹ B,
  - both: A g,

  using the appropriate additive or multiplicative operations. Its tests use three small published codes plus focused group-action and input-validation cases.

  This commit establishes valid group actions and matrices. The subsequent commit, “Validate Mirror stabilizers and assign consistent phases,” separately handles non-Abelian commutation and
  the exclusion of -I.
```

Point 2 is a bit complicated. One part of it is just to ensure that if G is non-abelian, A and B still define valid checks. Or maybe you would prefer that we forbid non-abelian G? 

The other part of check 2 is really interesting though and I would have not noticed it at all -- it is amazing to me that codex noticed it. If we just take the boolean matrix you have generated and turn it into a tableau, if we are not careful, we might end up with a set of stabilizers that can generated -I, i.e. "negative identity", which leads to an empty code space. This really does not seem like it would be an issue unique to the Mirror codes -- is this something you have seen discussed before? It seems like a fix we should add throughout the codebase, not just for this code.

Point 3 is adding the missing docs.

Point 4 makes sure that your mirror codes are actually part of the test infrastructure (the basic tests that run for all codes).

Let me know what you think about these changes and how you would like to proceed.

PR #698, opened 2026-03-14
move BCH code to QECCore

Fe-r-oz:


Krastanov:
thanks! Closing in favor of #799 which is build on top of this one and where I instructed the agent to also take care of the various version bumps and dependencies.

PR #707, opened 2026-04-01
Revert "2D Tile codes, a slight generalization of 2D Surface codes"

Krastanov:
Reverts QuantumSavory/QuantumClifford.jl#660


Krastanov:
reverting #660 as it was merged without review and without having most of the tests run

we should also not let the downgrade check just bitrot, that makes the test unusable

Fe-r-oz:
> we should also not let the downgrade check just bitrot, that makes the test unusable

The downgrade test failed the version of QECCore had not released. I was going to do that in an upcoming PR with additional polish.

Fe-r-oz:
I think the deletion of 2dtile codes in unnecessary here -- I have checked that all these codes passed basic tests, as indicated by the bespokte test file and given that I have been very pertinent with the tests

Krastanov:
I am certain you have been diligent and that tests actually pass. But it is important for the automated checks to be in place, because otherwise they rot bit by bit and it becomes very difficult to make them work again -- not because there is some big bug, but because there are tens of little papercuts. It takes me a lot of time to fix these small papercuts and currently I am the only one that is dealing with these boring papercuts, which is why I am investing in so many little automations.

Contributions like yours are the important ones and the various automations should not impede them, so if some of these tests are flaky, we should fix them so that your work is more pleasant. But I also want my work to be pleasant, which is why I am insisting on the automations.

The buildkite tests being broken and not running is my fault, sorry about that.

We should figure out a way to set up the downgrade checks in a way that does not break in between PRs -- could you just redo your PR so that we can see exactly how they break.

You had added bespoke tests, yes, but the common-denominator tests were not running, which should be a red flag -- when an automated check is not executing the first thought should be "danger, ANY check not running might be hiding the fact that even more important checks are not running".

I know it is churn, but please just plan to resubmit the PR and do a proper review with passing CI. Otherwise every two months I need to spend a couple of days fixing small papercuts that have accumulated in the CI.

Krastanov:
in summary: this is not reverting the 2d tile codes because there is a problem with them, this is reverting it because (1) there is a problem with CI and (2) they were merged without review which let the CI problem get a bit worse instead of having the CI problem fixed first and then simply clicking the "update" button on the 2d tile codes PR.

It is quite possible that after the CI is fixed your PR can be merged with no changes and minimal review.

PR #708, opened 2026-04-01
2D Tile codes, a slight generalization of 2D Surface codes

Fe-r-oz:
Adds 2D Tile codes with CI tests added to the `test_ecc.jl`. Reference #660 and #707


Krastanov:
thanks for the resubmission, I know the chore side of software dev can be... well, a chore

Krastanov (changes requested):
There are some minor citation issues in the docs -- probably due to not having the .bib changes from the old PR. You probably already noticed these. (edit: yup you fixed it while I was writing this)

The bigger question is, how do we deal with the downgrade tests in a good way. For the last half a year these tests had started failing and only recently I fixed them, so now issues like this actually show up.

One option is to have always a sequence of two PRs, ready for merging quickly one after the other. This is not great because we will have broken tests in between and we might miss something and make the second PR actually difficult to merge. And a version with broken tests that are difficult to fix will already be out.

Another option is to restructure the tests and move much of the ECC testing to QECCore. Doing that little by little is probably the correct thing to do, but that is a lot of work. Maybe codex/claude can do it for us.

Anyway, I am leaning towards having a sequence of two PRs one after the other right now. It feels like the least bad option.

If we are going that way: The Project.toml of QECCore needs to have its version bumped here and the compat bounds in the Project.toml of QuantumClifford need to be bumped in the next PR. Let's have the next PR ready to go before we merge this one.

I have not reviewed the code itself yet, rather this is just a comment on workflow.

Fe-r-oz:
> If we are going that way: The Project.toml of QECCore needs to have its version bumped here and the compat bounds in the Project.toml of QuantumClifford need to be bumped in the next PR. Let's have the next PR ready to go before we merge this one.

Sounds good. bumped means that change the `QECCore` version to `1.4` in this PR, and then in another follow-up PR, change the project.toml of `QC` to `1.4` as well

Krastanov-agent:
Superseded by #796. The replacement preserves the original commits and authorship, starts from the complete #708 head, syncs current master, and adds the correctness, API, test-cost, and release-metadata fixes from review. It targets master directly and runs CI from a branch on QuantumSavory.

Krastanov:
Keeping open for now, as there is a bit of a discussion to do and the AI agent is overeager with what it does, making replacement PRs.

@Fe-r-oz , as usual, thank you for this submission, your prolific additions are very valuable. You probably have noticed I have not had much time for review. AI agents both help and hinder, but at least AI agents make it easy for me to start pipelining reviews, so I have been using them for help with the review work.

There are a few issues that the AI agent detected and tried to correct in #796 but I am not necessarily happy with what the agent did. You can conveniently see each commit that the agent did, separate, on a specific small issue, with a detailed message about why the change was made at the bottom of this page: https://github.com/QuantumSavory/QuantumClifford.jl/pull/796/commits

There is a boring commit about validating input values, but then a commit that seems to be fixing real bugs. Here are the comments from the AI agent:

```
1. [P1] Qubit pruning uses union instead of intersection.
     qubits = unique(vcat(Xrows..., Zrows...))
     (https://github.com/QuantumSavory/QuantumClifford.jl/blob/3bbfb38762551317b01674852b95fa49796fb20b/lib/QECCore/src/codes/quantum/tile2d.jl#L107-L133) retains qubits supported by either
     check family. The cited construction removes qubits unsupported by either family, so only the intersection should survive, followed by retruncating rows and removing newly empty checks.
     A valid pruning-sensitive example currently returns 46 columns, including six with no X support and six with no Z support; correct pruning yields a [[34,4]] code. Add a test asserting
     no returned column has zero support in either matrix. This matches the paper’s construction (https://arxiv.org/abs/2504.09171).

  2. [P1] Parity matrices have the wrong element type and corrupt Python decoder input.
     spzeros(Int, ...) (https://github.com/QuantumSavory/QuantumClifford.jl/blob/3bbfb38762551317b01674852b95fa49796fb20b/lib/QECCore/src/codes/quantum/tile2d.jl#L123-L131) contradicts the
     documented Boolean matrix contract. More seriously, existing Python BP decoders reinterpret Boolean matrices as bytes
     (https://github.com/QuantumSavory/QuantumClifford.jl/blob/3bbfb38762551317b01674852b95fa49796fb20b/ext/QuantumCliffordPyQDecodersExt/QuantumCliffordPyQDecodersExt.jl#L37-L54). A 6×16
     Int64 Tile2D matrix becomes 48×16 after that reinterpretation, silently changing the check count. Use spzeros(Bool, ...), assign true, and test both eltype and a decoder integration
     path.

  3. [P1] Invalid coordinates can construct anticommuting “CSS” checks.
     The constructor (https://github.com/QuantumSavory/QuantumClifford.jl/blob/3bbfb38762551317b01674852b95fa49796fb20b/lib/QECCore/src/codes/quantum/tile2d.jl#L43-L46) never verifies 0 ≤
     x,y < B. For example, Tile2D(2, [(0,0)], [(2,0)], 1, 1) is accepted but produces an odd X/Z overlap. Validate every coordinate and add tests for negative and B-valued coordinates.

  4. [P1] The documented public import does not expose Tile2D.
     The example uses using QuantumClifford.ECC, but ECC’s explicit export list
     (https://github.com/QuantumSavory/QuantumClifford.jl/blob/3bbfb38762551317b01674852b95fa49796fb20b/src/ecc/ECC.jl#L26-L52) omits Tile2D; a clean session gets UndefVarError. The doctest
     passes only because its harness globally imports QECCore. Either re-export Tile2D from ECC and test the documented imports, or document using QECCore consistently.

  5. [P1] Release integration remains incomplete.
     The QECCore changelog announces 0.1.4, but lib/QECCore/Project.toml remains 0.1.3
     (https://github.com/QuantumSavory/QuantumClifford.jl/blob/3bbfb38762551317b01674852b95fa49796fb20b/lib/QECCore/Project.toml#L1-L4). Root compatibility still permits 0.1.3, while the
     local [sources] path masks registry resolution in CI. Bump QECCore to 0.1.4 here, then raise QuantumClifford’s lower bound before its release, as already requested in the existing
     review (https://github.com/QuantumSavory/QuantumClifford.jl/pull/708#pullrequestreview-4046943719).

  6. [P2] Equal horizontal/vertical support counts are not a Tile-code requirement.
     length(horiz) == length(vert) rejects valid asymmetric supports. T2 derives Z support by reflection and swapping; it does not require the two X-support cardinalities to match. Remove
     this restriction or explicitly narrow the type’s documented scope.

  7. [P2] The test design makes required CI dramatically slower.
     The 288-qubit exact-distance MIP runs in both the unit test
     (https://github.com/QuantumSavory/QuantumClifford.jl/blob/3bbfb38762551317b01674852b95fa49796fb20b/lib/QECCore/test/codes/tile2d.jl#L30-L40) and doctest, while six 288–512-qubit
     instances enter every generic code suite. On comparable Ubuntu jobs:
      - Master (https://github.com/QuantumSavory/QuantumClifford.jl/actions/runs/32495560230/job/96812996126): QuantumClifford 36m26s, QECCore 8m07s.
      - PR (https://github.com/QuantumSavory/QuantumClifford.jl/actions/runs/32448275404/job/96671807385): QuantumClifford 83m38s, QECCore 58m56s.

     Remove the MIP solve from the doctest, retain the exact d=12 check in a slow/nightly job, and use one small Tile2D instance in generic suites. The 900-second limit is per optimization—
     up to eight X plus eight Z solves—not per complete test.

  Minor consistency fixes: both changelogs say 2DTile instead of Tile2D; Appendix B contains an identical duplicated fixture at lines 78–79; and the generic constructor arguments should be
  tuples rather than Vector{Any} values.
```

I am a bit less sure whether I like what the AI agent has done here in the new #796 replacement for this PR. Let me know what you think and whether you want to bother with the AI generated suggestions. Unlike with the other PRs where I used AI for review, here I have not vetted the work of the AI well.

Issue #480, opened 2025-02-16
subsystem quantum codes


Fe-r-oz:
**Is your feature request related to a problem? Please describe.**

Recently, a new quantum LDPC code, the Subsystem Hypergraph Product Simplex ([SHYPS](https://arxiv.org/pdf/2502.07150)) code, was introduced. It utilizes the Subsystem Hypergraph Product (`SHGP`) for code construction. ✨ This presents a fascinating use case of the `SHGP` framework (subsystem quantum codes) in building a quantum LDPC codes!

This [nice paper](https://arxiv.org/pdf/2002.06257) provides the code construction for the two families of the quantum subsystem codes:
```mermaid
graph TD
    A[Subsystem quantum codes] -->|Family 1| B[Bravyi-Bacon-Shor codes]
    A -->|Family 2| C[Subsystem hypergraph product codes]
```

**Describe the solution you’d like**

- [ ] Bravyi-Bacon-Shor code
- [ ] Subsystem Hypergraph Product (`SHGP`) code

**Additional context**

Kenneth and Isaac have done inspiring and exciting work on `SubsystemCodeTableau` and new group theory tools for QEC! It might also be probably related and helpful in the present context of subsystem quantum codes: [grouptableaux.jl](https://github.com/QuantumSavory/QuantumClifford.jl/blob/master/src/grouptableaux.jl) ✨ This topic appears to be closely related to the excellent work by Kenneth and Isaac on more general QECCs in #293!

the-punisher-29:
Thanks @Fe-r-oz!
A small update: I started laying the groundwork for these subsystem codes a few weeks back while i was at finishing stages(debug and optimization) of the ZSZ codes PR. The last ~20 days or more, I have been building out the implementations following to the library's conventions.Have also been reading through the BScThesisPhysics you shared(to ensure everything is covered).

about the framework — after looking at the structure closely, I opted not to treat SubsystemCodeTableau as a black box. We have exact, closed-form analytical descriptions for these codes, so generic tableau canonicalization would've been messier than it's worth.

What I ended up doing:

- BravyiBaconShor, SubsystemHypergraphProduct, and SubsystemHypergraphProductSimplex directly subtype AbstractCSSCode
- The structs natively hold a Tableau for gauge_generators and a Stabilizer for parity_checks, no delegation to SubsystemCodeTableau
[...]

PR #710, opened 2026-04-07
Feature: Added SHGP and SHYPS Subsystem Quantum Codes

the-punisher-29:
## Resolves Issue: #480

This PR adds support for Subsystem Quantum Codes in `QuantumClifford.jl` — specifically Bravyi-Bacon-Shor (BBS), Subsystem Hypergraph Product (SHGP), and Subsystem Hypergraph Product Simplex (SHYPS).

As discussed with @Krastanov and @Fe-r-oz  in #480, I've intentionally skipped the generalized `SubsystemCodeTableau` path for these families. They have exact closed-form geometry, so building them natively avoids unnecessary solver overhead and produces tighter internal structure.

### Architecture & Design Decisions

**1. New Classical Seed**

Added `Simplex(r)` to `lib/QECCore`, structured as the dual of the Hamming code.

**2. Direct Analytical Framework**

`BravyiBaconShor`, `SubsystemHypergraphProduct`, and `SubsystemHypergraphProductSimplex` all subtype `AbstractCSSCode` directly. Each holds `Tableau` objects for `gauge_generators` and `Stabilizer` objects for `parity_checks`. Subsystem geometry is computed via GF(2) matrix operations (`Nemo.kernel` for nullspaces); $k$ is derived from exact Galois Field matrix ranks.

**3. Subsystem Ecosystem Integration**

Added `gauge_generators(c)` and `code_g(c)` overrides to the `ECC.jl` layer for generic subsystem support. `_stabilizer_rank()` connects the rank bounds between stabilizers and gauges.

### Verification & Test Suite

Tests are in `test_ecc_subsystem.jl`:

- **Paper-verified inputs**: Reproduces the $[[21, 4, 3]]$ BBS and $[[49, 16, 3]]$ SHP configurations using matrix inputs from Li & Yoder.
- **Commutation invariants**: Loop checks verify that stabilizers are Abelian, stabilizers commute with all gauge generators, and gauge generators have non-Abelian intersections where expected.
- **QLDPC bound**: For SHYPS, checks that gauge weight stays strictly below $r+1$, confirming the $O(\log n)$ scaling.

@Fe-r-oz ,the `BScThesisPhysics` doc you recommended was genuinely useful, especially for the final round of debugging. Looking forward to your thoughts.

Note:I have removed the BBS Code from here,and have raise a separate PR for the same


lib/QECCore/src/codes/classical/simplex.jl
```diff
@@ -0,0 +1,91 @@
...
+    end
+    N = zeros(Int, nullity, n)
+    for (k, fc) in enumerate(free_cols)
+        N[k, fc] = 1
+        for (pr, pc) in enumerate(pivot_cols)
+            N[k, pc] = A[pr, fc]
+        end
+    end
+    return N
+end
```
Fe-r-oz: I had previously implemented the `gf2_nullspace` method as part of some GF(2) algebra tools that could be useful here. See:

https://github.com/QuantumSavory/QuantumClifford.jl/blob/444f341a50d2926b16b63d98586b8b06a7b6ac10/src/QuantumClifford.jl#L1287

test/test_ecc_subsystem.jl
```diff
@@ -0,0 +1,238 @@
+@testitem "Subsystem Codes: BBS, SHP, and SHYPS" tags=[:ecc] begin
```
Fe-r-oz, line 1: Following up on my previous comment:

There are currently no tests for` _gf2_nullspace`. That’s fine for now, as I had already implemented and tested this method earlier. In general, though, if you are adding a new GF(2) method, it is important to include tests to ensure correctness.

I had added the tests in https://github.com/QuantumSavory/QuantumClifford.jl/blob/444f341a50d2926b16b63d98586b8b06a7b6ac10/test/test_row_echelon_with_pivots.jl#L47

Fe-r-oz (commented):
Thanks, @the-punisher-29! I have not looked into the codes yet, but I had a few quick comments on the GF(2) `nullspace` method. I had implemented and tested this earlier, anticipating it could be useful down the line

lib/QECCore/src/codes/classical/simplex.jl
the-punisher-29: Thanks @Fe-r-oz! we can use the existing implementation .
But one problem: `simplex.jl` is inside `lib/QECCore`, which sits below `QuantumClifford.jl` in the dependency graph, so it can't reach up to import `QuantumClifford.gf2_nullspace`.
Moving `gf2_nullspace` down into `QECCore` ; probably `lib/QECCore/src/codes/util.jl` ,would fix this cleanly. `QECCore` gets the function it needs, `QuantumClifford` re-exports it as before, and the duplicate goes away.

Does that work? Happy to push the update.

test/test_ecc_subsystem.jl
the-punisher-29: Since your `gf2_nullspace` already has a test block in `test_row_echelon_with_pivots.jl`, the move I proposed in the previous comment handles this too. Pull it down into `lib/QECCore`, delete `_gf2_nullspace` from `simplex.jl`, and we're relying on one tested implementation instead of two.

Let me know if that works and I'll push the migration.

lib/QECCore/src/codes/classical/simplex.jl
```diff
@@ -0,0 +1,91 @@
+"""
+    $TYPEDEF
+
+The `[2ʳ - 1, r, 2ʳ⁻¹]` simplex code family, dual to the binary Hamming codes.
+
+`C(r)` is the dual of `Hamming(r)`. Its codewords are the rows of the Hamming
+parity check matrix, and every nonzero codeword has weight `2ʳ⁻¹`.
+
```
Fe-r-oz: Ah, since `C(r)` (if you mean parity check matrix of the Simplex code) is a *dual* meaning H*G^T ==0 of `Hamming(r)`, then there is a much simpler way to do this. `Nemo/Oscar` has `nullspace` method that can be utilized


Here is the code that takes in H (parity check matrix, in this case of Hamming code), and returns a dual of H:

```
function _dual(H)
    r, c = size(H)
    R, _ = residue_ring(ZZ, 2)
    H = matrix_space(R, r, c)(R.(H)) # TODO there must be a cleaner way to write this
    null = nullspace(H)[2]
    @assert all(iszero, H*null)
    return transpose(null)
end

```

src/ecc/codes/subsystem_hypergraph_product.jl
```diff
@@ -0,0 +1,135 @@
...
+
+### Fields
+    $TYPEDFIELDS
+"""
+struct SubsystemHypergraphProduct <: AbstractCSSCode
+    H1::Matrix{Int}
+    H2::Matrix{Int}
+    gauge_generators::Tableau
+    stabilizer::Stabilizer
+end
```
Fe-r-oz, line 24: Generally, it is recommended to submit one PR dedicated to a single, specific topic. Although it can be exciting to submit all the new codes at once, it is much easier for reviewers if each PR focuses on a single change/feature/fix. Smaller PRs are typically easier to review than a single large PR containing many different changes, even if they fall under the same broader category. This also makes it easier to test all edge cases and add comprehensive tests for the specific change.

the-punisher-29: okay,since that  issue i.e 480 had all these listed.i worked on a single branch then made a pr.
should i close this, rework and break the solution into 2 or 3 smaller parts and raise separate pr's?

Fe-r-oz: No need to close this PR -- you can submit the` BravyiBaconShor` in a subsequent PR (after this one) if you’d like, and keep this PR focused on `SHYPS` codes. Also, great work implementing all of these codes!

lib/QECCore/src/codes/classical/simplex.jl
Fe-r-oz: A much simpler approach is described [here](https://github.com/QuantumSavory/QuantumClifford.jl/pull/710#discussion_r3048173915)

test/test_ecc_subsystem.jl
Fe-r-oz: same as above

lib/QECCore/src/codes/classical/simplex.jl
the-punisher-29: Hmm, this approach looks mathematically cleaner! I might be missing something about how the dependencies work, though.
Since Nemo is marked as [weakdeps] inside lib/QECCore/Project.toml, wouldn't a direct Nemo.nullspace call inside lib/QECCore/src/codes/classical/simplex.jl just throw an UndefVarError?
My understanding is the entire Simplex struct would have to live inside the QECCoreNemoExt extension for that to compile. Which would mean Simplex is only available when the user explicitly does using Nemo , is that right?
Not sure how to test that cleanly locally.

src/ecc/codes/subsystem_hypergraph_product.jl
the-punisher-29: i have reorganized the files in this way ,updating this pr and will soon raise a separate for BBS codes

lib/QECCore/src/codes/classical/simplex.jl
the-punisher-29: Hey @Fe-r-oz ,a follow-up on where we were last.
Nemo is under [weakdeps] in lib/QECCore/Project.toml, so calling Nemo.nullspace directly in simplex.jl would pull Simplex into the QECCoreNemoExt extension. Users would need using Nemo just to use Simplex, which seems like more friction than we want.
My thought is moving your existing tested gf2_nullspace from src/QuantumClifford.jl into lib/QECCore/src/codes/util.jl instead ;doing this will give QECCore ,a native nullspace, QuantumClifford re-exports it, and we delete the duplicate _gf2_nullspace from simplex.jl. Cleaner all around.
Can push that if you're good with it.

Fe-r-oz: > My understanding is the entire Simplex struct would have to live inside the QECCoreNemoExt extension for that to compile. Which would mean Simplex is only available when the user explicitly does using Nemo , is that right?
Not sure how to test that cleanly locally.

Thanks for looking into this and for keeping users in mind

For testing extension-based methods, here is a concrete example from `QECCore`:  
https://github.com/QuantumSavory/QuantumClifford.jl/blob/master/lib/QECCore/test/codes/goppa.jl

I would suggest using Nemo’s implementation of `nullspace` rather than moving our internal GF(2) routines. Moving `gf2_nullspace` into `QECCore` would likely also require relocating `gf2_row_echelon_with_pivots!` along with the associated tests, which would probably become a separate PR.

More generally, using `Nemo`’s methods via the extension mechanism keeps `QECCore` lighter and avoids duplicating GF(2) boolean algebra functionality we do not need to reimplement The idea is to rely on functions from computer algebra systems like `OSCAR` wherever possible, rather than relying on homegrown implementations.

I think that defining the `dual` method in `QECCore` (as suggested in the [comment](https://github.com/QuantumSavory/QuantumClifford.jl/pull/710#discussion_r3048173915)) has benefits beyond just `Simplex`, since it can be reused more broadly by users. I have defined a `dual` method in a separate library, but this likely belongs in `QECCore`

lib/QECCore/src/codes/classical/simplex.jl
the-punisher-29: done

lib/QECCore/src/codes/classical/simplex.jl
the-punisher-29: Done! i observed the Goppa extension pattern and pushed changes on simplex.jl inline with it:
 
* `lib/QECCore/src/codes/classical/simplex.jl` → lightweight stub (like Goppa), no Nemo dependency
* `lib/QECCore/ext/QECCoreNemoExt/simplex.jl` → real `struct Simplex` + constructor using `_dual` with `Nemo.nullspace`
* `_gf2_nullspace` → deleted
The `_dual` method lives in the extension now. Reusable for other dual-code constructions later.

lib/QECCore/ext/QECCoreNemoExt/simplex.jl
```diff
@@ -0,0 +1,66 @@
...
+
+This is a general-purpose utility: for any binary matrix `H`, the dual code's
+generator matrix `G` satisfies `H * Gᵀ = 0` over GF(2).
+"""
+function _dual(H)
+    H_nemo = matrix(GF(2), H)
+    null = Nemo.nullspace(H_nemo)[2]
+    @assert all(iszero, H_nemo * null)
+    return Nemo.transpose(null)
+end
```
Fe-r-oz: It's better to change it to `dual` and export this method from` QECCore` rather than just keeping it as helper so it can be used outside of `QECCore` by other libraries

the-punisher-29: addressed

Fe-r-oz: Thanks! I did not got the notification for the update this, please feel free to ping me in the future

Fe-r-oz:
> @Fe-r-oz ,the BScThesisPhysics doc you recommended was genuinely useful, especially for the final round of debugging. Looking forward to your thoughts.

I am glad you found the document useful. I would recommend looking into more recent work by the author as well, for further context on related developments in quantum low-density parity check codes. I am happy to help further review this PR pretty soon, but please do not hesitate to request review from other members from the ECC Team if convenient

PR #719, opened 2026-04-29
Add benchmark suite for GPU-accelerated canonicalization

### LINKED ISSUE #553: canonicalization operations to KA.jl
Note: This was intentionally submitted as an issue in order to potentially re-visit it later should the (Julia)GPU landscape hopefully continue to evolve.

For the sake of my own sanity and hopefully getting something at least usable to be merged at some point before the heat death of the universe, I am writing this in hopes of getting some opinions/consensus on which is the lesser of all evils that we must regrettably consent to.
Certain procedures, such as canonicalisation in this case, are cooperative[1] and iterative[2] in such a manner that **ABSOLUTELY CANNOT** be cleanly implemented with minimal idiomatic code (if at all) should any sane portability, performance, (a)synchronicity, or safety guarantees be provided.

[1] Used here to indicate that correct operation requires some level of communication and/or synchronisation between all threads in the executing context.
[2] Used here to indicate that a sequential series of similar procedures (which are permitted to be parallel) must be performed until a dynamically determined exit condition is satisfied, though this predicate can be deduced to be eventually satisfied within some known number of steps.

<br />

Below is a list of possibilities that I have explored, alongside the issues or limitations that they each present. It has been lovingly titled as `Here be dragons in canonicalisation: Whether 'tis nobler to block or to deadlock` and is open for further suggestions that I may have potentially missed.

1. Grid/NDrange native barriers
This is the most natural and idiomatic solution to this problem. The resulting implementation would be completely asynchronous, involve only a single kernel launch, remain compact and succinct for those who care about that aspect, as well as employ nothing more than the hardware's own native synchronisation primitives.
However, whilst AMD and NVIDIA both support the block-grid programming model and the relevant synchronisation calls (at least on any relatively recent hardware) Apple and Intel stick to the block-ndrange model and the necessary barriers are completely absent. `KernelAbstractions` does not support this for obvious reasons. 

2. Grid/NDrange mutex barriers
This solution is similar to the above with the exception that synchronisation is now handled via a mutex-based implementation rather than native platform primitives. It would definitely increase the implementation footprint and its performance would absolutely suffer to an extent, possibly quite severely, but it would still retain complete asynchronicity and utilise only a single kernel launch.
However, this implementation is not safe in the slightest without some runtime knowledge of the hardware's concurrency limits. Launching more threads than the hardware can handle simultaneously will result in a deadlock situation, since the running threads will be sitting in a busy loop awaiting for signals from other threads that have not been (and **CANNOT** be) scheduled for execution. In any case, `KernelAbstractions` does not support querying the hardware limits required to implement this and any "conservative underestimate" will either severely under-utilise the hardware resources (bad) or will fail to actually be an underestimate and deadlock the device (absolutely unacceptable). Providing a few light-weight vendor-specific extensions that query this information and then pass it along should be possible in principle, assuming that the backends themselves permit those queries, but I am not certain whether it is something that `QuantumClifford` wishes to support.

3. Dynamic/Nested Parallelism
This is the second-most natural and idiomatic solution this problem. Like the first solution, it would be completely asynchronous, compact and succinct, and employing nothing more than the hardware's own native features. However, there would now be a parent kernel that manages the procedure by monitoring the exit condition and launching a child kernel at each iteration. I would not envision that the performance penalty for the successive kernel launches will be a major limiting factor since they should only constitute a minuscule fraction of the overall execution time.
However, whilst this is part of the OpenCL specification and thus should also be present in every conformant vendor's own native platform (unleses you're AMD??), it does not appear to be supported by `KernelAbstractions` regardless of whether the pertinent backend actually allows it.

4. Host-side asynchronous kernel launch
This solution would have the responsible host thread/process/task would fork/spawn a child whose sole purpose is to sit in a loop and launch successive iterations of the canonicalisation kernel until the exit condition is satisfied, thereby maintaining (at least some level of) asynchronicity since the potentially blocking kernel launches are happening elsewhere. The safety implication of this solution differ from the second proposal in that it does not hang but can potentially cause a segmentation fault if the object is de-allocated or changes its address in memory. For safety and correctness, this solution would require forbidding enqueuing any further work on the same object before a synchronisation point is established, since it could possibly be scheduled at any point during the canonicalisation loop, as well as requiring both a host-side and a device-side synchronisation barrier.

5. Host-side synchronous kernel launch.
This "solution" should be taken to be synonymous with "I give up". It does not even bother with any asynchronicity concerns, it directly goes into the kernel launch loop and will block until it is finished.

satzgits:
Resolves #553 by adding canonicalization benchmarks to the existing KernelAbstractions benchmark framework. This addresses the benchmark gap identified during PR #576 review.

### Changes
- Add `benchmark_KA_canonicalize.jl` with CPU vs GPU comparison for:
  - `canonicalize!` (Gaussian elimination)
  - `canonicalize_rref!` (RREF canonicalization)
- Add `n_qubits` configuration array for canonicalization benchmarks (64, 128, 256, 512, 1024 qubits)
- Integrate into existing `benchmark_platform.jl` runner

### Output
- Runtime plots (ms vs qubit count) for host and device
- Speedup ratio plots (host/device) across batch sizes
- Results saved to `canonicalize/` subdirectory with SVG format

### Usage
Run the existing platform benchmark scripts:
```bash
julia benchmark/KernelAbstractions/benchmark_platform_CUDA.jl
```

This extends the existing `benchmark_KA_mul_leftright` suite to cover canonicalization, which was identified as a missing component during the KA.jl canonicalization implementation review.

cc @Hamiltonian-Action @Krastanov

### DIFF
--- a/benchmark/KernelAbstractions/implementation/benchmark_KA_canonicalize.jl
+++ b/benchmark/KernelAbstractions/implementation/benchmark_KA_canonicalize.jl
@@ -1,0 +1,110 @@
+# This must be done explicitly as they are not exported.
+using QuantumClifford: canonicalize!, canonicalize_rref!, Stabilizer, MixedDestabilizer, random_stabilizer
+
+@inline function host_f_canon!(s)
+    canonicalize!(s)
+end
+
+@inline function host_f_rref!(s)
+    canonicalize_rref!(s)
+end
+
+@inline function device_f_canon!(
+    s, synchronize;
+    block_size::Val{block_SZ} = Val(default_block_size),
+    batch_size::Val{batch_SZ} = Val(default_batch_size)
+    ) where {block_SZ, batch_SZ}
+
+    canonicalize!(
+        s;
+        block_size = block_size, batch_size = batch_size
+        )
+    synchronize()
+end
+
+@inline function device_f_rref!(
+    s, synchronize;
+    block_size::Val{block_SZ} = Val(default_block_size),
+    batch_size::Val{batch_SZ} = Val(default_batch_size)
+    ) where {block_SZ, batch_SZ}
+
+    canonicalize_rref!(
+        s;
+        block_size = block_size, batch_size = batch_size
+        )
+    synchronize()
+end
+
+@inline function benchmark_KA_canonicalize(
+    AT, synchronize, path;
+    canon_type::Symbol = :canonicalize
+    )
+
+    host_time = zeros(Float64, length(n_qubits))
+    device_time = zeros(Float64, length(batch_sizes), length(n_qubits))
+
+    cache = AllocCache()
+    for (i, n) in enumerate(n_qubits)
+        @cached cache begin
+            h_s = random_stabilizer(n)
+            d_s = Stabilizer(
+                AT(u32(h_s.phases)),
+                h_s.nqubits,
+                AT(reinterpret(UInt32, h_s.xzs))
+                )
+            synchronize()
+
+            if canon_type == :canonicalize
+                host_f = host_f_canon!
+                device_f = device_f_canon!
+            else
+                host_f = host_f_rref!
+                device_f = device_f_rref!
+            end
+
+            # Trigger compilation before benchmarking.
+            host_f(copy(h_s))
+            for (j, size) in enumerate(batch_sizes)
+                device_f(copy(d_s), synchronize; batch_size = Val(size))
+            end
+
+            host_time[i] = @belapsed $host_f(copy($h_s)) evals = evals samples = samples seconds = seconds
+            for (j, size) in enumerate(batch_sizes)
+                device_time[j, i] =
+                    @belapsed $device_f(copy($d_s), $synchronize; batch_size = Val($size)) evals = evals samples = samples seconds = seconds
+            end
+        end
+    end
+    unsafe_free!(cache)
+
+    device_cat = [device_time[i, :] for i = 1 : length(batch_sizes)]
+    title =
+    "Performance uplift - $(canon_type)\n\
+    Host threads = " * string(Threads.nthreads()) * " / " *
+    string(Sys.CPU_THREADS) * ", Device block size = $default_block_size"
+    xlabel = "Number of qubits"
+    label = hcat(("Device - batch size = " .* string.(batch_sizes))..., "Host")
+
+    plot(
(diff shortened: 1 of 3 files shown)


benchmark/KernelAbstractions/implementation/benchmark_KA_canonicalize.jl
```diff
@@ -0,0 +1,110 @@
...
+    canon_type::Symbol = :canonicalize
+    )
+
+    host_time = zeros(Float64, length(n_qubits))
+    device_time = zeros(Float64, length(batch_sizes), length(n_qubits))
+
+    cache = AllocCache()
+    for (i, n) in enumerate(n_qubits)
+        @cached cache begin
+            h_s = random_stabilizer(n)
```
Hamiltonian-Action, line 49: This is logically correct but computationally unnecessary. Stabilisers are expensive to compute, since the rows need to commute on top of being linearly independent, but the canonicalisation achieves maximum runtime for any full-rank tableau.
This arises (exercise to the implementer) almost-surely for any uniformly random square tableau in the limit of large nqubits -- actually, the convergence is so fast that the probability of **NOT** being full-rank with merely 32 qubits is already less than one in a billion.

benchmark/KernelAbstractions/implementation/benchmark_KA_canonicalize.jl
```diff
@@ -0,0 +1,110 @@
...
+            d_s = Stabilizer(
+                AT(u32(h_s.phases)),
+                h_s.nqubits,
+                AT(reinterpret(UInt32, h_s.xzs))
+                )
+            synchronize()
+
+            if canon_type == :canonicalize
```
Hamiltonian-Action, line 57: It would be more straightforward to skip the conditional branching and just pass the functions themselves as arguments to `benchmark_KA_canonicalize` from `benchmark_KA_canonicalize_all`.

Hamiltonian-Action (commented):
Thank you for submitting this PR.
Unfortunately, as it stands, this is dependent on the canonicalisation routines which have not yet been finalised themselves, ergo it is likely to be hanging around for a while.
Though it does indeed get the job done, I'm not certain whether it would be wise to have benchmark-specific qubit counts or to move towards a timing-based system whence the routines would abort should the benchmark run exceed a certain time interval.

PR #720, opened 2026-05-01
Add two-qubit depolarization noise

Mingyuan1231:
disscussed in zoom meeting


src/noise.jl
```diff
@@ -28,6 +28,17 @@ struct UnbiasedUncorrelatedNoise{T} <: AbstractNoise
...
+"""
+Two-qubit depolarization noise model with total probability of error `p`.
+
+The two-qubit identity `II` occurs with probability `1-p`, while each of the
+15 non-identity two-qubit Pauli errors occurs with probability `p/15`.
+"""
+struct DepolarizationNoise{T} <: AbstractNoise
+    p::T
+end
+DepolarizationNoise(p::Integer) = DepolarizationNoise(float(p))
```
Krastanov, line 52: As we were discussing on the call, how do you feel about the following change.

In order for us to not hardcode the number of qubits, let's keep the name DepolarizationNoise, but use the parameterization:

E(rho) = (1-p)rho + 1/4^n Sum (P rho P) where the sum is over all paulis including identity and then in the docstring we explain the relationship between this and the "error probability" because here p includes 1/4^n chance for no error.

The docstring will need to be updated

src/noise.jl
```diff
@@ -72,6 +83,45 @@ function applynoise!(s::AbstractStabilizer,noise::PauliNoise,i::Int)
...
+    end
+    return s
+end
+
+const _TWO_QUBIT_DEPOLARIZATION_PAULIS = (
+    (4,1), (4,2), (4,3),
+    (1,4), (1,1), (1,2), (1,3),
+    (2,4), (2,1), (2,2), (2,3),
+    (3,4), (3,1), (3,2), (3,3),
+)
```
Krastanov: because we are switching to parameterization by depolarization probability we can skip this list

src/noise.jl
```diff
@@ -72,6 +83,45 @@ function applynoise!(s::AbstractStabilizer,noise::PauliNoise,i::Int)
...
+    r = rand()
+    if r >= noise.p
+        return s
+    end
+
+    e1, e2 = rand(_TWO_QUBIT_DEPOLARIZATION_PAULIS)
+    _apply_pauli_idx!(s, i, e1)
+    _apply_pauli_idx!(s, j, e2)
+    return s
+end
```
Krastanov: simplify this to do something like

- pick a random number between 1 and 16
- do division with remainder to get the paulis for the first and second qubit
- apply those paulis

src/noise.jl
```diff
@@ -72,6 +83,45 @@ function applynoise!(s::AbstractStabilizer,noise::PauliNoise,i::Int)
...
+
+const _TWO_QUBIT_DEPOLARIZATION_PAULIS = (
+    (4,1), (4,2), (4,3),
+    (1,4), (1,1), (1,2), (1,3),
+    (2,4), (2,1), (2,2), (2,3),
+    (3,4), (3,1), (3,2), (3,3),
+)
+
+function applynoise!(s::AbstractStabilizer, noise::DepolarizationNoise, i::Int, j::Int)
```
Krastanov: and also add a method with just an `i` argument so it works for single qubits

src/noise.jl
Krastanov, line 52: if you want to have two constructors, one of them will need to use keyword arguments instead of normal arguments:

- if by default you have the error probability, you can do `DepolarizationNoise(;depolarization_probability=...) = ..`
- if by default you have the deporalization probability, you can do `DepolarizationNoise(;error_probability=...) = ...`

src/noise.jl
```diff
@@ -178,6 +228,24 @@ function applynoise_branches(s::AbstractStabilizer,noise::UnbiasedUncorrelatedNo
     results
 end
 
+function applynoise_branches(s::AbstractStabilizer, noise::DepolarizationNoise, indices::Base.AbstractVecOrTuple; max_order=1)
+    length(indices) == 2 || throw(ArgumentError("DepolarizationNoise acts on exactly two qubits"))
```
Krastanov: implement for single qubits by calling into and returning what is already implemented for the simpler type of noise

update the error message to mention you can do only single or two qubits

src/noise.jl
```diff
@@ -72,6 +83,45 @@ function applynoise!(s::AbstractStabilizer,noise::PauliNoise,i::Int)
...
+    end
+
+    e1, e2 = rand(_TWO_QUBIT_DEPOLARIZATION_PAULIS)
+    _apply_pauli_idx!(s, i, e1)
+    _apply_pauli_idx!(s, j, e2)
+    return s
+end
+
+function applynoise!(s::AbstractStabilizer, noise::DepolarizationNoise, indices::Base.AbstractVecOrTuple)
+    length(indices) == 2 || throw(ArgumentError("DepolarizationNoise acts on exactly two qubits"))
```
Krastanov: update the error message to include single qubits as well

Krastanov (changes requested):
looks great, requesting some small changes

src/noise.jl
Mingyuan1231: removed

src/noise.jl
Mingyuan1231: fixed

src/noise.jl
Mingyuan1231: fixed

src/noise.jl
Mingyuan1231: added

src/noise.jl
Mingyuan1231: fixed

src/noise.jl
Mingyuan1231: fixed

Krastanov-agent:
Superseded by #724.

Krastanov:
closed in favor of #724 which has a small addition to the tests

PR #734, opened 2026-06-03
Add OpenQASM 3 import support

### LINKED ISSUE #727: Read OpenQASM 3 files into QuantumClifford circuits

---

QuantumClifford circuits are represented as vectors of symbolic operations such
as `sHadamard`, `sCNOT`, `sMZ`, and `NoiseOp`. OpenQASM 3 is a common circuit
exchange format, and Quasar.jl provides a Julia parser for OpenQASM 3. Adding an
import path from OpenQASM 3 into QuantumClifford would make it easier to bring in
Clifford circuits generated by other tools and simulate them with
`mctrajectories`, `pftrajectories`, and related backends.

Relevant links:

- Symbolic Clifford operations: https://github.com/QuantumSavory/QuantumClifford.jl/blob/main/src/symbolic_cliffords.jl
- Circuit simulation docs: https://qc.quantumsavory.org/stable/API/#Circuit-simulation
- Quasar.jl: https://github.com/kshyatt-aws/Quasar.jl
- OpenQASM 3 specification: https://openqasm.com/versions/3.0/language/

A user should be able to read a QASM 3 file or string into a QuantumClifford
circuit:

```julia
using QuantumClifford

circuit = read_qasm3("bell.qasm")
```

For this input:

```qasm
OPENQASM 3.0;

qubit[2] q;
bit[2] c;

h q[0];
cx q[0], q[1];
c[0] = measure q[0];
c[1] = measure q[1];
```

the imported circuit should be equivalent to:

```julia
[
    sHadamard(1),
    sCNOT(1, 2),
    sMZ(1, 1),
    sMZ(2, 2),
]
```

Note the conversion from OpenQASM's zero-based indices to QuantumClifford's
one-based qubit and bit indices.

## Suggested implementation

Prefer implementing this as an optional extension or weak dependency if Quasar.jl
is used, so users who do not need QASM import do not pay for the dependency.

The first implementation should focus on Clifford-compatible statements:

- `qubit` and `bit` declarations
- single-qubit Clifford gates: `id`, `x`, `y`, `z`, `h`, `s`, `sdg`
- two-qubit Clifford gates: `cx`, `cz`, `swap`
- computational-basis measurement into classical bits
- `reset`, if the mapping to existing reset-measurement operations is clear

Unsupported operations should fail with an informative error that includes the
operation name and source location if Quasar exposes one. Non-Clifford gates
such as `t` and parameterized rotations can be left unsupported initially, or
accepted only behind an explicit option if there is an existing compatible
QuantumClifford operation.

## In scope

- `read_qasm3(path)` and/or `parse_qasm3(str)` returning a `Vector` of
  QuantumClifford operations.
- Register bookkeeping for qubit and classical bit declarations.
- Correct zero-based to one-based index conversion.
- Tests with small `.qasm` fixtures.
- A documented table mapping supported OpenQASM gate names to QuantumClifford
  operation constructors.
- Clear errors for unsupported OpenQASM features.

## Out of scope

- Complete OpenQASM 3 language coverage.
- Timing, calibration, pulse-level constructs, and hardware annotations.
- Automatic decomposition of arbitrary custom gates.
- Non-Clifford circuit simulation support beyond what QuantumClifford already
  provides.

## Acceptance criteria

- A simple Bell circuit in OpenQASM 3 imports to the expected QuantumClifford
  operations.
- Imported Clifford circuits can run through `pftrajectories`.
- Tests cover qubit arrays, bit arrays, measurement assignment, and at least the
  listed Clifford gates.
- Unsupported gates produce actionable error messages.
- The docs include a complete QASM input example and the resulting
  QuantumClifford circuit.

sagnikpal2004:
Adds OpenQASM 3 import support to QuantumClifford using Quasar.jl as an optional extension.

This PR introduces:
- `parse_qasm3(qasm::AbstractString)`
- `read_qasm3(path::AbstractString)`

The implementation is provided through a package extension and weak dependency on Quasar.jl, so users who do not need OpenQASM support do not incur the dependency.

Rather than relying on Quasar's built-in `QasmProgramVisitor`, this PR uses its own visitor `Qasm2CliffordVisitor` to perform a traversal of the parsed OpenQASM AST and constructs QuantumClifford symbolic operations, allowing measurement destinations and reset operations to be mapped correctly into QuantumClifford's circuit representation.

## Supported OpenQASM features
- `qubit` and `bit` declarations
- single-qubit Clifford gates: `id`, `x`, `y`, `z`, `h`, `s`, `sdg`
- two-qubit Clifford gates: `cx`, `cz`, `swap`
- computational-basis measurement into classical bits
- `reset` operation

## Notes
- OpenQASM's zero-based indexing is automatically converted to QuantumClifford's one-based indexing.
- Unsupported gates and unsupported OpenQASM features produce informative errors.
- Measurement destinations are preserved and mapped to the corresponding classical bit indices used by QuantumClifford measurement operations.

## Checklist
- [ ] All of the automated tests on github pass
- [X] We recently started enforcing formatting checks. If formatting issues are reported in the new code I have written, I will correct them.

# CHANGELOG
- Added `parse_qasm3` and `read_qasm3` for converting OpenQASM 3 files containing supported Clifford operations into QuantumClifford symbolic operations.

Closes #727 for unitaryHack


Krastanov-agent:
Review notes after checking PR 734 against Quasar.jl v0.0.2 documentation and source.

I focused on whether this uses Quasar idiomatically, whether the implementation is as simple as possible, user-facing errors, weak-dependency UX, coverage, unused code, and type annotations.

## Findings

1. **Valid/common QASM currently fails before conversion.**

   `bit[1] c = "0";` is rejected because initialized declarations go through the same `:classical_assignment` branch that only accepts `measure` on the RHS:

   - `ext/QuantumCliffordQASMExt/CliffordQasmVisitor.jl:115`
   - `ext/QuantumCliffordQASMExt/CliffordQasmVisitor.jl:140`

   I verified this throws:

   ```text
   QasmVisitorError: unsupported assignment operation: QasmExpression :classical_assignment ...
   ```

   This matters because initialized bit registers are normal OpenQASM and appear in Quasar’s own examples, e.g. `bit[2] b = "00";`.

   Also, `include "stdgates.inc";` currently fails at parse time with:

   ```text
   SystemError: opening file "stdgates.inc": No such file or directory
   ```

   That happens before the converter gets control, via `parse_qasm(qasm)` in `ext/QuantumCliffordQASMExt/QuantumCliffordQASMExt.jl:6`. Many QASM files contain this include even when the downstream simulator wants to treat standard gates natively.

2. **The visitor is more manual than Quasar’s intended extension style.**

   Quasar documents a two-step workflow: parse to AST, run `QasmProgramVisitor`, then consume `visitor.instructions`, which are `CircuitInstruction` named tuples. It also documents custom gate dialects through `BuiltinGateDefinition` and `Quasar.builtin_gates[]`.

   This PR instead implements a partial AST visitor at `ext/QuantumCliffordQASMExt/CliffordQasmVisitor.jl:92`, duplicating Quasar internals for declarations, gate calls, literals, ranges, measurement, and reset. That loses generic Quasar behavior for aliases, custom gate definitions, modifiers, loops, conditionals, function calls, and Quasar’s gate argument validation.

   Caveat: Quasar’s current generic visitor does not preserve classical measurement destinations in its `CircuitInstruction`s, so a completely direct “run `QasmProgramVisitor` then translate instructions” path is not fully drop-in for `sMZ(q, c)`. But this still argues for a much narrower custom layer: reuse Quasar’s gate dialect machinery and traversal where possible, and only special-case the measurement-destination information that Quasar does not expose cleanly today.

3. **`WeakDepHelpers` should be wired for the exported QASM stubs.**

   Without `import Quasar`, calling `parse_qasm3("OPENQASM 3.0;")` currently gives only:

   ```text
   MethodError: no method matching parse_qasm3(::String)
   The function `parse_qasm3` exists, but no method is defined for this combination of argument types.
   ```

   The empty public stubs at `src/qasm_ext.jl:56` and `src/qasm_ext.jl:63` should be registered with `WeakDepHelpers.@declare_method_is_in_extension` against `QuantumClifford.WEAKDEP_METHOD_ERROR_HINTS`, for both `parse_qasm3` and `read_qasm3`, with a hint along the lines of “load `Quasar` first with `import Quasar`”. The package already has `WEAKDEP_METHOD_ERROR_HINTS` and uses this pattern elsewhere.

4. **User-facing error messages should be more specific and actionable.**

   Some examples I observed with `Quasar` loaded:

   ```text
   QasmVisitorError: gate t not defined!
   QasmVisitorError: gate rx not defined!
   QasmVisitorError: unsupported assignment operation: QasmExpression :classical_assignment ...
   QasmVisitorError: measurement source and destination sizes must match.
   QasmVisitorError: Invalid qubit index '3' in 'q'.
   ```

   The main throw sites are:

   - `ext/QuantumCliffordQASMExt/CliffordQasmVisitor.jl:129`
   - `ext/QuantumCliffordQASMExt/CliffordQasmVisitor.jl:140`
   - `ext/QuantumCliffordQASMExt/CliffordQasmVisitor.jl:143`
   - `ext/QuantumCliffordQASMExt/CliffordQasmVisitor.jl:167`

   Suggested direction:

   - distinguish “unsupported non-Clifford gate” from “undefined gate”;
   - for parameterized gates, mention that this importer currently supports only the Clifford subset (`id`, `x`, `y`, `z`, `h`, `s`, `sdg`, `cx`, `cz`, `swap`, measurement, reset);
   - for measurement size mismatch, include source and destination sizes and names if available;
   - avoid dumping full `QasmExpression` trees in normal user errors unless in a debug path;
   - consider wrapping/rethrowing parse-time include errors for `stdgates.inc` with a QuantumClifford-specific explanation.

5. **Coverage is good for happy paths, but not complete.**

   The targeted QASM test item passes, but coverage on the extension visitor was not complete. I ran the QASM tests with Julia coverage and saw:

   ```text
   ext/QuantumCliffordQASMExt/QuantumCliffordQASMExt.jl: 2/2 executable lines covered
   ext/QuantumCliffordQASMExt/CliffordQasmVisitor.jl: 100/111 executable lines covered
   ```

   Uncovered branches include:

   - negative classical bit indices in `_evaluate_bits`;
   - classical declarations with initializers;
   - `:empty` handling;
   - the generic unsupported-expression fallback;
   - some error branches.

   The tests currently assert only `@test_throws QasmVisitorError` for failures, so they do not protect error-message quality. I would add tests for:

   - weakdep hint when `Quasar` has not been imported;
   - initialized `bit` declarations;
   - `include "stdgates.inc";` behavior;
   - negative bit indices if intended to be supported;
   - exact or substring error messages for unsupported gates, parameterized gates, unsupported assignments, and measurement size mismatch.

6. **Some abstractions and annotations should be simplified.**

   `CliffordGateDefinition.qubit_targets` is not literally unused: it is consumed indirectly by Quasar’s `process_gate_targets` at `ext/QuantumCliffordQASMExt/CliffordQasmVisitor.jl:131`. But the custom `CliffordGateDefinition` abstraction is probably not necessary if the importer uses Quasar’s `BuiltinGateDefinition`/`CircuitInstruction` dialect path.

   `classical_defs` is stored but not meaningfully used beyond declaration bookkeeping. If the importer only needs bit mappings, that can probably be simpler.

   `_evaluate_bits(::Val{:hw_bit}, ...)` appears dead/mistyped in this context. Quasar has `:hw_qubit`; I did not find `:hw_bit` in Quasar v0.0.2. If there is no parser path that produces `:hw_bit`, this method should be removed.

   Many local type annotations add noise without improving Julia inference or readability, for example:

   - `qubit_name::String = ...`
   - `qubit_size::Int = ...`
   - `gate_name = name(program_expr)::String`
   - `start::Int, step::Int, stop::Int = ...`
   - return annotations on many small helpers.

   I would keep field types where they define storage, but remove most local assertion-style annotations unless they are deliberately validating external data or fixing a known inference issue.

7. **Docs and formatting need cleanup.**

   The doc table currently says:

   - `id -> sId` at `src/qasm_ext.jl:42`, but the implementation uses `sId1`;
   - `cz -> sCZ` at `src/qasm_ext.jl:50`, but the implementation uses `sCPHASE`.

   The doctest example imports only `Quasar`, not `QuantumClifford`; that may pass under package doctest setup, but it is not the best user-facing example. I would show:

   ```julia
   using QuantumClifford
   import Quasar
   ```

   Finally, `git diff --check upstream/master...upstream/pr-734` fails on trailing whitespace and missing final newlines in new files, including `src/qasm_ext.jl`, `test/test_qasm.jl`, and the new extension/QASM fixture files.

## Verification I ran

Targeted QASM test item:

```text
Test Summary: | Pass  Total  Time
Package       |   15     15  3.0s
```

Coverage run of the same targeted QASM test item also passed:

```text
Test Summary: | Pass  Total  Time
Package       |   15     15  6.1s
```

I did not run the full package suite.

## Quasar references checked

- Quasar README/docs describe `parse_qasm`, `QasmProgramVisitor`, and consuming `visitor.instructions`.
- Quasar docs for custom gate sets describe providing a function returning `Dict{String, BuiltinGateDefinition}` and assigning it to `Quasar.builtin_gates[]`.
- Quasar source has `QasmProgramVisitor`, `CircuitInstruction`, `BuiltinGateDefinition`, `process_gate_targets`, `splat_gate_targets`, and generic handling for much more AST surface than this PR currently supports.

--
Reviewed by OpenAI Codex, based on GPT-5, running in the Codex coding-agent environment. Review date: 2026-06-12.

Krastanov:
The comment above was generated by an LLM after a bit of back and forth prompting from me and after I requested that it reports on a few specific things I saw.

Please specifically look into:

1. A potential parsing error. The rest of your code/comments seems to imply that this is not meant to be supported (and it is potentially difficult to support in QuantumClifford). If it is not supported on purpose, document that a bit more clearly.

2. This one is pretty important. The level of complication in the visitor is not warranted -- you already have a parsing library that should handle most of it.

3. Just so that the error messages are clearer when Quasar.jl is not installed.

4. that one is not too important -- the error messages are already reasonable enough

5. Just the part about uncovered branches. The rest of the comment about "error-message quality" does not seem too pertinent.

6. That one is fairly important, please look into it.

7. not too important

Krastanov (changes requested):
Thanks for starting this. Marking as "request changes" given the issues discussed above.

Krastanov:
A changelog is also needed.

The downgrade check also does not fail on master, so the issue must be coming from here. Can probably be fixed by changing our lowest DataStructures compat bound to 0.18.20, the oldest version supported by Quasar.jl.

Krastanov:
convert back to not-draft when you would like a second review

sagnikpal2004:
Hi @Krastanov, thank you for the valuable feedback. I have addressed each of the itemized issues:
1. added "unsupported operations" section in the docstring
2. The reason for implementing a visitor from scratch instead of using Quasar's visitor are follows:
2.1. Quasar does not have support for `measure` and `reset` operations. For `measure`, it actually stores a value of 0 in the bit, no matter what, and continues the program, instead of emitting an instruction that stores the qubit measurement value into that bit. `reset` is currently implemented as a no-op, and a warning is printed to remind that.
2.2. Quasar does not perform bookkeeping for classical variables. Quasar actually precomputes all classical operations, including flow control, and produces quantum-only instructions. This is a great feature, however, for any branches that are dependent on a measurement, this does not work properly, as measurements always produce a zero, no matter what, and the program uses that zero to do flow control.

    Using Quasar's inbuilt QasmProgramVisitor and then converting those instructions to QuantumClifford.jl, as well as add extra layers of classical bit bookkeeping, and measurement locations, and implementing reset are indeed possible, and my first attempts were indeed adding layers on top of the existing Quasar visitor,
but **since these would require a visiting of the AST (for a second time) anyway**, I found creating a QuantumClifford specific visitor, that directly produces a QC circuit is cleaner.
    Also, even if we are indeed able to implement layers on top of the Quasar's functionality to fix its shortcomings, we would still have incorrectness, if a qasm program with flow control was inputted. Since, the measurement results are always zero, and all while loops and if statements would be flattened out.

    See https://github.com/kshyatt-aws/Quasar.jl/issues/31 and https://github.com/kshyatt-aws/Quasar.jl/issues/31
I think, if some of the things in Quasar are fixed, then it can definitely be used in its intended way. Basically, I could upstream some of the code from QuantumClifford into Quasar. But for now, we can still use the AST parser.

3. Registered WeakDepHelpers! (used the default message)
```
ERROR: MethodError: no method matching parse_qasm3(::String)
The function `parse_qasm3` exists, but no method is defined for this combination of argument types.
HINT: `parse_qasm3` depends on the package(s) `Quasar` but you have not installed or imported them yet. Immediately after an `import Quasar`, `parse_qasm3` will be available.
```

4. improved the error messages for:
4.1. measurement register size mismatch: `QasmVisitorError: measurement size mismatch: cannot measure 3 qubit(s) from q1 into 1 bit(s) from c2.`
4.2. preventing the dumping of full AST on error messages.

5. Added test coverage for:
5.1. negative indices (removed some redundant code)
5.2. classical declaration with initialization failure
5.3. (removed unreachable :empty AST node)
5.4. more unsupported operations

6. Removed the following
6.2. `classical_defs` is not required since we are not storing the values of the classical variables. It was a semantic copy of `quantum_defs` from Quasar
6.3. deleted `_evaluate_bits(::Val{:hw_bit}, ...)` as `hw_bit` does not exist. It was a semantic copy of `hw_qubit` from Quasar
6.4. Removed type annotations where not necessary. Quasar seems to have a lot of unnecessary type annotations, and I ended up copying the same style over, in hopes to upstream this functionality.

    Did not remove
    6.1. I had tried using the `CircuitInstruction` dialect path, but because the inbuilt visitor was throwing away so much of the information from the AST, it was just easier to simply create our own `CliffordGateDefinition` and have the `sXX` as our circuit instructions directly

7. Fixed the docs for `sId1` and `CPHASE`

Please let me know your thoughts on if the workaround seems like a good way to solve this issue. Please let me know if there are any other changes I should make.

sagnikpal2004:
### Changelog
- Added `parse_qasm3` and `read_qasm3` for converting OpenQASM 3 files containing supported Clifford operations into QuantumClifford symbolic operations using `Quasar`
- Updated compatibility bounds for `DataStructures` and `PrecompileTools` to support OpenQASM import weak dependency: `Quasar`.

My understanding is that you will add these into the changelog yourself?

## Downgrade check
I lifted the compat bounds of `DataStructures` to 0.18.20 and that of `PrecompileTools` to 1.2.1

sagnikpal2004:
I am unsure why the `[QuantumClifford Tests (with local QECCore) - GPU - ROCm](https://buildkite.com/quantumsavory/quantumclifford-dot-jl/builds/3534/list?jid=019ed648-69fd-47cb-a566-27b2c25dbee1)` test is running infinitely long. There isn't any debug output either.

I tried restarting the buildkite, but no luck.
It started happening after lifting compat bounds, so that might be the culprit

PR #740, opened 2026-06-09
Add reinterpret for PauliOperators and Tableaux/Stabilizers/etc

arnavk23:
**Supersedes #620** (closed due to the git merge issues)

**If this is your first submission to this organization and you are not a developer known in the Julia ecosystem, do not use LLMs -- we need to trust you first before we trust the LLM under your control.**

Before considering your pull request ready for review and merging, make sure that all of the following are completed (**please keep the clecklist as part of your PR**):

If you are submitting for a bug bounty:
- [x] I have read and followed the [AI usage and disclosure policy](https://github.com/QuantumSavory/.github/blob/main/BUG_BOUNTIES.md)

### DIFF
--- a/src/reinterpret.jl
+++ b/src/reinterpret.jl
@@ -1,0 +1,113 @@
+function Base.reinterpret(::Type{U}, p::PauliOperator) where {U <: Unsigned}
+	old = eltype(p.xz)
+	total_bytes = length(p.xz) * sizeof(old)
+	q, r = divrem(total_bytes, sizeof(U))
+	if r != 0 || isodd(q)
+		throw(ArgumentError(THROW_REINTERPRET_SIZE_MISMATCH(:pauli, old, U, total_bytes, q, r)))
+	end
+	new_xz = reinterpret(U, p.xz)
+	return PauliOperator(p.phase, p.nqubits, new_xz)
+end
+
+function Base.reinterpret(::Type{U}, t::Tableau) where {U <: Unsigned}
+	parent_xzs = t.xzs isa LinearAlgebra.Adjoint ? parent(t.xzs) : t.xzs isa LinearAlgebra.Transpose ? parent(t.xzs) : t.xzs
+	old = eltype(parent_xzs)
+	chunk_axis = size(parent_xzs, 1)
+	rows_axis = size(parent_xzs, 2)
+	old_chunks_expected = 2 * QuantumClifford._nchunks(t.nqubits, old)
+	if chunk_axis != old_chunks_expected
+		total_bytes = length(parent_xzs) * sizeof(old)
+		q, r = divrem(total_bytes, sizeof(U))
+		throw(ArgumentError(THROW_REINTERPRET_SIZE_MISMATCH(:tableau, old, U, total_bytes, q, r)))
+	end
+	new_chunk_axis = 2 * QuantumClifford._nchunks(t.nqubits, U)
+	if iszero(new_chunk_axis) || isodd(new_chunk_axis)
+		total_bytes = length(parent_xzs) * sizeof(old)
+		q, r = divrem(total_bytes, sizeof(U))
+		throw(ArgumentError(THROW_REINTERPRET_SIZE_MISMATCH(:tableau, old, U, total_bytes, q, r)))
+	end
+	total_bytes = length(parent_xzs) * sizeof(old)
+	q, r = divrem(total_bytes, sizeof(U))
+	if r != 0 || q != length(parent_xzs) * sizeof(old) ÷ sizeof(U)
+		throw(ArgumentError(THROW_REINTERPRET_SIZE_MISMATCH(:tableau, old, U, total_bytes, q, r)))
+	end
+
+	reint = reinterpret(U, vec(parent_xzs))
+	new_mat = reshape(reint, new_chunk_axis, rows_axis)
+	new_mat = copy(new_mat) # materialize to plain Array if needed
+
+	if t.xzs isa LinearAlgebra.Adjoint
+		return Tableau(t.phases, t.nqubits, new_mat')
+	elseif t.xzs isa LinearAlgebra.Transpose
+		return Tableau(t.phases, t.nqubits, transpose(new_mat))
+	else
+		return Tableau(t.phases, t.nqubits, new_mat)
+	end
+end
+
+function Base.reinterpret(::Type{U}, s::Stabilizer) where {U <: Unsigned}
+	try
+		return Stabilizer(reinterpret(U, tab(s)))
+	catch e
+		if isa(e, ArgumentError) && occursin("Unable to reinterpret tableau storage", sprint(showerror, e))
+			# Re-throw with stabilizer context but preserve the original error message
+			err_msg = replace(sprint(showerror, e), "tableau" => "stabilizer")
+			throw(ArgumentError(err_msg))
+		end
+		rethrow(e)
+	end
+end
+
+function Base.reinterpret(::Type{U}, s::Destabilizer) where {U <: Unsigned}
+	try
+		return Destabilizer(reinterpret(U, tab(s)))
+	catch e
+		if isa(e, ArgumentError) && occursin("Unable to reinterpret tableau storage", sprint(showerror, e))
+			# Re-throw with destabilizer context but preserve the original error message
+			err_msg = replace(sprint(showerror, e), "tableau" => "destabilizer")
+			throw(ArgumentError(err_msg))
+		end
+		rethrow(e)
+	end
+end
+
+function Base.reinterpret(::Type{U}, ms::MixedStabilizer) where {U <: Unsigned}
+	try
+		return MixedStabilizer(reinterpret(U, tab(ms)), rank(ms))
+	catch e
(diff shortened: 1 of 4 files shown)


src/reinterpret.jl
```diff
@@ -0,0 +1,113 @@
...
+	try
+		return Stabilizer(reinterpret(U, tab(s)))
+	catch e
+		if isa(e, ArgumentError) && occursin("Unable to reinterpret tableau storage", sprint(showerror, e))
+			# Re-throw with stabilizer context but preserve the original error message
+			err_msg = replace(sprint(showerror, e), "tableau" => "stabilizer")
+			throw(ArgumentError(err_msg))
+		end
+		rethrow(e)
+	end
```
Krastanov: These try/catch statements and error rewriting steps are pretty wild and overly engineered. Going back to the simpler previous version would be much cleaner.

Krastanov: THROW_REINTERPRET_SIZE_MISMATCH also will become much simpler

Krastanov (changes requested):
I vaguely remember the previous implementation in #620 looking much simpler than this -- am I misremembering, is this the same as #620?

There are a few other issues (upcoming post from a bot will list them) that I believe we had previously addressed in 620.

Krastanov-agent:
I reviewed this after reading through the comments on the earlier version in #620. The main open issues I see are below, with concrete examples.

1. `src/reinterpret.jl:17` and `src/reinterpret.jl:23`: tableau chunk counts are doubled. `_nchunks(q, T)` already includes both X and Z halves; `zero(Tableau, r, q)` uses it directly.

```julia
using QuantumClifford

t = zero(Tableau, 3, 64)

# Existing invariant from QuantumClifford:
@assert size(t.xzs, 1) == QuantumClifford._nchunks(nqubits(t), eltype(t.xzs))

# PR 740 expects this instead, so normal tableaus fail:
@assert size(t.xzs, 1) != 2 * QuantumClifford._nchunks(nqubits(t), eltype(t.xzs))

reinterpret(UInt8, t)  # should work, but currently throws
```

This also breaks `Stabilizer`, `Destabilizer`, mixed stabilizers, and `PauliFrame`, since they all go through `tab(...)`.

2. `test/test_reinterpret.jl:122`, `:141`, `:162`, `:184`, `:206`, `:229`, `:253`: positive-path tests accept exceptions as success, masking broken behavior.

```julia
t = zero(Tableau, 3, 64)

try
    reinterpret(UInt8, t)
    @test true
catch e
    # This makes a broken valid reinterpret pass the test.
    reinterpret_error_matches(e, "Unable to reinterpret tableau storage")
end
```

These should instead assert success directly:

```julia
t2 = reinterpret(UInt8, t)
t3 = reinterpret(eltype(t.xzs), t2)
@test t == t3
```

Only invalid alignment/size cases should use `@test_throws`.

3. `src/reinterpret.jl:13`: fast-column layout is handled by stripping `Adjoint` and then treating the parent’s first axis as the chunk axis. For `fastcolumn`, the parent axes are transposed.

```julia
using LinearAlgebra
using QuantumClifford

t = fastcolumn(zero(Tableau, 3, 7))

@assert t.xzs isa Adjoint
@show size(t.xzs)          # logical tableau storage axes
@show size(parent(t.xzs))  # parent axes are reversed

reinterpret(UInt8, t)      # should preserve fast-column behavior, but line 13 reads parent axes incorrectly
```

This is one of the cases explicitly requested in the #620 discussion: the tests should prove both `fastrow` and `fastcolumn` work.

4. `src/reinterpret.jl:37` and `src/reinterpret.jl:103`: the implementation copies/collects storage, which violates `reinterpret`’s expected shared-storage behavior.

```julia
t = zero(Tableau, 2, 64)
t2 = reinterpret(UInt8, t)

t2.xzs[1, 1] = 0xff

# For reinterpret-like behavior, mutating t2 should affect t's backing storage.
# The copy at src/reinterpret.jl:37 prevents that.
@test !all(iszero, t.xzs)
```

One of the main points from #620 was that this should be allocation-free and should reuse the same backing memory where possible.

5. `src/reinterpret.jl:104`: `PauliFrame` reconstructs with the original frame type after changing the tableau storage type. That can fail conversion back to the old type.

```julia
pf = PauliFrame(3, 4, 2)

new_frame = reinterpret(UInt8, pf.frame)

# PR code effectively does this:
typeof(pf.frame)(tab(new_frame))

# But typeof(pf.frame) still encodes the old tableau/xzs storage type.
# The result should use new_frame directly, not force the old frame type.
```

The test currently hides this by accepting conversion errors:

```julia
reinterpret_error_matches(e, (
    "Unable to reinterpret pauliframe storage",
    "Unable to reinterpret tableau storage",
    "Cannot `convert`",
))
```

I was not able to run local Julia probes to completion because the temporary checkout environment was not instantiated and `Pkg.instantiate()` stalled during artifact installation. GitHub CI reports the main unit/docs jobs passing, but these are static behavioral issues in the new code and tests.

-- Reviewed with OpenAI Codex CLI, GPT-5-based coding agent.

src/reinterpret.jl
```diff
@@ -0,0 +1,113 @@
...
+	total_bytes = length(parent_xzs) * sizeof(old)
+	q, r = divrem(total_bytes, sizeof(U))
+	if r != 0 || q != length(parent_xzs) * sizeof(old) ÷ sizeof(U)
+		throw(ArgumentError(THROW_REINTERPRET_SIZE_MISMATCH(:tableau, old, U, total_bytes, q, r)))
+	end
+
+	reint = reinterpret(U, vec(parent_xzs))
+	new_mat = reshape(reint, new_chunk_axis, rows_axis)
+	new_mat = copy(new_mat) # materialize to plain Array if needed
```
Hamiltonian-Action: This absolutely breaks the standard (if there is an official one) for how reinterpret should behave. The underlying data for the input and output structures are no longer the same, operations performed on one of them would not reflect on the other.
There is a utility function in the `Adapt.jl` extension that does something similar, perhaps changing the nomenclature a bit and then employing (a modified version of) that method would be a cleaner solution?
I, alongside a non-insignificant section of the ecosystem, do concur that wrappers need an overhaul in order to handle such situations. However, until such a time as the core language team addresses that issue, we must content with whatever is presently available.

Krastanov:
@arnavk23 , it seems this PR is very different from the #620 that HA and I had reviewed some time ago. If that is indeed the case and we are not misremembering, please close this PR and consider reopening the older one or resubmitting the older one from a newly named branch. The branch of the old one seems to be deleted.

Krastanov:
@arnavk23 , is it possible to explicitly bring back the previous version that was already reviewed. Otherwise we have to review it again, which is a lot of work and I probably will struggle to find the motivation to repeat work that was already done.

arnavk23:
> @arnavk23 , is it possible to explicitly bring back the previous version that was already reviewed. Otherwise we have to review it again, which is a lot of work and I probably will struggle to find the motivation to repeat work that was already done.

I have done just that, Stefan but wanted to go through all the comments once more on this code, specially the llm reply to see if the reviewed version has any faults before I asked you or HA for review.

Krastanov:
I meant doing that in some machine-verifiable form, e.g. seeing that this has the same hash or same branch or reviving the previous PR. Otherwise we still need to put in a lot of work to do a full review because there is no proof that this is actually exactly the same code as the one we had review already.

arnavk23:
> I meant doing that in some machine-verifiable form, e.g. seeing that this has the same hash or same branch or reviving the previous PR. Otherwise we still need to put in a lot of work to do a full review because there is no proof that this is actually exactly the same code as the one we had review already.

That can be seen in pr #696 , the pr is on a different issue but it contains all the information of the reinterpret branch before commits focused on the issue. Can be seen at https://github.com/QuantumSavory/QuantumClifford.jl/pull/696/changes/652d50a58fc91a480c1fe3ce5e93fa6aa196bcba

Hamiltonian-Action:
@Krastanov, just as a curiosity. Is there any reason that `QuantumClifford` would even like to have a `reinterpret` call?

From my understanding, the underlying implementation details should be transparent to any project or user the employs the library. In so far as they are concerned, we could be storing states on actual quantum hardware and they would be none the wiser.
If someone is concerned about performance differences between the various underlying unsigned types, I believe that there are other mechanisms available to produce that effect starting at the constructor level. There is also the `Adapt.jl` utility that I mentioned earlier, which does suitable zero-padding in order to always produce a copy of the desired typing -- the garbage collector will come and reap the old object if it is no longer in use, hence memory should not be too much of a concern except for some highly specific edge cases.

Krastanov:
@Hamiltonian-Action , I agree with your logic, this is not a super important feature. It is a well defined feature that is not unreasonable to exist, but it would not see wide use -- as such, if it gets implemented and reviewed, great, we can merge it, but probably it would not be a high priority to review for me.

arnavk23:
@Krastanov this is safe to review and merge. All the files (except `test_reinterpret.jl`) are the same to the previous pr on the topic and I have also mentioned your particular commit which I referenced. And the file `test_reinterpret.jl` was changed to follow `reinterpret.jl` properly and for the tests to pass on the current setup.

PR #742, opened 2026-06-10
Skiptree implementation based on Swaroop et al.

riteshbhirud:
# Universal qLDPC adapter construction (Swaroop, Jochym-O'Connor, Yoder 2026)

Adds a new `QuantumClifford.ECC.Adapters` submodule implementing the universal-adapter
construction of [Swaroop, Jochym-O'Connor, Yoder, *PRX Quantum* **7**, 010324 (2026)](https://arxiv.org/abs/2410.03628).
Given two CSS qLDPC code blocks (or two logicals on the same block), the construction
attaches an auxiliary graph plus bridge X-checks and produces a merged CSS code in
which the joint logical Pauli Z̄₁ ⊗ Z̄₂ is a stabilizer product, i.e. directly
measurable in one shot of syndrome extraction.

This is intended to be the first of two PRs.

## LLM Disclaimer
An LLM was used to assist with writing and refining comments and docstrings, drafting this PR description, and generating portions of the test suite based on detailed specifications fed by me. 

## Public API

Five functions and five types, exported from `QuantumClifford.ECC`:

| | |
|---|---|
| `build_adapter(::CodePair; kwargs...)` | top-level dispatcher (inter- vs intra-code) |
| `build_adapter_intercode` / `build_adapter_intracode` | the two explicit entry points |
| `deform_code(::CSS, support; kwargs...)` | single-logical deformation (paper Fig. 3 / Eq. 9) |
| `skiptree(::SimpleGraph; root=1)` | Algorithm 2 (Appendix E, Theorem 7) |
| `CodePair`, `AuxiliaryGraph`, `SkipTreeOutput`, `Adapter`, `AdapterMergedCode` | construction inputs and outputs |

`AdapterMergedCode` is a thin `AbstractCSSCode` wrapper around the result so the
merged code participates in the standard `QECCore` dispatch (`code_n`,
`parity_matrix_x`, `distance`, decoder harnesses, and so on) while still exposing
the originating `Adapter` for downstream consumers that need the auxiliary graphs
or SkipTree outputs.

## What's in here

The work splits across four source files under `src/ecc/adapters/`.

**`skiptree.jl`** implements Algorithm 2 / Appendix E, targeting the full-rank
repetition check `H_R(n)`. Given a connected graph, it returns `(T, P, :H_R)`
with `T · G · P ≡ H_R(n) (mod 2)`, where `T` is `(3, 2)`-sparse. The
implementation is a translation of the paper authors' reference
`skip_tree_algorithm.py`, with sorted neighbour iteration so Julia gives
deterministic output. We don't aim for byte-identical reproduction of the
Zenodo `T_X` matrices, since different MST and child orders produce different
but equally valid `T`s, but the algebraic guarantee is verified, including
against three independent `G_mat_*` graphs distributed with the Zenodo data.

**`aux_graph.jl`** handles initial-graph construction (G_0 from §II.C: one
vertex per support qubit, one edge per consecutive vertex pair in
`sort(supp(s) ∩ logical_support)` for each X-stabilizer), plus the cellulation
pass `cellulate_long_cycles!` that splits long basis cycles by midpoint chord
to satisfy desideratum 3a. Includes `relative_expansion` (exact brute force
over `2^n`, capped at `n ≤ 24`) and `verify_desideratum_4`, plus an
`F = port_function_as_matrix` helper for the vertex-Z block.

**`merge.jl`** is the Figure 7 / Eq. 37 assembly. Inter-code and intra-code
variants, plus `deform_code` for the single-logical case. The intra-code
path is the trickier one: when the two logical supports differ in size, the
adapter width is `w = min(w_l, w_r)` and we trim the SkipTree outputs
accordingly (`T_{l,r}` to `(w-1, m)` and `P_{l,r}` to `(*, w)`). The "extra"
aux1 vertices keep their F and G^T entries (needed for vertex-Z commutation)
but get a zero A-block, and the bridge only touches the first `w` label
positions so commutation still holds. Also, in the intra-code case a single
X-stabilizer can touch both logicals, so M_l and M_r live in the *same*
code-X-stab row block rather than two separate blocks. Both of these are
documented in the function docstring.

**`Adapters.jl`** is the small module file that glues things together and
re-exports the public surface.

One design choice worth flagging: `cellulate_long_cycles!` takes an optional
`chords::Vector{Tuple{Int,Int}}` kwarg. Without it, we use a midpoint
heuristic that splits the longest current basis cycle by a chord between
`cycle[1]` and `cycle[n÷2 + 1]`. This works for paper Examples B and C, but
for the paper's BB / Z_3 single-logical deformation the midpoint algorithm
picks chord `(4, 9)` where the paper's NetworkX-based pipeline picks
`(1, 11)`. Both are valid 17-edge cellulations, but the two choices produce
single-logical deformed codes of different distance: our default gives
`[[115, 5, ≤11]]` and the paper's gives `[[115, 5, 12]]`. The `chords`
kwarg is the escape hatch for byte-identical paper reproduction. The
joint-measurement case via `build_adapter` is unaffected: the bridge X-checks
empirically absorb the extra weight-`(d-1)` X-logical that the chord choice
introduces, so the merged code's distance comes out at 12 either way. This
is verified in the test suite by independently confirming that three
weight-11 X-logical supports (found out-of-band by HiGHS MIP) are genuine
non-trivial X-logicals of the standalone deformed code, then checking the
bridge syndrome on each.

## Paper reproduction

End-to-end against the authors' Zenodo data (`zenodo_17527545`, optional;
the relevant testsets `@test_skip` when not present, so CI without the data
still runs the non-Zenodo testsets):

| What | Setup | Result |
|---|---|---|
| Example B (inter-code) | BB[[98,6,12]] × LP[[200,20,10]] | merged code on n=355; Hx and Hz row-space ranks equal Zenodo's |
| Example C (intra-code) | BB[[98,6,12]] / Z₁ ⊗ Z₃ | [[150, 5, 12]]; n, Hx rows, Hz rows, ranks, and weight bounds all match paper |
| Table II aux sizes | Z₁ → 14 v / 23 e, Z₂ → 14 v / 20 e, Z₃ → 12 v / 17 e | byte-identical |
| Table II `deform_code` | [[121,5,12]], [[220,19,10]], [[115,5,12]] | match (Z₃ needs explicit `chords=[(1,11)]` per the note above) |
| §VII.A relative expansion | β = 5/6, 1/2, 3/4 | exact rationals |
| §VII.A cellulation edge counts (auto-tune defaults) | 21 → 23 (BB/Z₁), 20 → 20 (LP/Z₂, weight 7 covers everything), 16 → 17 (BB/Z₃) | match |

Distances for the merged codes (Example B at d=10, Example C at d=12) were
verified out-of-band with HiGHS MIP. The full proofs take long enough that
running them in CI isn't reasonable. The test file does run a smaller MIP
sanity testset that proves `d = 3` on Steane7 and Surface(3,3) via the same
MIP formulation, which catches any regression in the MIP plumbing itself.

## Beyond the paper

Smoke-tested on the [[144, 12, 12]] gross bivariate-bicycle code from
[Cross, He, Rall, Yoder (arXiv:2407.18393)](https://arxiv.org/abs/2407.18393)
(`l=12, m=6, A = x³+y+y², B = y³+x+x²`). Intra-code joint measurement of
two `logz_ops` representatives builds a structurally valid merged code on
230 qubits with adapter width 16; CSS commutation holds, ranks line up,
`code_k` drops from 12 to 11 as expected, and the weight bounds satisfy
paper Theorem 5 / Table III. Distance is not asserted (a full MIP cert at
`d=12` on this scale is hours, not seconds, and is out-of-band for the same
reason as above).

Also exercised on Surface(3,3) × Surface(3,3) as a small sanity case,
producing [[33, 1, 3]] with width-3 adapter.

## Future Work

- Hardware-aware placement, subdivision, or connectivity constraints, coming
  in a separate Williamson-Yoder PR.

## References

- E. Swaroop, T. Jochym-O'Connor, T. J. Yoder, *Universal adapters between
  qLDPC codes via code surgery*, PRX Quantum **7**, 010324 (2026);
  [arXiv:2410.03628](https://arxiv.org/abs/2410.03628).
- A. Cross, Z. He, P. Rall, T. J. Yoder, *Improved QLDPC surgery: logical
  measurements and bridging codes*,
  [arXiv:2407.18393](https://arxiv.org/abs/2407.18393).
- Zenodo data: [10.5281/zenodo.17527545](https://doi.org/10.5281/zenodo.17527545).

### DIFF
--- a/docs/src/adapter_tutorial.md
+++ b/docs/src/adapter_tutorial.md
@@ -72,6 +72,8 @@ measurement result.
 
 ## Step 3: algebraic verification of the recipe
 
+This step is simply a verification for correctness, not necessary in actual use.
+
 This step is simply a verification for correctness — not necessary in actual
 use. Each recipe row is a single ``Z``-type stabilizer of the merged code;
 their Pauli product equals the joint logical
--- a/docs/src/adapter_tutorial.md
+++ b/docs/src/adapter_tutorial.md
@@ -1,0 +1,298 @@
+# [Universal qLDPC adapters: joint logical measurement](@id adapter_tutorial)
+
+```@meta
+DocTestSetup = quote
+    using QuantumClifford
+    using QuantumClifford.ECC
+    using QuantumClifford.ECC.Adapters
+end
+```
+
+This page walks through the `Adapters` submodule on a concrete example: given
+two CSS codeblocks and a specific logical operator between them, build the
+universal adapter of [swaroop2026universal](@cite), read off the list of
+stabilizer measurements that implements the joint logical operator, and
+compare against the naive direct-measurement protocol in simulation.
+
+The construction realises a **joint Pauli measurement** ``\bar Z_l \otimes \bar Z_r``
+between a logical of one codeblock and a logical of another, using only
+weight-bounded ``X``- and ``Z``-type stabilizer measurements on a merged code.
+This is one half of a code-surgery protocol — the other half (the *split*
+step that restores two independent codeblocks afterwards) is not yet
+implemented and is called out at the end.
+
+## Step 1: pick the two codes and the logicals to fuse
+
+We use two independent copies of the ``3 \times 3`` planar surface code
+(`Surface(3, 3)` from `QECCore`) — `code_n = 13`, `code_k = 1`,
+`distance = 3` — and target the ``\bar Z`` logical on each block.
+
+```@example adapter
+using QuantumClifford
+using QuantumClifford: stab_to_gf2
+using QuantumClifford.ECC: CSS, Surface, parity_matrix_x, parity_matrix_z,
+                            logz_ops, logx_ops, code_n, code_k, parity_checks,
+                            naive_encoding_circuit
+using QuantumClifford.ECC.Adapters: CodePair, build_adapter, joint_logical_recipe
+
+as_css(c) = CSS(Matrix{Bool}(parity_matrix_x(c)),
+                Matrix{Bool}(parity_matrix_z(c)))
+
+codeA = Surface(3, 3)
+codeB = Surface(3, 3)
+nA = code_n(codeA); kA = code_k(codeA)
+nB = code_n(codeB); kB = code_k(codeB)
+
+# Z-logical support on a single Surface(3,3) block
+lz = stab_to_gf2(logz_ops(codeA))
+z = sort(findall(!iszero, lz[1, nA+1:2nA]))
+```
+
+`z` is the list of data qubits that ``\bar Z`` acts on within one block. We
+use the same support on both blocks since they are copies of the same code.
+
+## Step 2: build the adapter and read the measurement recipe
+
+`build_adapter` builds the two auxiliary graphs, cellulates them, runs the
+SkipTree algorithm on each, and assembles the merged CSS code that joins
+the two blocks.
+
+```@example adapter
+adapter = build_adapter(CodePair(as_css(codeA), as_css(codeB), z, z))
+
+(code_n(adapter), code_k(adapter))
+```
+
+The merged code has `code_n(adapter)` data qubits and `code_k(adapter) = 1`
+surviving logical qubit (the two original logicals have fused, as one expects
+from a surgery merge).
+
+`joint_logical_recipe` returns the row indices of the merged ``Z`` parity-check
+matrix whose XOR over GF(2) equals ``\bar Z_l \otimes \bar Z_r`` in the
+merged-qubit layout:
+
+```@example adapter
+recipe = joint_logical_recipe(adapter)
+```
+
+This is the *list of measurements* that implements the joint logical operator:
+after standard syndrome extraction on the merged code, XOR the measurement
+outcomes at these indices to recover the joint
+``\bar Z_l \otimes \bar Z_r`` measurement result.
+
+## Step 3: algebraic verification of the recipe
+
+Each recipe row is a single ``Z``-type stabilizer of the merged code. We can
+form the Pauli product directly and check that it equals the joint logical
+``\bar Z_l \otimes \bar Z_r`` embedded in the merged qubit layout (the merged
+column order is ``[\text{code}_1 \mid G_1 \mid A \mid G_2 \mid \text{code}_2]``,
+with ``A`` the adapter block and ``G_1, G_2`` the auxiliary-graph edge
+qubits).
+
+```@example adapter
+function row_to_zpauli(row, n)
+    p = zero(PauliOperator, n)
+    for i in 1:n
+        row[i] != 0 && (p[i] = (false, true))
+    end
+    p
+end
+
+n_total = code_n(adapter)
+Hz = parity_matrix_z(adapter)
+recipe_paulis = [row_to_zpauli(Hz[r, :], n_total) for r in recipe]
+
+prod(recipe_paulis)
+```
+
+The product is supported only on data qubits 3, 6, 9 (the ``\bar Z`` support
+of block 1) and 23, 26, 29 (the ``\bar Z`` support of block 2 — offset by
+``n_A + n_\text{anc} = 13 + 7 = 20``). All auxiliary qubits are trivial. This
+is exactly ``\bar Z_l \otimes \bar Z_r``.
+
+The cancellation on the auxiliary qubits is by design: the ``V_l`` row block
+contributes ``Z`` on every column of the adapter sub-block ``A`` (because the
+SkipTree permutation ``P_l`` places exactly one ``1`` per ``A`` column), and
+the ``V_r`` row block contributes the same ``Z`` on ``A``. Since
+``Z^2 = I`` over ``\mathbb F_2``, the auxiliary support cancels and only the
+joint logical survives on the data qubits.
+
+## Step 4: prepare encoded logical states in simulation
+
+We follow the standard `QuantumClifford` pattern (also used in the ECC
+syndrome tests): a random ``k``-qubit logical input is buffered with
+``n-k`` ``|0\rangle`` ancilla qubits and passed through
+`naive_encoding_circuit`. This produces a fully-encoded `MixedDestabilizer`
+state of the surface code.
+
+```@example adapter
+using Random; Random.seed!(2026)
+
+physical_state_A = one(Stabilizer, nA - kA) ⊗ random_stabilizer(kA)
+physical_state_B = one(Stabilizer, nB - kB) ⊗ random_stabilizer(kB)
+
+encoded_A = MixedDestabilizer(physical_state_A)
+mctrajectory!(encoded_A, naive_encoding_circuit(codeA))
+
+encoded_B = MixedDestabilizer(physical_state_B)
+mctrajectory!(encoded_B, naive_encoding_circuit(codeB))
+
+full_state = encoded_A ⊗ encoded_B
+nqubits(full_state)
+```
+
+`full_state` is the joint encoded state on the ``n_A + n_B = 26`` data qubits
+of the two surface code blocks.
+
+The logical Pauli we want to measure on this joint state is the tensor
+product of the two ``\bar Z`` operators. The single-block ``\bar Z`` is the
+first row of `logz_ops(codeA)` (and the same for `codeB`):
+
+```@example adapter
+logicalA = logz_ops(codeA)[1]
+logicalB = logz_ops(codeB)[1]
+full_logical = logicalA ⊗ logicalB
+```
+
+## Step 5: naive measurement of the joint logical
+
+The *naive* protocol is to construct ``\bar Z_l \otimes \bar Z_r`` as a single
+weight-6 Pauli and measure it directly with [`PauliMeasurement`](@ref) on a
+[`Register`](@ref) wrapping `full_state`. We harvest one classical bit per
+trajectory.
+
+```@example adapter
+nshots = 1000
+
+naive_bits = falses(nshots)
+for shot in 1:nshots
+    reg = Register(copy(full_state), 1)
+    mctrajectory!(reg, [PauliMeasurement(full_logical, 1)])
+    naive_bits[shot] = bitview(reg)[1]
+end
+
+(count(naive_bits), nshots)
+```
+
+The pair `(count(naive_bits), nshots)` reports how many shots returned
+``-1`` (i.e. measured `true`). For the seed-pinned random logical input above,
+the empirical distribution should be near ``50/50``.
+
+## Step 6: adapter-mediated measurement of the same logical
+
+The adapter implements the same joint measurement using only ``X``- and
+``Z``-type stabilizers of the merged code. The protocol on each shot is:
+
+1. Insert ``n_\text{anc}`` ancilla qubits in ``|+\rangle`` between the two
+   encoded blocks, matching the merged-code column layout.
+2. Measure every ``X``-type merged stabilizer (these are deterministic on the
+   chosen ancilla state and provide the standard CSS syndrome bits).
+3. Measure every ``Z``-type merged stabilizer.
+4. XOR the outcomes at the indices in `joint_logical_recipe(adapter)` to
+   recover the joint logical bit.
+
+Each *individual* merged stabilizer has weight at most ``4`` for this
+adapter (the original Surface(3,3) checks are weight 3 or 4, and the
+new ``V`` rows and bridge ``X``-rows the adapter introduces are weight
+bounded by the auxiliary-graph degree); the joint logical itself has
+weight 6, so even at distance 3 the adapter trades one weight-6
+measurement for a sequence of low-weight ones. The gap widens with
+code distance — joint logicals scale as ``O(d)`` while adapter
+stabilizers stay ``O(1)`` — and that is the whole point of the
+construction: a hardware that can execute only weight-bounded
+measurements (as on a qLDPC architecture) can still perform an
+arbitrary joint logical Pauli measurement.
+
+```@example adapter
+function row_to_xpauli(row, n)
+    p = zero(PauliOperator, n)
+    for i in 1:n
+        row[i] != 0 && (p[i] = (true, false))
+    end
+    p
+end
+
+Hx = parity_matrix_x(adapter)
+n_xstabs = size(Hx, 1)
+n_zstabs = size(Hz, 1)
+
+adapter_stabilizers = vcat(
+    [PauliMeasurement(row_to_xpauli(Hx[r, :], n_total), r)
+        for r in 1:n_xstabs],
+    [PauliMeasurement(row_to_zpauli(Hz[r, :], n_total), n_xstabs + r)
+        for r in 1:n_zstabs],
+)
+
+# Ancilla qubits live BETWEEN block A and block B in the merged column layout.
+n_anc = n_total - nA - nB
+plus_anc = MixedDestabilizer(Stabilizer([single_x(n_anc, i) for i in 1:n_anc]))
+merged_state = encoded_A ⊗ plus_anc ⊗ encoded_B
+
+adapter_bits = falses(nshots)
+for shot in 1:nshots
+    reg = Register(copy(merged_state), n_xstabs + n_zstabs)
+    mctrajectory!(reg, adapter_stabilizers)
+    bits = bitview(reg)
+    recipe_outcomes = [bits[n_xstabs + r] for r in recipe]
+    adapter_bits[shot] = reduce(xor, recipe_outcomes)
+end
+
+(count(adapter_bits), nshots)
+```
+
+## Step 7: compare the two protocols
+
+The joint-logical bit distributions sampled by the two protocols agree
+within Monte Carlo error:
+
+```@example adapter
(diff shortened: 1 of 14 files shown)


test/test_ecc_adapters.jl
Krastanov, line 1: This test file seems to depend on nonexistant data.

LLMs frequently make tests that have a bit "if" statement that just make sure that most of the time the test does not even run -- given the absence of actual data files, these tests seem to be exactly like that.

Also, LLMs generate test files that are very verbose -- there is probably a way to make this test file just as useful, but only about 200 lines long, given how much mocking/data input/etc is reinvented in the file.

src/ecc/adapters/types.jl
```diff
@@ -0,0 +1,150 @@
+"""
+$TYPEDEF
+
+A pair of logical Z operators we want to jointly measure via a universal adapter.
+`code1` and `code2` are CSS code blocks; `code1 === code2` triggers the intra-code
+path. `logical{1,2}_support` are sorted 1-indexed lists of qubits in their respective
```
Krastanov: LLMism in the docstring (it is julia, no need to specify 1-indexed)

src/ecc/adapters/types.jl
```diff
@@ -0,0 +1,150 @@
...
+end
+
+"""
+$TYPEDEF
+
+Auxiliary graph G for measuring a single logical, built by
+[`build_initial_aux_graph`](@ref) and finalised by
+[`cellulate_long_cycles!`](@ref). One graph per side of the adapter.
+
+`port_function` is set-valued per Appendix B of [swaroop2026universal](@cite) to
```
Krastanov: port_function seems to be a vector of vectors, this docstring is confusing

src/ecc/adapters/types.jl
```diff
@@ -0,0 +1,150 @@
...
+julia> z = sort(findall(!iszero, stab_to_gf2(logz_ops(Surface(3,3)))[1, 14:26]));
+
+julia> wrap = AdapterMergedCode(build_adapter(CodePair(c1, c2, z, z)));
+
+julia> (code_n(wrap), code_k(wrap), code_s(wrap))
+(33, 1, 32)
+```
+
+### Fields
+$TYPEDFIELDS
```
Krastanov: the user does not need to know the internal structure because there is a API of methods defined to be used as a way to interact with this struct

src/ecc/adapters/types.jl
```diff
@@ -0,0 +1,150 @@
...
+    "The merged CSS code containing Z̄_l Z̄_r as a stabilizer product."
+    merged_code::CSS
+end
+
+"""
+$TYPEDEF
+
+`AbstractCSSCode` wrapper around an [`Adapter`](@ref) so the merged code plugs into
+the standard `QECCore.AbstractECC` dispatch (`code_n`, `parity_matrix_x`, …) while
+still exposing the originating `Adapter` via `.adapter` for consumers that need the
```
Krastanov: given the other comment, remove the exposes `.adapter` commenter

Krastanov (changes requested):
- go through the docstrings to make sure they do not have a ton of LLMisms

- To make reviewing this a bit easier, and to make it a bit easier to use, and for us to be able to judge the API design choices, let's add a tutorial page that explains the following:

1. Take two codes and a specific logical operator between those codes, and give a list of measurements that need to be executed in order to measure that logical operator
2. give an example where such a measurement is done in the naive fashion (in simulation) vs with the adapter:

```
codeA
codeB

physical_state_A = random_stab...
physical_state_B = random_stab...

encoded_A = mctrajectory!(naive_encoding_circuit(codeA), physical_state_A)
encoded_B = ...

full_state = tensor(encoded_A, encoded_B)

logicalA = logicalzview(code_A)[1]
logicalB = ...
full_logical = tensor(logicalA, logicalB)

# full_state might need to be wrapped in a Register

measurement_results = mctrajectories(full_state, [projector_on_full_logical])

adapter_stabilizers = ...

measurement_results_adapter = mctrajectories(full_state, list_of_projectors_on_each_adapter_stabilizer)

# compare the statistics of measurement_results and measurement_results_adapter

# missing from this example is what to do at the end of the adapter measurements so that we are left with two separate codeblocks again
```

src/ecc/adapters/types.jl
```diff
@@ -0,0 +1,91 @@
...
+of the two logical Z operators to be jointly measured. `code1 === code2`
+selects the intra-code path. X- and Y-type joint measurements reduce to the
+Z-type case via local Cliffords (paper §IV).
+
+- `code1`, `code2`: the input codes
+- `logical1_support`, `logical2_support`: qubit indices supporting Z̄_l and Z̄_r
+"""
+struct CodePair
+    code1::CSS
+    code2::CSS
```
Krastanov: these should probably just be Abstract... code type and the constructor can check that parity_matrix_x/z does not throw a MethodError.

docs/src/adapter_tutorial.md
```diff
@@ -0,0 +1,298 @@
...
+```@example adapter
+using QuantumClifford
+using QuantumClifford: stab_to_gf2
+using QuantumClifford.ECC: CSS, Surface, parity_matrix_x, parity_matrix_z,
+                            logz_ops, logx_ops, code_n, code_k, parity_checks,
+                            naive_encoding_circuit
+using QuantumClifford.ECC.Adapters: CodePair, build_adapter, joint_logical_recipe
+
+as_css(c) = CSS(Matrix{Bool}(parity_matrix_x(c)),
+                Matrix{Bool}(parity_matrix_z(c)))
```
Krastanov: This conversion should not be necessary. Moreover, even if it is necessary, the type cast to `Matrix{Bool}` should not be something that the user is responsible to write.

src/ecc/adapters/merge.jl
```diff
@@ -0,0 +1,516 @@
...
+"""
+function build_adapter_intercode(
+    pair::CodePair;
+    max_cycle_len_1::Union{Int,Nothing} = nothing,
+    max_cycle_len_2::Union{Int,Nothing} = nothing,
+    chords_1::Union{Nothing, Vector{Tuple{Int,Int}}} = nothing,
+    chords_2::Union{Nothing, Vector{Tuple{Int,Int}}} = nothing,
+)::Adapter
+    pair.code1 !== pair.code2 ||
+        throw(ArgumentError("build_adapter_intercode: code1 and code2 must be distinct objects; use intra-code path for measurements on the same codeblock"))
```
Krastanov:
```suggestion
function build_adapter(
    code1, code2, logical1_support, logical2_support;
    max_cycle_len_1::Union{Int,Nothing} = nothing,
    max_cycle_len_2::Union{Int,Nothing} = nothing,
    chords_1::Union{Nothing, Vector{Tuple{Int,Int}}} = nothing,
    chords_2::Union{Nothing, Vector{Tuple{Int,Int}}} = nothing,
)::Adapter
    code1 !== code2 ||
        throw(ArgumentError("build_adapter_intercode: code1 and code2 must be distinct objects; use intra-code path for measurements on the same codeblock"))
```

src/ecc/adapters/merge.jl
```diff
@@ -0,0 +1,516 @@
...
+function build_adapter_intracode(
+    pair::CodePair;
+    max_cycle_len_1::Union{Int,Nothing} = nothing,
+    max_cycle_len_2::Union{Int,Nothing} = nothing,
+    chords_1::Union{Nothing, Vector{Tuple{Int,Int}}} = nothing,
+    chords_2::Union{Nothing, Vector{Tuple{Int,Int}}} = nothing,
+)::Adapter
+    pair.code1 === pair.code2 ||
+        throw(ArgumentError("build_adapter_intracode: code1 and code2 must be the same object; use build_adapter_intercode for distinct codeblocks"))
+    code = pair.code1
```
Krastanov:
```suggestion
function build_adapter(
    code, logical_support_1, logical_support_2;
    max_cycle_len_1::Union{Int,Nothing} = nothing,
    max_cycle_len_2::Union{Int,Nothing} = nothing,
    chords_1::Union{Nothing, Vector{Tuple{Int,Int}}} = nothing,
    chords_2::Union{Nothing, Vector{Tuple{Int,Int}}} = nothing,
)::Adapter
```

src/ecc/adapters/merge.jl
```diff
@@ -0,0 +1,516 @@
...
+(33, 1, 3)
+```
+"""
+function build_adapter(pair::CodePair; kwargs...)::Adapter
+    if pair.code1 === pair.code2
+        build_adapter_intracode(pair; kwargs...)
+    else
+        build_adapter_intercode(pair; kwargs...)
+    end
+end
```
Krastanov:
```suggestion
function build_adapter end
```

or completely delete and move the docstring content as appropriate

src/ecc/adapters/types.jl
```diff
@@ -0,0 +1,91 @@
...
+    T::SparseMatrixCSC{Bool, Int}
+    P::SparseMatrixCSC{Bool, Int}
+    target::Symbol
+end
+
+"""
+A universal adapter between two CSS codes — the result of [`build_adapter`](@ref).
+Subtypes `AbstractCSSCode`, so it plugs directly into the standard `QECCore`
+dispatch (`code_n`, `parity_matrix_x`, distance, decoder harnesses) without an
+intermediate wrapper.
```
Krastanov:
```suggestion
A universal adapter between two CSS codes — the result of [`build_adapter`](@ref).
Subtypes `AbstractCSSCode`, so it behaves like and can be used as any other stabilizer code in this library.
```

src/ecc/adapters/types.jl
```diff
@@ -0,0 +1,91 @@
...
+julia> c1 = CSS(Matrix{Bool}(parity_matrix_x(Surface(3,3))), Matrix{Bool}(parity_matrix_z(Surface(3,3))));
+
+julia> c2 = CSS(Matrix{Bool}(parity_matrix_x(Surface(3,3))), Matrix{Bool}(parity_matrix_z(Surface(3,3))));
+
+julia> z = sort(findall(!iszero, stab_to_gf2(logz_ops(Surface(3,3)))[1, 14:26]));
+
+julia> adapter = build_adapter(CodePair(c1, c2, z, z));
+
+julia> (code_n(adapter), code_k(adapter), code_s(adapter))
+(33, 1, 32)
```
Krastanov, line 99:
```suggestion
julia> using QuantumClifford, QECCore

julia> using QuantumClifford.ECC

julia> c1 = Surface(3,3);

julia> c2 = Surface(3,3);

julia> z = logz_ops(Surface(3,3))[1];

julia> adapter = build_adapter(c1, c2, z, z);

julia> (code_n(adapter), code_k(adapter), code_s(adapter))
(33, 1, 32)
```

for this to work, we will need to add an extra method to `build_adapter` that takes a Pauli string, verifies it is only Zs, verifies it is the correct length, and converts it to a "support".

I would suggest also adding public `QuantumClifford.supportz(::PauliOperator)` and `supportx` for ease of use.

docs/src/adapter_tutorial.md
```diff
@@ -0,0 +1,298 @@
...
+
+codeA = Surface(3, 3)
+codeB = Surface(3, 3)
+nA = code_n(codeA); kA = code_k(codeA)
+nB = code_n(codeB); kB = code_k(codeB)
+
+# Z-logical support on a single Surface(3,3) block
+lz = stab_to_gf2(logz_ops(codeA))
+z = sort(findall(!iszero, lz[1, nA+1:2nA]))
```
Krastanov:
```suggestion
# Z-logical operator to adapt
z = logz_ops(codeA)[1]
```

docs/src/adapter_tutorial.md
```diff
@@ -0,0 +1,298 @@
...
+merged-qubit layout:
+
+```@example adapter
+recipe = joint_logical_recipe(adapter)
+```
+
+This is the *list of measurements* that implements the joint logical operator:
+after standard syndrome extraction on the merged code, XOR the measurement
+outcomes at these indices to recover the joint
+``\bar Z_l \otimes \bar Z_r`` measurement result.
```
Krastanov:
```suggestion
`joint_logical_recipe` returns the row indices of the merged ``Z`` parity-check
matrix whose product ``\bar Z_l \otimes \bar Z_r`` in the original joint logical operator
we are building an adapter for:

```@example adapter
recipe = joint_logical_recipe(adapter)
```

The indices correspond to the row in `parity_checks(adapter)`.
```

File: docs/src/adapter_tutorial.md [hunk 14]
```diff
@@ -0,0 +1,298 @@
...
+    p = zero(PauliOperator, n)
+    for i in 1:n
+        row[i] != 0 && (p[i] = (false, true))
+    end
+    p
+end
+
+n_total = code_n(adapter)
+Hz = parity_matrix_z(adapter)
+recipe_paulis = [row_to_zpauli(Hz[r, :], n_total) for r in recipe]
```
Krastanov:
```suggestion
H = parity_checks(adapter)
recipe_paulis = H[recipe]
```

docs/src/adapter_tutorial.md
```diff
@@ -0,0 +1,298 @@
...
+recipe = joint_logical_recipe(adapter)
+```
+
+This is the *list of measurements* that implements the joint logical operator:
+after standard syndrome extraction on the merged code, XOR the measurement
+outcomes at these indices to recover the joint
+``\bar Z_l \otimes \bar Z_r`` measurement result.
+
+## Step 3: algebraic verification of the recipe
+
```
Krastanov, line 74:
```suggestion
## Step 3: algebraic verification of the recipe

This step is simply a verification for correctness, not necessary in actual use.

```

docs/src/adapter_tutorial.md
```diff
@@ -0,0 +1,298 @@
...
+The cancellation on the auxiliary qubits is by design: the ``V_l`` row block
+contributes ``Z`` on every column of the adapter sub-block ``A`` (because the
+SkipTree permutation ``P_l`` places exactly one ``1`` per ``A`` column), and
+the ``V_r`` row block contributes the same ``Z`` on ``A``. Since
+``Z^2 = I`` over ``\mathbb F_2``, the auxiliary support cancels and only the
+joint logical survives on the data qubits.
+
+## Step 4: prepare encoded logical states in simulation
+
```
Krastanov, line 105:
```suggestion
## Step 4: prepare encoded logical states in simulation

The next two steps are also simply a verification of correctness, this type by simulating either the measurement of the adapter we have built, or the much more naive (and not fault tolerant or easy to implement on real hardware) direct measurement of the large cross-code logical operator without the use of an adapter.

```

docs/src/adapter_tutorial.md
```diff
@@ -0,0 +1,298 @@
...
+We follow the standard `QuantumClifford` pattern (also used in the ECC
+syndrome tests): a random ``k``-qubit logical input is buffered with
+``n-k`` ``|0\rangle`` ancilla qubits and passed through
+`naive_encoding_circuit`. This produces a fully-encoded `MixedDestabilizer`
+state of the surface code.
+
+```@example adapter
+using Random; Random.seed!(2026)
```
Krastanov:
```suggestion
using Random; Random.seed!(2026) # hide
```

not sure this actually works in example blocks, we should double check

docs/src/adapter_tutorial.md
```diff
@@ -0,0 +1,298 @@
...
+nshots = 1000
+
+naive_bits = falses(nshots)
+for shot in 1:nshots
+    reg = Register(copy(full_state), 1)
+    mctrajectory!(reg, [PauliMeasurement(full_logical, 1)])
+    naive_bits[shot] = bitview(reg)[1]
+end
+
+(count(naive_bits), nshots)
```
Krastanov, line 163: TODO on merging this remind me to make an issue for fixing this mctrajectory API -- this is very boilerplaty for something that should be a single call

docs/src/adapter_tutorial.md
```diff
@@ -0,0 +1,298 @@
...
+```
+
+The pair `(count(naive_bits), nshots)` reports how many shots returned
+``-1`` (i.e. measured `true`). For the seed-pinned random logical input above,
+the empirical distribution should be near ``50/50``.
```
Krastanov:
```suggestion
The pair `(count(naive_bits), nshots)` reports how many shots measured `true`.
The fraction will be 0 or 1 for deterministic measurement or 0.5 for a logical operator
that does not stabilizer the current logical state.
```

docs/src/adapter_tutorial.md
```diff
@@ -0,0 +1,298 @@
...
+
+## Step 6: adapter-mediated measurement of the same logical
```
Krastanov, line 171:
```suggestion
## Step 6: adapter-mediated measurement of the same logical

Here we get back to actually using the adapters, the proper way to do fault-tolerant logical measurements and see that it gives the same result as the naive method from above.
This adapter approach is what you would actually want when doing real QEC modeling work.
```

Krastanov (changes requested):
this looks good, have not really had a chance to review everything, have mostly focused on the example tutorial as a way to vet the API

PR #743, opened 2026-06-10
Add noisify API with structured NoiseModel (issue #729)

### LINKED ISSUE #729: Add a dispatch-based API for turning noiseless circuits into noisy circuits

---

QuantumClifford already has useful building blocks for noisy simulations:
`NoiseOp`, `NoiseOpAll`, `NoisyGate`, `PauliNoise`,
`UnbiasedUncorrelatedNoise`, `mctrajectories`, and `pftrajectories`. However,
users currently have to manually insert noise operations into circuit vectors or
write one-off helper functions. A small dispatch-based `noisify` API would make
it much easier to define a noise model once and apply it consistently to a
noiseless circuit.

Relevant links:

- Noise operations: https://github.com/QuantumSavory/QuantumClifford.jl/blob/main/src/noise.jl
- Affected-qubit helpers: https://github.com/QuantumSavory/QuantumClifford.jl/blob/main/src/affectedqubits.jl
- Symbolic operation types: https://github.com/QuantumSavory/QuantumClifford.jl/blob/main/src/symbolic_cliffords.jl
- Existing noisy-circuit docs: https://qc.quantumsavory.org/stable/noisycircuits_ops/

A simple noise model should be easy to apply to a vector of operations:

```julia
using QuantumClifford

circuit = [sHadamard(1), sCNOT(1, 2), sMZ(1, 1)]
noisy = noisify(circuit, PauliNoise(1e-3, 1e-3, 1e-3))
```

The core behavior can start from this multiple-dispatch pattern:

```julia
noisify(circuit::AbstractVector, noise) = reduce(vcat, noisify.(circuit, (noise,)))
noisify(anything, noise) = [anything]
noisify(g::AbstractSingleQubitOperator, noise) = [NoiseOp(noise, affectedqubits(g)), g]
noisify(g::AbstractTwoQubitOperator, noise) = [NoiseOp(noise, affectedqubits(g)), g]
```

For practical use, there should also be a structured configuration object for
different noise locations:

```julia
noise = CircuitNoise(
    single_qubit = PauliNoise(1e-4, 1e-4, 1e-4),
    two_qubit = PauliNoise(1e-3, 1e-3, 1e-3),
    idle = PauliNoise(1e-5, 1e-5, 1e-5),
    measurement = PauliNoise(2e-3, 2e-3, 2e-3),
)

noisy = noisify(circuit, noise; nqubits=2)
frames = pftrajectories(noisy; trajectories=10_000)
```

The exact names can change, but the API should preserve the ability to extend
behavior with method definitions for new operation types and new noise-model
types.

## Suggested implementation

Implement a small `CircuitNoise` or `NoiseModel` type with fields for common
locations:

- single-qubit gate noise
- two-qubit gate noise
- idle/waiting noise
- measurement noise
- reset noise, if separate from measurement noise

Also consider a `NoNoise` sentinel for omitted locations. This avoids requiring
users to pass `nothing` and makes dispatch cleaner.

Useful behavior to define:

- `noisify(circuit::AbstractVector, noise_model; nqubits=nothing)` returns a new
  vector and does not mutate the input.
- Single-qubit and two-qubit Clifford gates get the configured gate noise on
  their affected qubits.
- Measurements get the configured measurement noise. It is acceptable for the
  first version to model this as pre-measurement Pauli noise on the measured
  qubits, as long as the semantics are documented.
- `NoiseOp` and `NoisyGate` should not accidentally get double-noisified unless
  explicitly requested.
- Classical-only operations such as `ClassicalXOR` should pass through unchanged.
- Idle noise can be added to `setdiff(1:nqubits, affectedqubits(op))` for each
  operation when `nqubits` is provided. If idle noise is requested and the number
  of qubits cannot be inferred, throw a helpful error.

## In scope

- A documented `noisify` API.
- A structured noise configuration type.
- Dispatch methods for single-qubit gates, two-qubit gates, measurements, reset
  measurements, existing noise operations, and classical operations.
- Tests that verify the inserted operation order and affected qubits.
- Tests showing the resulting noisy circuit can be simulated with
  `pftrajectories` or `mctrajectories`.
- Documentation with a minimal example and a configuration example.

## Out of scope

- Automatic circuit scheduling or momentization.
- Importing hardware calibration files.
- General non-Pauli quantum channels.
- Rewriting the existing trajectory simulators.

## Acceptance criteria

- `noisify([sHadamard(1), sCNOT(1, 2)], noise)` returns a circuit containing
  `NoiseOp` entries on the expected qubits.
- A structured model can assign different noise to one-qubit gates, two-qubit
  gates, idle qubits, and measurements.
- The original circuit is not mutated.
- Existing explicit `NoiseOp` operations are handled in a documented way.
- Tests cover the simple dispatch example and the structured configuration case.
- The docs explain where the noise is inserted relative to each operation.

yi-hsiu-yang:
Closes #729

This PR adds the `noisify` API for inserting noise operations into a circuit.

## What's implemented

- **Simple API**: `noisify(circuit, noise)` applies the same noise to all gate / measurement / reset operations.
- **Structured API**: `noisify(circuit, NoiseModel(...))` allows different noise per operation type (single-qubit, two-qubit, measurement, reset, idle).
- **Idle noise**: When `nqubits` is provided, noise is also inserted on qubits not touched by each operation. A clear error is raised when idle noise is requested without `nqubits`.
- **Fallback**: Unknown operations (e.g. classical operations) pass through unchanged.

## Tests

Added 9 tests in `test/test_noisify.jl` covering the simple API, structured `NoiseModel`, idle noise insertion, error handling, insertion order, and pass-through behavior.

## Note

I'm a first-time contributor. All code was written by me; I used an AI assistant only to learn Julia syntax and discuss design choices conceptually, not to generate the implementation. Happy to adjust based on review feedback.

### DIFF
--- a/test/test_noisify.jl
+++ b/test/test_noisify.jl
@@ -1,0 +1,28 @@
+@testitem "noisify" begin
+    using QuantumClifford
+
+    @testset "simple API" begin
+        result = noisify([sHadamard(1)], PauliNoise(1e-3, 1e-3, 1e-3))
+        @test length(result) == 2
+        @test result[2] == sHadamard(1)
+    end
+
+    @testset "errors when idle requested without nqubits" begin
+        @test_throws ErrorException noisify([sHadamard(1)], DefaultNoiseModel())
+    end
+
+    @testset "noise inserted before op with correct qubits" begin
+        result = noisify([sCNOT(1, 2)], PauliNoise(1e-3, 1e-3, 1e-3))
+        @test length(result) == 2
+        @test result[1] isa NoiseOp
+        @test result[2] == sCNOT(1, 2)
+        @test affectedqubits(result[1]) == (1, 2)
+    end
+
+    @testset "unknown operations pass through unchanged" begin
+        op = ClassicalXOR((1, 2), 3)
+        result = noisify([op], PauliNoise(1e-3, 1e-3, 1e-3))
+        @test length(result) == 1
+        @test result[1] === op
+    end
+end
--- a/test/test_noisify.jl
+++ b/test/test_noisify.jl
@@ -21,8 +21,21 @@
 
     @testset "unknown operations pass through unchanged" begin
         op = ClassicalXOR((1, 2), 3)
-        result = noisify([op], PauliNoise(1e-3, 1e-3, 1e-3))
+        result = noisify([op], PauliNoise(1e-3, 1e-3, 1e-3); nqubits=3)
         @test length(result) == 1
         @test result[1] === op
+    end
+    @testset "structured noise model with all op types and idle" begin
+        nm = NoiseModel(
+            single_qubit = PauliNoise(1e-4, 1e-4, 1e-4),
+            two_qubit    = PauliNoise(1e-3, 1e-3, 1e-3),
+            measurement  = PauliNoise(2e-3, 2e-3, 2e-3),
+            reset        = PauliNoise(5e-3, 5e-3, 5e-3),
+            idle         = PauliNoise(1e-5, 1e-5, 1e-5),
+        )
+        circuit = [sHadamard(1), sCNOT(1, 2), sMZ(2, 1), sMRZ(1, 2)]
+        result = noisify(circuit, nm; nqubits=3)
+        @test length(result) > length(circuit)
+        @test result[end] == sMRZ(1, 2)
     end
 end
--- a/test/test_noisify.jl
+++ b/test/test_noisify.jl
@@ -8,7 +8,9 @@
     end
 
     @testset "errors when idle requested without nqubits" begin
-        @test_throws ErrorException noisify([sHadamard(1)], DefaultNoiseModel())
+        @test_throws "nqubits must be provided" noisify(
+            [sHadamard(1)], DefaultNoiseModel()
+        )
     end
 
     @testset "noise inserted before op with correct qubits" begin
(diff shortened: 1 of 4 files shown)


test/test_noisify.jl
```diff
@@ -0,0 +1,41 @@
...
+    using QuantumClifford
+
+    @testset "simple API" begin
+        result = noisify([sHadamard(1)], PauliNoise(1e-3, 1e-3, 1e-3))
+        @test length(result) == 2
+        @test result[2] == sHadamard(1)
+    end
+
+    @testset "errors when idle requested without nqubits" begin
+        @test_throws ErrorException noisify([sHadamard(1)], DefaultNoiseModel())
```
Krastanov: a `test_throws` should generally speaking not be this permissive -- you need to check it was actually the error you expected, not some completely unrelated error (thus hiding an actual bug, making the test actively harmful)

Krastanov (changes requested):
this does not really seem to distinguish the single-qubit, two-qubit, idling, measurement, and reset noise -- all are just doing the same thing. A way to prove this works is to show a noisified circuit and where each different type of noise ended up in it.

test/test_noisify.jl
yi-hsiu-yang: Thanks for the review @Krastanov! I've pushed an update addressing both points, and would like to flag one design question that came up while working through the dispatch verification.

## 1. Tests now explicitly verify dispatch routing

I've added a new testset `"noisify dispatches NoiseModel fields correctly"` that exercises each dispatch path with **visually distinguishable** noise per field: different probabilities (0.01–0.05) and different Pauli biases (X-only / Y-only / Z-only / depolarizing),  and asserts that each `NoiseOp` points back to the correct `NoiseModel` field, on the correct qubits.

Here's the noisified output for a 3-qubit, 5-operation circuit (5 ops → 15 ops):

```
 [1]  NoiseOp(PauliNoise(0.05, 0.0, 0.0), (2, 3))   ← idle on q2, q3
 [2]  NoiseOp(PauliNoise(0.01, 0.0, 0.0), (1,))     ← single_qubit on q1
 [3]  sHadamard(1)
 [4]  NoiseOp(PauliNoise(0.05, 0.0, 0.0), (3,))     ← idle on q3
 [5]  NoiseOp(PauliNoise(0.0, 0.02, 0.0), (1, 2))   ← two_qubit on q1, q2
 [6]  sCNOT(1, 2)
 [7]  NoiseOp(PauliNoise(0.05, 0.0, 0.0), (1, 3))   ← idle on q1, q3
 [8]  NoiseOp(PauliNoise(0.01, 0.0, 0.0), (2,))     ← single_qubit on q2
 [9]  sX(2)
[10]  NoiseOp(PauliNoise(0.05, 0.0, 0.0), (2, 3))   ← idle on q2, q3
[11]  NoiseOp(PauliNoise(0.0, 0.0, 0.03), (1,))     ← measurement on q1
[12]  sMZ(1)
[13]  NoiseOp(PauliNoise(0.05, 0.0, 0.0), (1, 2))   ← idle on q1, q2
[14]  NoiseOp(PauliNoise(0.04, 0.04, 0.04), (3,))   ← reset on q3
[15]  sMRZ(3)
```

Every gate is preceded by (a) idle noise on the non-affected qubits and (b) the op-specific noise from the matching `NoiseModel` field. The testset asserts both `noise` and `indices` for each `NoiseOp` and that the original gate is preserved at the expected position. Because each field uses a distinct `PauliNoise` parameter, any misrouted dispatch would fail the assertion.

## 2. `@test_throws` now matches the specific error message

Changed `@test_throws ErrorException ...` to `@test_throws "nqubits must be provided" ...`, so the test only passes on the intended error path rather than on any unrelated `ErrorException` that might be thrown.

## 3. Design question: should `noise_model.reset` apply to plain `Reset`?

I noticed that `Reset` (without measurement) currently falls through to the no-op fallback, it picks up `noise_model.idle` on non-affected qubits but not `noise_model.reset`:

```julia
julia> noisify([Reset(S"Z", [1])], model; nqubits=3)
2-element Vector{Any}:
 NoiseOp(PauliNoise(0.05, 0.0, 0.0), (2, 3))   # idle on q2, q3
 Reset(...)                                     # no reset noise on q1
```

The reset dispatch is currently `noisify(g::AbstractResetMeasurement, ...)`, which covers `sMRZ` and friends but not `Reset` itself (which is `AbstractOperation` but not a subtype of `AbstractResetMeasurement`).

Is the intended semantics that `noise_model.reset` applies only to measure-and-reset operations (current behavior), or should it also apply to plain `Reset`?

PR #745, opened 2026-06-11
Import Stim detector error models for Pauli-frame sampling

### LINKED ISSUE #728: Import Stim detector error models for Pauli-frame-style sampling

---

Many quantum error-correction workflows use Stim detector error model (`.dem`)
files as an interchange format. QuantumClifford has fast Pauli-frame simulation
machinery, but there is currently no convenient way to ingest an existing `.dem`
file and turn it into a QuantumClifford circuit. Importing `.dem` files as
circuits would let users compare Stim-generated models with QuantumClifford
decoders and analysis tools while continuing to use the existing
`pftrajectories` backend for sampling.

Relevant links:

- Pauli-frame implementation: https://github.com/QuantumSavory/QuantumClifford.jl/blob/main/src/pauli_frames.jl
- Existing noisy circuit operations: https://github.com/QuantumSavory/QuantumClifford.jl/blob/main/src/noise.jl

A user should be able to load a `.dem` file and receive a corresponding
QuantumClifford circuit that can be passed directly to `pftrajectories`:

```julia
using QuantumClifford

circuit = read_detector_error_model("surface_code.dem")

measurements(frames)
```

The exact function name can change, but the output should be a circuit-like
object, a vector of operations, that is accepted by the
existing `pftrajectories` methods:

```julia
circuit = detector_error_model_circuit("surface_code.dem")
measurements(frames)
```

The important requirement is that sampling remains implemented by
`pftrajectories`. This bounty should not introduce a separate sampling backend.

## Suggested implementation

Implement a parser for the useful subset of Stim detector error model syntax:

- `error(p) D... L...`
- `detector ...`
- `logical_observable ...`
- `shift_detectors ...`
- `repeat K { ... }`
- comments and blank lines

Coordinates on `detector` or `shift_detectors` lines can be parsed and stored if
convenient, but it is acceptable to ignore them initially as long as the detector
index shifts are handled correctly.

The parser should lower the `.dem` file to operations that `pftrajectories` can
execute. If the existing gate set is not expressive enough, add a small number of
new circuit operations for detector error model semantics. For example, an
operation like this may be useful:

```julia
DetectorError(probability, detector_bits, logical_bits)
```

When applied to a `PauliFrame`, this operation would sample one independent
Bernoulli event per trajectory and XOR the corresponding detector and logical
bits into the frame's measurement storage. A no-op declaration operation may also
be needed so explicitly declared detectors or logical observables with no errors
still allocate output columns.

The semantics should match Stim's `.dem` documentation: each `error(p)` line is
an independent error mechanism, and when it occurs it flips the listed detector
and logical targets. The implementation should express that behavior through
QuantumClifford circuit operations, then rely on `pftrajectories` to run many
trajectories efficiently.

## Minimal example

Given this file:

```text
detector D0
detector D1
logical_observable L0
error(0.1) D0
error(0.2) D0 D1 L0
```

the importer should produce a model with two detectors, one logical observable,
and two independent error mechanisms. The returned circuit should be accepted by
`pftrajectories`; sampling that circuit many times should produce detector and
logical outcome frequencies consistent with those probabilities.

## In scope

- A documented circuit representation for imported detector error models.
- A parser for common Stim `.dem` files, including nested `repeat` blocks and
  `shift_detectors`.
- Any small new `AbstractOperation` subtypes needed to represent `.dem`
  semantics in circuits.
- `apply!`/`affectedbits` support needed for those operations to work with
  `PauliFrame` and `pftrajectories`.
- Tests comparing small hand-written `.dem` files against known expected
  detector/logical distributions.
- Optional statistical comparison against Stim's own sampler when Stim is
  available.
- Clear error messages for unsupported syntax.

## Out of scope

- Parsing Stim circuit (`.stim`) files.
- Decoding the sampled detector data.
- A new sampling engine separate from `pftrajectories`.
- Exact bit-for-bit matching of Stim's random number generator.
- Full preservation of coordinate metadata, unless this is straightforward.

## Acceptance criteria

- `read_detector_error_model(path)` or an equivalent API imports a `.dem` file
  and returns a circuit-like object accepted by `pftrajectories`.
- The parser handles `error`, `detector`, `logical_observable`,
  `shift_detectors`, and `repeat` for representative files.
- New `.dem`-specific operations, if added, implement the `affectedbits` and
  `apply!` methods needed by `pftrajectories`.
- `pftrajectories(imported_circuit; trajectories=...)` returns detector and
  logical bit arrays with documented dimensions.
- Statistical tests validate the imported circuit on small models.
- Documentation includes a short imported `.dem` example and the expected output
  shape.

GrovyleX:
Closes #728

Adds a way to load Stim `.dem` files into QuantumClifford and sample them with the existing `pftrajectories` backend. No separate sampler - each `error(p)` line becomes one `DetectorError` op that flips detector/observable measurement columns with probability `p`.

### What changed

- added `src/dem.jl` to import `.dem` files via `read_detector_error_model` / `parse_detector_error_model`
- new ops `DetectorError` and `DeclareMeasurementBits` so `pftrajectories` can sample detector/observable flips directly
- parser covers the usual Stim subset (`error`, `detector`, `logical_observable`, `shift_detectors`, `repeat`, comments/tags); coords and `^` separators are ignored for sampling
- docs + tests (`test/test_dem.jl`, including a statistical sanity check on flip rates)
- CHANGELOG + version bump to 0.11.5-dev

### Output layout

After `pftrajectories`, `measurements(frames)` is `trajectories × (num_detectors + num_observables)`. Detectors are the first columns, observables after. Values are relative to the noiseless reference run, same as other Pauli-frame sims.

### Notes

- `repeat` blocks expand eagerly (same idea as Stim, but a huge repeat count means a huge op vector)
- mechanisms with more than 16 targets fall back to the non-compactified path (warning from `pftrajectories`) - should not matter for typical graphlike DEMs

### Test plan

- ran `test/test_dem.jl` via TestItemRunner locally (34 tests)
- checked the minimal issue example: empirical freqs ~ [0.26, 0.2, 0.2] vs analytic values
- smoke-tested `repeat` / `shift_detectors` / nested blocks / file read

### AI usage disclosure

I used Claude (Claude Code) for some planning and early implementation scaffolding on this PR. I reviewed the full diff myself, adjusted code to match repo style, fixed parser edge cases, wrote and ran the tests (including statistical checks).

I have read and understood the submitted code, and I ran the tests locally before opening this PR.

If you are submitting for a bug bounty:
- [x] I have read and followed the [AI usage and disclosure policy](https://github.com/QuantumSavory/.github/blob/main/BUG_BOUNTIES.md)

### DIFF
--- a/src/dem.jl
+++ b/src/dem.jl
@@ -38,9 +38,6 @@ function apply!(frame::PauliFrame, op::DetectorError)
     return frame
 end
 
-# The reference trajectory of a Register is noiseless, so the mechanism never fires there.
-apply!(r::Register, ::DetectorError) = r
-
 affectedqubits(::DetectorError) = ()
 affectedbits(op::DetectorError) = op.bits
 
@@ -56,7 +53,6 @@ struct DeclareMeasurementBits <: AbstractOperation
 end
 
 apply!(frame::PauliFrame, ::DeclareMeasurementBits) = frame
-apply!(r::Register, ::DeclareMeasurementBits) = r
 
 affectedqubits(::DeclareMeasurementBits) = ()
 affectedbits(op::DeclareMeasurementBits) = (op.nbits,)
@@ -76,8 +72,8 @@ reference run.
 
 See also: [`read_detector_error_model`](@ref).
 """
-struct DetectorErrorModelCircuit <: AbstractVector{AbstractOperation}
-    circuit::Vector{AbstractOperation}
+struct DetectorErrorModelCircuit <: AbstractVector{Union{DetectorError,DeclareMeasurementBits}}
+    circuit::Vector{Union{DetectorError,DeclareMeasurementBits}}
     num_detectors::Int
     num_observables::Int
 end
@@ -113,16 +109,9 @@ for sampling (the mechanism flips every listed target) and coordinates on
 `detector`/`shift_detectors` lines are ignored too. Unsupported instructions
 raise an `ArgumentError`.
 
-See also: [`detector_error_model_circuit`](@ref), [`parse_detector_error_model`](@ref).
+See also: [`parse_detector_error_model`](@ref).
 """
 read_detector_error_model(path::AbstractString) = parse_detector_error_model(read(path, String))
-
-"""
-$(TYPEDSIGNATURES)
-
-Alias for [`read_detector_error_model`](@ref).
-"""
-detector_error_model_circuit(path::AbstractString) = read_detector_error_model(path)
 
 """
 $(TYPEDSIGNATURES)
@@ -305,7 +294,7 @@ function _dem_build_circuit(state)
     D = state.max_detector + 1
     L = state.max_observable + 1
     D + L == 0 && throw(ArgumentError("the detector error model declares no detectors or logical observables"))
-    circuit = AbstractOperation[]
+    circuit = Union{DetectorError,DeclareMeasurementBits}[]
     for (p, dets, obs) in state.mechanisms
         push!(circuit, DetectorError(p, [d + 1 for d in dets], [D + o + 1 for o in obs]))
     end
--- a/src/dem.jl
+++ b/src/dem.jl
@@ -178,11 +178,13 @@ function _dem_parse_nodes(lines, i, depth)
         elseif line == "{"
             throw(ArgumentError("unexpected '{' in detector error model"))
         elseif _dem_head(line) == "repeat"
-            m = match(r"^repeat\s+(\d+)$"i, line)
-            m === nothing && throw(ArgumentError("malformed repeat instruction: \"$line\""))
+            parts = split(line)
+            length(parts) == 2 || throw(ArgumentError("malformed repeat instruction: \"$line\""))
+            reps = _dem_parse_int(parts[2], line)
+            reps >= 0 || throw(ArgumentError("repeat count must be non-negative in line: \"$line\""))
             (i < length(lines) && lines[i+1] == "{") || throw(ArgumentError("expected '{' after \"$line\""))
             body, i = _dem_parse_nodes(lines, i + 2, depth + 1)
-            push!(nodes, (parse(Int, m.captures[1]), body))
+            push!(nodes, (reps, body))
         else
             push!(nodes, line)
             i += 1
--- a/src/dem.jl
+++ b/src/dem.jl
@@ -1,0 +1,312 @@
+# Importing Stim detector error models (`.dem` files) as Pauli frame circuits.
+# Every `error` line becomes an independent `DetectorError` operation and the
+# sampling is left to `pftrajectories`. The format is documented at
+# https://github.com/quantumlib/Stim/blob/main/doc/file_format_dem_detector_error_model.md
+
+"""
+$(TYPEDEF)
+
+A single error mechanism of a detector error model. For each Pauli frame it fires
+with probability `p`, and when it does it flips the measurement bits in `bits`.
+
+The entries of `bits` index columns of the [`PauliFrame`](@ref) measurement
+buffer. [`read_detector_error_model`](@ref) places the detectors in the first
+columns and the logical observables after them, so one `DetectorError` can flip
+both kinds of bit.
+
+See also: [`read_detector_error_model`](@ref), [`DeclareMeasurementBits`](@ref).
+"""
+struct DetectorError{N} <: AbstractOperation
+    "The probability that the mechanism fires."
+    p::Float64
+    "The measurement bit columns flipped when it fires."
+    bits::NTuple{N,Int}
+end
+
+DetectorError(p, bits::Tuple) = DetectorError{length(bits)}(float(p), bits)
+DetectorError(p, detector_bits, logical_bits) = DetectorError(p, (detector_bits..., logical_bits...))
+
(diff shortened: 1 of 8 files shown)


Krastanov:
We do not accept LLM-generated PRs by unknown developers for our bounties.

Krastanov:
Apologies, this was trigger happy on me -- I did not notice that you had actually done the checkboxes and that this is not completely bot-written.

GrovyleX:
I am glad you noticed. I will continue working with this PR. 1 CI check is failing. I will try fixing that. Apart from that will working on improving the PR... Is possible give some review so that i can improve the PR

src/dem.jl
```diff
@@ -0,0 +1,314 @@
...
+            for b in op.bits
+                frame.measurements[f, b] ⊻= true
+            end
+        end
+    end
+    return frame
+end
+
+# The reference trajectory of a Register is noiseless, so the mechanism never fires there.
+apply!(r::Register, ::DetectorError) = r
```
Krastanov: I am not quite sure I agree with this decision. Is there anywhere else in the codebase that we talk about reference trajectories on a Register? If not, I think it is best to just delete these methods, as they would just cause more confusion.

src/dem.jl
```diff
@@ -0,0 +1,314 @@
...
+"""
+struct DeclareMeasurementBits <: AbstractOperation
+    nbits::Int
+end
+
+apply!(frame::PauliFrame, ::DeclareMeasurementBits) = frame
+apply!(r::Register, ::DeclareMeasurementBits) = r
+
+affectedqubits(::DeclareMeasurementBits) = ()
+affectedbits(op::DeclareMeasurementBits) = (op.nbits,)
```
Krastanov, line 58: Why is this necessary? Is it used anywhere?

src/dem.jl
```diff
@@ -0,0 +1,314 @@
...
+See also: [`detector_error_model_circuit`](@ref), [`parse_detector_error_model`](@ref).
+"""
+read_detector_error_model(path::AbstractString) = parse_detector_error_model(read(path, String))
+
+"""
+$(TYPEDSIGNATURES)
+
+Alias for [`read_detector_error_model`](@ref).
+"""
+detector_error_model_circuit(path::AbstractString) = read_detector_error_model(path)
```
Krastanov: Why is this alias created? I do not understand how it is helpful.

src/dem.jl
```diff
@@ -0,0 +1,314 @@
...
+After sampling, `measurements(frame)` is a `trajectories × (num_detectors +
+num_observables)` `Bool` matrix. The first `num_detectors` columns are the
+detector outcomes `D0, D1, …` and the rest are the observable outcomes
+`L0, L1, …`. As always for Pauli frames, these are relative to the noiseless
+reference run.
+
+See also: [`read_detector_error_model`](@ref).
+"""
+struct DetectorErrorModelCircuit <: AbstractVector{AbstractOperation}
+    circuit::Vector{AbstractOperation}
```
Krastanov: Should this be abstract operation vector? It seems it can be much more concrete.

In what way is this structure better than just having a list of DetectorErrors?

Krastanov (commented):
I gave this a quick first skim. I have a few questions below before I can do a full review. Thanks for looking into this.

src/dem.jl
GrovyleX: Agreed, no precedent for reference-trajectory semantics on Register and DEM circuits only ever run through the frame-path pftrajectories(circuit). Removed both Register methods.

src/dem.jl
GrovyleX: It's load-bearing. pftrajectories sizes the measurement buffer from maximum(affectedbits(...)) over the circuit, so a detector/observable that's declared but never appears in an error line would have its column dropped and the output width would disagree with Stim. read_detector_error_model appends one DeclareMeasurementBits(num_detectors+num_observables) to reserve the full width - test_dem.jl covers exactly this (detector D3 with only error D0 ⇒ still 4 columns). Happy to rename if DeclareMeasurementBits isn't the clearest name.

src/dem.jl
GrovyleX: No god reason - the issue listed two candidate names and I kept both by mistake. Dropped the alias, keeping read_detector_error_model.

src/dem.jl
GrovyleX: Fixed the element type - it's only DetectorError + the single DeclareMeasurementBits, so it's now Vector{Union{DetectorError,DeclareMeasurementBits}} (compactify handles fast dispatch in the hot loop).
  The wrapper exists to carry num_detectors/num_observables so callers can split measurements(frames) into detector vs observable columns - the documented output dims the issue asks for. If you'd rather I return a plain vector and surface the counts another way, I'll switch - your call.

GrovyleX:
I chekc the CI.. the only red check is Test QECCore … windows-latest x64 t=1, which errors inside lib/QECCore/test/runtests.jl (1 of 45446, Windows only). This PR doesn't touch lib/QECCore/, and the job tests QECCore against released QuantumClifford, so it looks like a pre-existing flake on the Windows runner rather than something from this change. [If confirmed:] the same test errors on master's Windows job. Could you re-run it? Happy to look closer if you think it's related.

PR #747, opened 2026-06-11
Add Stim detector error model importer

### LINKED ISSUE #728: Import Stim detector error models for Pauli-frame-style sampling

DennisWayo:
Fixes #728.

This PR adds support for importing a focused subset of Stim detector error model (`.dem`) files into QuantumClifford as circuit-like objects that can be sampled with `pftrajectories`.

The imported model stores measurement columns as detectors first, followed by logical observables. Helper functions are provided to split sampled `PauliFrame` measurements back into detector and logical-observable matrices.

## Changes

- Added `DetectorErrorModelCircuit`, a vector-like circuit container accepted by `pftrajectories`.
- Added `DetectorError`, an independent DEM error mechanism that XORs detector/logical measurement bits with the requested probability.
- Added `DetectorErrorModelDeclare`, a no-op operation used to reserve detector/logical measurement columns even when a model has declarations without sampled errors.
- Added public import helpers:
  - `read_detector_error_model(path)`
  - `detector_error_model_circuit(path)`
  - `detector_error_model_circuit(io)`
- Added public metadata/extraction helpers:
  - `detector_count(circuit)`
  - `logical_observable_count(circuit)`
  - `detector_measurements(frame, circuit)`
  - `logical_observable_measurements(frame, circuit)`
- Added docs showing how to import and sample a `.dem` model.
- Added focused tests for parsing, sampling, repeats, shifts, duplicate-target cancellation, path loading, and unsupported syntax errors.

## DEM Syntax Support

This importer intentionally supports the subset needed to sample detector/logical error mechanisms through the existing Pauli-frame trajectory machinery:

- `error(p) D... L...`
- `detector D...`
- `logical_observable L...`
- `shift_detectors n`
- nested `repeat K { ... }` blocks
- comments and blank lines
- `^` separators, ignored as decomposition hints while preserving duplicate-target XOR cancellation semantics

Unsupported instructions and unsupported targets throw `ArgumentError`s with source and line-number context.

## Notes

Stim DEM detector targets are zero-based. QuantumClifford measurement columns are one-based, so imported detector bits are mapped to columns `1:detector_count`, and logical observable bits are mapped after those columns.

Duplicate targets in a single `error` instruction cancel by XOR semantics. For example, `error(1.0) D0 ^ D0` produces no net detector flip.

`shift_detectors` updates the detector offset during parsing, including inside nested repeat blocks. Repeated blocks are expanded into independent `DetectorError` operations.

## Validation

Focused checks run:

```text
git diff --check
```

Result: passed.

Direct importer smoke tests were run for:

- basic detector and logical sampling
- declarations allocating measurement columns without sampled errors
- nested `repeat` blocks
- `shift_detectors`
- duplicate target cancellation
- ignored `^` separators
- path-based loading
- unsupported syntax errors
- sampled frequencies matching configured error probabilities

Full `Pkg.test` was not completed locally because the broad test run attempted optional doctest/Tesseract setup and hit an unrelated `PyTesseractDecoder` Python dependency issue. The focused importer behavior was verified directly.

## Checklist

- [x] I have checked that the new code is formatted consistently and includes comments where they add useful context.
- [x] I have documented the new user-facing functionality in the project documentation.
- [x] I have added tests covering the new functionality.
- [ ] The full automated test suite passes on GitHub Actions.
- [x] I understand that formatting checks are now enforced, and I will fix any formatting issues reported in code I hand changed.

## OSS Contributions
https://github.com/PennyLaneAI/pennylane/pull/9566
https://github.com/PennyLaneAI/pennylane/pull/9604
https://www.linkedin.com/posts/activity-7443523751436091392-depM?utm_source=share&utm_medium=member_desktop&rcm=ACoAABfS9T0BJJPP6KT13py3QtCdMEGtOb3Q3PE

AI Use Disclosure: Code syntax was assisted by Codex model. All handwritten code and accompanying text were thoroughly reviewed and verified by me prior to submission. I retain full responsibility for the final contribution.


src/detector_error_models.jl
```diff
@@ -0,0 +1,413 @@
...
+
+A circuit operation representing an independent Stim detector error model mechanism.
+With probability `probability`, this operation XORs all `detector_bits` and `logical_bits`
+in a [`PauliFrame`](@ref).
+"""
+struct DetectorError{T,D,L} <: AbstractOperation
+    probability::T
+    detector_bits::NTuple{D,Int}
+    logical_bits::NTuple{L,Int}
+end
```
Krastanov: You will need to make it possible for this to "compactify" so that fast simulations can be done even when iterating over poorly typed lists of operations. I will comment further below.

src/detector_error_models.jl
```diff
@@ -0,0 +1,413 @@
...
+
+function DetectorError(probability::Real, detector_bits, logical_bits)
+    0 <= probability <= 1 || throw(ArgumentError("Detector error probability must be in [0, 1]."))
+    detectors = tuple(sort!(collect(Int, detector_bits))...)
+    logicals = tuple(sort!(collect(Int, logical_bits))...)
+    typed_probability = float(probability)
+    return DetectorError{typeof(typed_probability),length(detectors),length(logicals)}(
+        typed_probability, detectors, logicals
+    )
+end
```
Krastanov: Why do you need two separate constructors? Also, the top one probably should be an inner constructor.

src/detector_error_models.jl
```diff
@@ -0,0 +1,413 @@
...
+"""
+$(TYPEDEF)
+
+No-op operation that reserves detector and logical measurement columns for an imported
+detector error model.
+"""
+struct DetectorErrorModelDeclare <: AbstractOperation
+    detector_count::Int
+    logical_count::Int
+end
```
Krastanov: What is this used for? I do not see where it is necessary.

src/QuantumClifford.jl
```diff
@@ -78,6 +78,11 @@ export
     # Pauli frames
     PauliFrame, pftrajectories, pfmeasurements,
     measurements,
+    # Stim detector error models
+    DetectorError, DetectorErrorModelCircuit, DetectorErrorModelDeclare,
+    detector_count, logical_observable_count,
+    read_detector_error_model, detector_error_model_circuit,
+    detector_measurements, logical_observable_measurements,
```
Krastanov: not quite sure whether we should export all of these things

test/test_detector_error_models.jl
Krastanov, line 1: A pretty good correctness test here would be to compare your new implementation versus something like `naive_encoding_circuit` and `naive_syndrome_circuit` on a few small codes. Might need to wait on #751 first though

Krastanov (commented):
Looks like a good start, maybe a bit overengineered but knowing that would require a more in-depth review. Thank you for starting it, I left a few questions below.

src/detector_error_models.jl
Krastanov: You probably will need to look at this to make the compactification work https://github.com/QuantumSavory/QuantumClifford.jl/blob/f676df79b99da5eca8c9904d91e5d3713cea3259/src/sumtypes.jl#L219-L230

DennisWayo: Thanks for the notice Prof., its been fixed. `DetectorError` now participates in compactification through `concrete_typeparams`, and the detector model circuit keeps its metadata while storing compactified operations.

src/detector_error_models.jl
DennisWayo: Great catch. I folded the validation and canonical sorting into an inner constructor, and kept only one outer convenience constructor for generic iterables.

src/detector_error_models.jl
DennisWayo: Agreed, it was not necessary. I removed `DetectorErrorModelDeclare` and now use the detector model circuit metadata directly to allocate the measurement columns. I admire your thorough reviews.

src/QuantumClifford.jl
DennisWayo: Fair. I removed the unnecessary `DetectorErrorModelDeclare` export, and kept the importer plus the count/measurement helpers as the intended user-facing API for .dem files. I hope this satisfies.

test/test_detector_error_models.jl
DennisWayo: That makes lots of sense, Prof. I’ll leave this as a follow-up once #751 lands, since that gives us the right reference path for a stronger correctness test. I will be on standby.

PR #750, opened 2026-06-11
Added a dispatch based Noisify API ( For Issue #729)

### LINKED ISSUE #729: Add a dispatch-based API for turning noiseless circuits into noisy circuits

brohan-byte:
Key Notes:
- Added a CircuitNoise object in order to capture distinct noise models for single-qubit gates, two-qubit gates, idle qubits, measurements, and reset operations. This is used by the Noisify API later on.
- Added a NoNoise Sentinel to simplify the dispatch of omitted noise. 
- Added a time-step/layer based idle noise functionality based on how Quantikz.jl handles circuit scheduling and gate parallelism.
- Added a dispatch-based noisify API that supports both structured CircuitNoise models and individual AbstractNoise instances.
- Added tests for noisify (these are pretty hardcoded, any suggestions on improving them would be appreciated)

Most of the code was written by me but was passed through an LLM to restructure. In particular, the function apply_idle_noise was revised with LLM help. I've checked and verified the functionality of it, and I'll make sure that the logic is clearly documented

Fixes #729

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -74,7 +74,7 @@ pages = [
     "Perturbative Expansions" => "noisycircuits_perturb.md",
     "ECC example" => "ecc_example_sim.md",
     "Circuit Operations" => "noisycircuits_ops.md",
-    "Noisify" => "noisify.md"
+    "Constructing Noisy Circuits" => "noisify.md"
 ],
 "ECC compendium" => [
     "Evaluating codes and decoders" => "ECC_evaluating.md",
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -55,8 +55,8 @@ format = Documenter.HTML(
     assets = anythingllm_assets,
 ),
 modules = doc_modules,
-warnonly = [:missing_docs, :linkcheck],
-linkcheck = true,
+warnonly = [:missing_docs],
+linkcheck = false,
 authors = "Stefan Krastanov",
 pages = [
 "QuantumClifford.jl" => "index.md",
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -55,8 +55,8 @@ format = Documenter.HTML(
     assets = anythingllm_assets,
 ),
 modules = doc_modules,
-warnonly = [:missing_docs],
-linkcheck = false,
+warnonly = [:missing_docs, :linkcheck],
+linkcheck = true,
 authors = "Stefan Krastanov",
 pages = [
 "QuantumClifford.jl" => "index.md",
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -74,6 +74,7 @@ pages = [
     "Perturbative Expansions" => "noisycircuits_perturb.md",
     "ECC example" => "ecc_example_sim.md",
     "Circuit Operations" => "noisycircuits_ops.md",
+    "Noisify" => "noisify.md"
 ],
 "ECC compendium" => [
     "Evaluating codes and decoders" => "ECC_evaluating.md",
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -74,7 +74,7 @@ pages = [
     "Perturbative Expansions" => "noisycircuits_perturb.md",
     "ECC example" => "ecc_example_sim.md",
     "Circuit Operations" => "noisycircuits_ops.md",
-    "Constructing Noisy Circuits" => "noisify.md"
+    "Constructing Noisy Circuits" => "noisify.md",
 ],
 "ECC compendium" => [
     "Evaluating codes and decoders" => "ECC_evaluating.md",
--- a/docs/src/noisify.md
+++ b/docs/src/noisify.md
@@ -1,0 +1,79 @@
+## Noise Insertion
+
+```@meta
+DocTestSetup = quote
+    using QuantumClifford
+end
+```
+
+The `noisify` API converts an ideal circuit into a noisy circuit by inserting `NoiseOp`s at configurable locations. The original circuit remains unchanged and a new noisy circuit is returned.
+
+Using this functionality, you can apply:
+
+- A simple noise model, where the same noise is applied to all operations.
+- A structured `CircuitNoise` model, where different noise can be assigned to single-qubit gates, two-qubit gates, idle qubits, measurements, and reset operations.
+
+### Simple Noise Model
+
+```@example noisify
+using QuantumClifford
+state = one(Stabilizer, 1)
+
+circuit = [
+    sHadamard(1),
+    sCNOT(1,2),
+    sMZ(1,1),
+    Reset(state, [2])
+]
+
+noisy = noisify(circuit, PauliNoise(1e-3, 1e-3, 1e-3))
+```
+
+This inserts noise before single-qubit gates, two-qubit gates, measurements, and reset operations.
+
+### Structured Noise Model
+
+```@example noisify
+state = one(Stabilizer, 1)
+
+circuit = [
+    sHadamard(1),
+    sCNOT(1,2),
+    sMZ(1,1),
+    Reset(state, [2])
+]
+
+noise_model = CircuitNoise(
+    single_qubit = PauliNoise(1e-4, 1e-4, 1e-4),
+    two_qubit    = PauliNoise(1e-3, 1e-3, 1e-3),
+    idle_noise   = PauliNoise(1e-5, 1e-5, 1e-5),
+    measurement  = PauliNoise(2e-3, 2e-3, 2e-3),
+    reset        = PauliNoise(2e-3, 2e-3, 2e-3),
+)
+
+noisy_circuit = noisify(circuit, noise_model; nqubits=2)
+```
+
+To avoid double-noisification, `noisify` leaves existing `NoiseOp`s unchanged. It also does not modify `VerifyOp`s or classical operations such as `ClassicalXOR`.
+
+### Idle Noise
+
+When the `idle_noise` option is configured in `CircuitNoise`, noise is inserted on qubits that are not participating in any operation during a given circuit layer.
+
+Idle noise requires the total number of qubits in the circuit to be specified.
+
+```@example noisify
+noisy_circuit = noisify(circuit, noise_model; nqubits=2)
+```
+
+If idle noise is requested and `nqubits` is not provided, an error is thrown.
+
+### Simulation
+
+The resulting noisy circuits can be simulated directly using the existing trajectory simulators.
+
+```@example noisify
+register = Register(one(Stabilizer, 2), falses(1))
+
+mctrajectories(register, noisy_circuit; trajectories=100)
+```
--- a/docs/src/noisify.md
+++ b/docs/src/noisify.md
@@ -16,16 +16,14 @@ Using this functionality, you can apply:
 ### Simple Noise Model
 
 ```@example noisify
-using QuantumClifford
-state = one(Stabilizer, 1)
-
+reset_state = one(Stabilizer, 1)
 circuit = [
     sHadamard(1),
     sCNOT(1,2),
     sMZ(1,1),
-    Reset(state, [2])
+    Reset(reset_state, [2]),
+    VerifyOp(one(Stabilizer, 1), [2]),
 ]
-
 noisy = noisify(circuit, PauliNoise(1e-3, 1e-3, 1e-3))
 ```
 
@@ -34,13 +32,14 @@ This inserts noise before single-qubit gates, two-qubit gates, measurements, and
 ### Structured Noise Model
 
 ```@example noisify
-state = one(Stabilizer, 1)
+reset_state = one(Stabilizer, 1)
 
 circuit = [
     sHadamard(1),
     sCNOT(1,2),
     sMZ(1,1),
-    Reset(state, [2])
+    Reset(reset_state, [2]),
+    VerifyOp(one(Stabilizer, 1), [2]),
 ]
 
 noise_model = CircuitNoise(
@@ -70,10 +69,36 @@ If idle noise is requested and `nqubits` is not provided, an error is thrown.
 
 ### Simulation
 
-The resulting noisy circuits can be simulated directly using the existing trajectory simulators.
+The resulting noisy circuits can be simulated directly using the existing trajectory simulators such as `mctrajectories` and `pftrajectories`.
 
 ```@example noisify
-register = Register(one(Stabilizer, 2), falses(1))
+circuit = [
+    sHadamard(1),
+    sCNOT(1, 2),
+    sHadamard(2),
+    sCNOT(2, 1),
+    sMZ(1, 1),
+    sMZ(2, 2),
+]
+
+noise_model = CircuitNoise(
+    single_qubit = PauliNoise(1e-4, 1e-4, 1e-4),
+    two_qubit    = PauliNoise(1e-3, 1e-3, 1e-3),
+    idle_noise   = PauliNoise(1e-5, 1e-5, 1e-5),
+    measurement  = PauliNoise(2e-3, 2e-3, 2e-3),
+)
+
+noisy_circuit = noisify(circuit, noise_model; nqubits=2)
+
+register = Register(one(Stabilizer, 2), falses(2))
 
 mctrajectories(register, noisy_circuit; trajectories=100)
+```
+
+The same noisy circuit can also be simulated using pauli-frame trajectories.
+
+```@example noisify
+register = Register(one(Stabilizer, 2), falses(2))
+
+pftrajectories(register, noisy_circuit; trajectories=100)
 ```
--- a/docs/src/noisify.md
+++ b/docs/src/noisify.md
@@ -50,7 +50,7 @@ noise_model = CircuitNoise(
     reset        = PauliNoise(2e-3, 2e-3, 2e-3),
 )
 
-noisy_circuit = noisify(circuit, noise_model; nqubits=2)
+noisy_circuit = noisify(circuit, noise_model)
 ```
 
 To avoid double-noisification, `noisify` leaves existing `NoiseOp`s unchanged. It also does not modify `VerifyOp`s or classical operations such as `ClassicalXOR`.
@@ -62,10 +62,8 @@ When the `idle_noise` option is configured in `CircuitNoise`, noise is inserted 
 Idle noise requires the total number of qubits in the circuit to be specified.
 
 ```@example noisify
-noisy_circuit = noisify(circuit, noise_model; nqubits=2)
+noisy_circuit = noisify(circuit, noise_model)
 ```
-
-If idle noise is requested and `nqubits` is not provided, an error is thrown.
 
 ### Simulation
 
@@ -88,7 +86,7 @@ noise_model = CircuitNoise(
     measurement  = PauliNoise(2e-3, 2e-3, 2e-3),
 )
 
-noisy_circuit = noisify(circuit, noise_model; nqubits=2)
+noisy_circuit = noisify(circuit, noise_model)
 
 register = Register(one(Stabilizer, 2), falses(2))
 
--- a/docs/src/noisify.md
+++ b/docs/src/noisify.md
@@ -104,3 +104,5 @@ register = Register(state, falses(1))
 
 pftrajectories(register, pf_noisy_circuit; trajectories=100)
 ```
+
+All other simulator routines are also supported by this API
--- a/docs/src/noisify.md
+++ b/docs/src/noisify.md
@@ -6,17 +6,22 @@ DocTestSetup = quote
 end
 ```
 
-The `noisify` API converts an ideal circuit into a noisy circuit by inserting `NoiseOp`s at configurable locations. The original circuit remains unchanged and a new noisy circuit is returned.
+The [`noisify`](@ref) API converts an ideal circuit into a noisy circuit by inserting [`NoiseOp`](@ref)'s at configurable locations. The original circuit remains unchanged and a new noisy circuit is returned.
 
 Using this functionality, you can apply:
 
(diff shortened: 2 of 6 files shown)


docs/src/noisify.md
```diff
@@ -0,0 +1,79 @@
...
+To avoid double-noisification, `noisify` leaves existing `NoiseOp`s unchanged. It also does not modify `VerifyOp`s or classical operations such as `ClassicalXOR`.
+
+### Idle Noise
+
+When the `idle_noise` option is configured in `CircuitNoise`, noise is inserted on qubits that are not participating in any operation during a given circuit layer.
+
+Idle noise requires the total number of qubits in the circuit to be specified.
+
+```@example noisify
+noisy_circuit = noisify(circuit, noise_model; nqubits=2)
```
Krastanov: nqubits can be deduced by doing something like `maximum(affectedqubits.(circuit))`.

docs/src/noisify.md
```diff
@@ -0,0 +1,79 @@
...
+If idle noise is requested and `nqubits` is not provided, an error is thrown.
+
+### Simulation
+
+The resulting noisy circuits can be simulated directly using the existing trajectory simulators.
+
+```@example noisify
+register = Register(one(Stabilizer, 2), falses(1))
+
+mctrajectories(register, noisy_circuit; trajectories=100)
```
Krastanov: let's have two examples, one that uses `pftrajectories` as well

src/noisify.jl
```diff
@@ -0,0 +1,154 @@
...
+
+CircuitNoise(noise::AbstractNoise) = CircuitNoise(
+    single_qubit = noise,
+    two_qubit    = noise,
+    idle_noise   = noise,
+    measurement  = noise,
+    reset = noise
+)
+
+apply_idle_noise(circuit::AbstractVector, ::NoNoise, nqubits::Integer) = circuit
```
Krastanov: Can we rename it to `insert_...`?

src/noisify.jl
```diff
@@ -0,0 +1,154 @@
...
+# regular noise
+
+noisify(op::AbstractSingleQubitOperator, noise::AbstractNoise) = Any[NoiseOp(noise, affectedqubits(op)), op]
+noisify(op::AbstractTwoQubitOperator, noise::AbstractNoise) = Any[NoiseOp(noise, affectedqubits(op)), op]
+noisify(op::AbstractMeasurement, noise::AbstractNoise) = Any[NoiseOp(noise, affectedqubits(op)), op]
+noisify(op::Reset, noise::AbstractNoise) = Any[op, NoiseOp(noise, affectedqubits(op))]
+noisify(op::AbstractSingleQubitOperator, ::NoNoise) = Any[op]
+noisify(op::AbstractTwoQubitOperator, ::NoNoise) = Any[op]
+noisify(op::AbstractMeasurement, ::NoNoise) = Any[op]
+noisify(op::Reset, ::NoNoise) = Any[op]
```
Krastanov: can't nonoise just have a single method?

src/noisify.jl
```diff
@@ -0,0 +1,154 @@
...
+end
+
+noisify(op, ::CircuitNoise) = Any[op]
+noisify(op::AbstractNoiseOp, ::CircuitNoise) = Any[op]
+noisify(op::ClassicalXOR, ::CircuitNoise) = Any[op]
+noisify(op::VerifyOp, ::CircuitNoise) = Any[op]
+
+function noisify(op::AbstractSingleQubitOperator, noise_model::CircuitNoise)
+    out = Any[]
+    add_noise_op!(out, noise_model.single_qubit, affectedqubits(op))
```
Krastanov: Why can't this just be noisify(op, noise_model.single_qubit)?

src/noisify.jl
```diff
@@ -0,0 +1,154 @@
...
+    push!(out, op)
+    return out
+end
+
+function noisify(op::AbstractTwoQubitOperator, noise_model::CircuitNoise)
+    out = Any[]
+    add_noise_op!(out, noise_model.two_qubit, affectedqubits(op))
```
Krastanov: same question here

src/noisify.jl
```diff
@@ -0,0 +1,154 @@
...
+    push!(out, op)
+    return out
+end
+
+function noisify(op::AbstractMeasurement, noise_model::CircuitNoise)
+    out = Any[]
+    add_noise_op!(out, noise_model.measurement, affectedqubits(op))
```
Krastanov: same

src/noisify.jl
```diff
@@ -0,0 +1,154 @@
...
+    push!(out, op)
+    return out
+end
+
+function noisify(op::Reset, noise_model::CircuitNoise)
+    out = Any[]
+    push!(out, op)
+    add_noise_op!(out, noise_model.reset, affectedqubits(op))
```
Krastanov: same, and also just curious about the change of order

Krastanov (commented):
Looks overall good, need to do another pass of review to focus on a few things.

If you need to figure out type trees, use TypeTree.jl and do something like: `println(join(tt(supertype(some_starting_type))))`

docs/src/noisify.md
brohan-byte: This is helpful, but would this still work when the circuit does not touch all qubits in the provided stabilizer state? In that case, inferring nqubits from the circuit would undercount the register size, so idle noise would not be applied to qubits outside the maximum qubit index appearing in the circuit.

Krastanov: I think that situation would be awkward to arise. Let's just warn in the docstring of idle noise / circuit noise that this is the case.

src/noisify.jl
brohan-byte: I had to introduce separate dispatch methods for NoNoise because i defined NoNoise as a subtype of AbstractNoise. Since noisify already has methods specialized on AbstractNoise, those methods would be selected over the generic NoNoise handler unless explicit NoNoise dispatches were provided for the affected operation types.

Krastanov: ah, I guess this is something about method resolution ambiguities. Sounds good

docs/make.jl
```diff
@@ -74,6 +74,7 @@ pages = [
     "Perturbative Expansions" => "noisycircuits_perturb.md",
     "ECC example" => "ecc_example_sim.md",
     "Circuit Operations" => "noisycircuits_ops.md",
+    "Noisify" => "noisify.md"
```
Krastanov: this probably needs a better name -- seeing this in the tables of content would not mean much to a reader

src/noisify.jl
```diff
@@ -0,0 +1,117 @@
...
+insert_idle_noise(circuit::AbstractVector, ::NoNoise) = circuit
+
+function insert_idle_noise(circuit::AbstractVector, idle_noise::AbstractNoise)
+    isempty(circuit) && return Any[]
+    nqubits = maximum(q for op in circuit for q in affectedqubits(op))
+    filled_up_to = fill(1, nqubits)
+    layers = Dict{Int, Vector{Any}}()
+    active_qubits = Dict{Int, Set{Int}}()
+    for op in circuit
+        if op isa AbstractNoiseOp || op isa VerifyOp || op isa ClassicalXOR
```
Krastanov: Why is AbstractNoiseOp here?

More importantly, this seems to be not extendable. If a user or an external library creates a type that requires this same type of special treatment, they will not be able to use this noisify method.

Could you explain, maybe in a few drawings, why is this special cased in this fashion?

src/noisify.jl
```diff
@@ -0,0 +1,117 @@
...
+)
+
+insert_idle_noise(circuit::AbstractVector, ::NoNoise) = circuit
+
+function insert_idle_noise(circuit::AbstractVector, idle_noise::AbstractNoise)
+    isempty(circuit) && return Any[]
+    nqubits = maximum(q for op in circuit for q in affectedqubits(op))
+    filled_up_to = fill(1, nqubits)
+    layers = Dict{Int, Vector{Any}}()
+    active_qubits = Dict{Int, Set{Int}}()
```
Krastanov: What is the meaning of these two? Why are they needed in addition to `filled_up_to`?

brohan-byte: In the `insert_idle_noise function`, I use these `layers` and `active qubits `in line 64 in order to compute idle qubits for each time step layer ( this function makes the assumption that each gate only takes up one step in time, let me know if this is inaccurate). During the initial pass through the circuit, `filled_up_to` tracks the next available layer for each qubit, allowing operations to be assigned to their respective layers. `layers` records which operations occur in each layer so that the circuit can be reconstructed after idle-noise insertion. `active_qubits` records which qubits participate in operations within each layer. While active_qubits is not strictly necessary (it could be recomputed from layers), storing it avoids additional work during the reconstruction pass.

src/noisify.jl
brohan-byte: Initially I special cased AbstractNoiseOp because I didn't want it to undergo idle noise (I'm still not sure if it is supposed to, do let me know). But I realize now that it can still follow the layering logic like a regular operation using affectedqubits, I'll make that change.

The part I'm still unsure about is how to handle ops such as ClassicalXOR or VerifyOp. The idle-noise insertion pass first schedules the circuit into timesteps/layers using the `filled_up_to vector`, so the resulting circuit is effectively reordered according to layer structure rather than preserving the original linear ordering. For ordinary operations this is fine. However, it’s not clear to me how VerifyOp and ClassicalXOR should interact with this scheduling pass. However, for VerifyOp and ClassicalXOR, I’m not sure what the correct scheduling semantics should be, or even how their timestep should be determined. Should they participate in the layer assignment process somehow, or should they instead preserve their position in the original circuit?

brohan-byte: The dilemma is as follows:

Suppose our circuit was: `[H(1), ClassicalXOR(1), H(2)]`

The current layering logic would have re order this according to the time steps as:
t=1: H(1), H(2)
t=2:ClassicalXOR(...)

In other words, it doesn't take place in any of the layering logic and is just placed in the current maximum timestep. 

Another possible model could be that the circuit should be re-ordered as:
t=1: H(1)
t=2: ClassicalXOR(...)
t=3:H(2)

and similarly for VerifyOp.
The issue is that, unlike ordinary quantum operations, how should I assign a timestep to ClassicalXOR and VerifyOp and other possible ops in the future. As a result, I’m unsure whether they should participate in the layering process at all, or whether they should instead preserve their position in the original circuit.

Let me know if I was able to get my thoughts across clearly!

docs/src/noisify.md
```diff
@@ -0,0 +1,102 @@
...
+```@example noisify
+reset_state = one(Stabilizer, 1)
+circuit = [
+    sHadamard(1),
+    sCNOT(1,2),
+    sMZ(1,1),
+    Reset(reset_state, [2]),
+    VerifyOp(one(Stabilizer, 1), [2]),
+]
+noisy = noisify(circuit, PauliNoise(1e-3, 1e-3, 1e-3))
```
Krastanov, line 33: when the doc ci is fixed, I want to look at how this renders in the docs

docs/src/noisify.md
```diff
@@ -0,0 +1,102 @@
...
+noisy_circuit = noisify(circuit, noise_model)
+```
+
+To avoid double-noisification, `noisify` leaves existing `NoiseOp`s unchanged. It also does not modify `VerifyOp`s or classical operations such as `ClassicalXOR`.
+
+### Idle Noise
+
+When the `idle_noise` option is configured in `CircuitNoise`, noise is inserted on qubits that are not participating in any operation during a given circuit layer.
+
+Idle noise requires the total number of qubits in the circuit to be specified.
```
Krastanov: not true anymore?

docs/src/noisify.md
```diff
@@ -0,0 +1,102 @@
...
+
+register = Register(one(Stabilizer, 2), falses(2))
+
+mctrajectories(register, noisy_circuit; trajectories=100)
+```
+
+The same noisy circuit can also be simulated using pauli-frame trajectories.
+
+```@example noisify
+register = Register(one(Stabilizer, 2), falses(2))
```
Krastanov: pftrajectories does not take a register -- you can probably just directly give it the circuit

src/noisify.jl
```diff
@@ -0,0 +1,117 @@
...
+    return output
+end
+
+
+
+noisify(circuit::AbstractVector, noise::AbstractNoise) = reduce(vcat, noisify.(circuit, (noise,)))
+noisify(op, noise::AbstractNoise) = Any[op]
+noisify(op::AbstractNoiseOp, noise::AbstractNoise) = Any[op]
+noisify(op::ClassicalXOR, noise::AbstractNoise) = Any[op]
+noisify(op::VerifyOp, noise::AbstractNoise) = Any[op]
```
Krastanov: these do not seem to be used, unless there is some method ambiguity -- if there is method ambiguity, leave a comment mentioning ambiguity with which other method

Krastanov: a simpler solution to fixing all the ambiguity is to NOT use `NoNoise <: AbstractNoise`, but rather to just use `Nothing`

src/noisify.jl
```diff
@@ -0,0 +1,117 @@
...
+noisify(op::Reset, noise::AbstractNoise) = Any[op, NoiseOp(noise, affectedqubits(op))]
+
+noisify(op::AbstractSingleQubitOperator, ::NoNoise) = Any[op]
+noisify(op::AbstractTwoQubitOperator, ::NoNoise) = Any[op]
+noisify(op::AbstractMeasurement, ::NoNoise) = Any[op]
+noisify(op::Reset, ::NoNoise) = Any[op]
+
+
+function noisify(circuit::AbstractVector, noise_model::CircuitNoise)
+    if noise_model.idle_noise isa NoNoise
```
Krastanov: generally never use `isa` in Julia -- it makes things unextendable. It is there for the rare need for introspection.

Krastanov: thanks to line 29, you can just skip this

src/noisify.jl
```diff
@@ -0,0 +1,117 @@
...
+    single_qubit = noise,
+    two_qubit    = noise,
+    idle_noise   = noise,
+    measurement  = noise,
+    reset = noise
+)
+
+insert_idle_noise(circuit::AbstractVector, ::NoNoise) = circuit
+
+function insert_idle_noise(circuit::AbstractVector, idle_noise::AbstractNoise)
```
Krastanov, line 49:
```
adding CNOT 2 3 to a circuit for which filled_up_to looks like this:
* * * * *
*
* * *

would result into a:

filled_up_to is modified to
* * * * *
* /  /  * (where / denotes "skip")
* * * *
idling noise will be added to each skip with push!(noisy_circuit, idlenoiseop)
push!(noisy_circuit, sCNOT(2,3))
```

And then for some specific operations, it just directly goes to push! without looking at filled_up_to. Instead of `isa` though, lets use a private method `skip_idling_noise(_) = false` and `skip_idling_noise(::ParticularType) = true`. Keep the method private but documented.

Krastanov (changes requested):
looks good, some very minor things to change, and fixing the doc build

I want to see the docs before merging.

docs/src/noisify.md
```diff
@@ -0,0 +1,106 @@
...
+    sCNOT(1,2),
+    sMZ(1,1),
+]
+
+pf_noisy_circuit = noisify(pf_circuit, noise_model)
+
+register = Register(state, falses(1))
+
+pftrajectories(register, pf_noisy_circuit; trajectories=100)
+```
```
Krastanov, line 106: and of course by all other simulator routines

src/noisify.jl
```diff
@@ -0,0 +1,106 @@
...
+            end
+
+            push!(emitted, step)
+        end
+
+        push!(output, op)
+    end
+
+    output
+end
```
Krastanov, line 87: this can be simplified further as we discussed on the call

Krastanov (changes requested):
besides the minor comments, this would be ready to merge after they are addressed

Also, if you add this line to the doc builder, we will be able to see the docs inside of this PR https://github.com/QuantumSavory/QuantumSavory.jl/blob/6263ed03f0be553d5d75d26562563be513c41036/docs/make.jl#L109

Things to check before merging:

- a mention that all other simulator routines (no need to enumerate) are supported
- simplifying the idle noise addition
- for CircuitNoise add a per-field docstring and then use docstringhelpers to automatically populate the maindocstring (check out how it is done in QuantumSavory.ProtocolZoo)
- link the related Quantikz.jl issue about the broken circuit diagram

Krastanov:
Add a changelog entry and feel free to merge (do ping on zulip me if there are unrelated failing tests)

Krastanov:
if you want to have a release after this is merged, in this PR bump the version 0.11.x+1 and update the changelog to say the version instead of saying "unreleased"

src/QuantumClifford.jl
```diff
@@ -78,6 +78,8 @@ export
     # Pauli frames
     PauliFrame, pftrajectories, pfmeasurements,
     measurements,
+    #noisify
+    noisify, CircuitNoise,skip_idling_noise, append_idle_noise!,
```
Krastanov: remove the private ones

Krastanov (changes requested):
Feel free to merge after the private symbols are removed from the export, as long as tests pass (if there are unrelated failures, just explain them in a comment). Make sure to use squash+merge instead of the other merge options.

PR #751, opened 2026-06-12
Export Stim detector error models from ECC codes

### LINKED ISSUE #726: Export Stim detector error models from QuantumClifford codes

---

QuantumClifford already has a strong error-correction layer: code definitions,
parity checks, syndrome circuits, and the `faults_matrix` helper that maps
physical Pauli faults to logical faults. Stim uses detector error model (`.dem`)
files as a compact interchange format for syndrome-detector graphs and logical
observable flips. Exporting `.dem` files from QuantumClifford codes would make it
much easier to compare QuantumClifford simulations with Stim-based workflows and
external decoders.

Relevant links:

- ECC source: https://github.com/QuantumSavory/QuantumClifford.jl/blob/main/src/ecc/ECC.jl
- ECC circuit helpers: https://github.com/QuantumSavory/QuantumClifford.jl/blob/main/src/ecc/circuits.jl
- Existing Pauli-frame simulation docs: https://qc.quantumsavory.org/stable/API/#Circuit-simulation

A user should be able to create a detector error model from one of the ECC code
types and write it to a `.dem` file that Stim can parse:

```julia
using QuantumClifford
using QuantumClifford.ECC

dem = detector_error_model(Steane7(); px=1e-3, py=0.0, pz=1e-3)
write_detector_error_model("steane7.dem", dem)
```

The exact API can be adjusted, but it should support both returning an in-memory
model and writing directly to an `IO` or filename:

```julia
detector_error_model(code; px, py, pz)
write_detector_error_model(io, dem)
write_detector_error_model(path::AbstractString, dem)
```

## Suggested implementation

Start with a code-capacity detector error model generated from the code's parity
checks:

1. Use `parity_checks(code)` as the detector definitions.
2. For each physical qubit and each enabled Pauli fault (`X`, `Y`, `Z`), compute
   the stabilizer checks flipped by that fault.
3. Use `faults_matrix(code)` to compute which logical observables are flipped by
   the same fault.
4. Emit one `error(p) D... L...` line per nonzero physical fault probability.
5. Emit deterministic detector and logical-observable declarations so output is
   stable across runs and easy to test.

For example, a single-qubit Pauli error that flips checks 1 and 4 and logical
observable 0 should produce a line equivalent to:

```text
error(0.001) D0 D3 L0
```

The exact detector numbering can follow QuantumClifford's internal stabilizer
row order. Please document the convention.

## In scope

- A small internal representation for detector error models, or a direct writer
  if that is simpler.
- Support for independent `X`, `Y`, and `Z` physical fault probabilities.
- Export for at least the built-in ECC codes where `parity_checks` and
  `faults_matrix` are available.
- Deterministic `.dem` text output.
- Unit tests for a few small codes.
- A smoke test that verifies the emitted `.dem` file can be parsed by Stim when
  Stim is available in the test environment. This can be optional or guarded so
  the Julia test suite does not require Python by default.
- Documentation with a short example.

## Out of scope

- Full circuit-to-DEM extraction for noisy syndrome extraction circuits.
- Modeling correlated gate faults from a full hardware schedule.
- Coordinates, repeated rounds, or `repeat` blocks, unless they fall out
  naturally from the implementation.
- Building or integrating a decoder.

## Acceptance criteria

- `detector_error_model(Steane7(); px=..., py=..., pz=...)` or an equivalent API
  exists and is exported or documented.
- The generated `.dem` files follow Stim's detector error model syntax.
- Tests check that detector targets match the stabilizers flipped by the
  corresponding Pauli error.
- Tests check that logical targets match `faults_matrix(code)`.
- The output ordering is deterministic.
- Docs show how to generate a `.dem` file and parse it with Stim.

CrewRiz:
Fixes #726.

## Summary

This PR adds a small code-capacity Stim detector error model exporter for ECC codes.

- Adds `detector_error_model(code; px, py, pz)` returning deterministic `.dem` text.
- Uses `parity_checks(code)` for detector targets and `faults_matrix(code)` for logical observable targets.
- Adds `write_detector_error_model(io, dem_text)` and `write_detector_error_model(path, dem_text)` for writing the generated text to an `IO` object or file.
- Documents the detector/logical target numbering convention and includes a short Stim parsing example.
- Adds focused tests for detector targets, logical targets, deterministic output, file writing, probability validation, and zero-probability omission.

## Numbering convention

Detector targets are zero-based and follow the `parity_checks(code)` stabilizer row order: stabilizer row `i` maps to `D(i - 1)`.

Logical observable targets are zero-based and follow `faults_matrix(code)` row order: logical-X observables first, then logical-Z observables. Physical `Y` faults use the xor of the corresponding physical `X` and `Z` fault-matrix columns.

## Review response

The exporter now writes the DEM text directly instead of building custom `DetectorErrorModel`/`DetectorErrorModelError` structs. The probability validation was simplified to one direct range check, and the optional Python/Stim test was removed from the Julia test item to avoid adding noisy environment-dependent test output.

## Validation

Focused checks run locally:

```text
julia --project=test -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd())); Pkg.instantiate(); using TestItemRunner; @run_package_tests filter=ti -> ti.name == "ECC detector error model export"'
# 17 passed

git diff --check
# passed

julia --project=test -e 'using QuantumClifford, QuantumClifford.ECC; path = "/tmp/quantumclifford-steane7.dem"; write_detector_error_model(path, detector_error_model(Steane7(); px=1e-3, pz=1e-3)); println(path)'
test/.CondaPkg/.pixi/envs/default/bin/python -c 'import pathlib, stim; stim.DetectorErrorModel(pathlib.Path("/tmp/quantumclifford-steane7.dem").read_text()); print("stim parse ok")'
# stim parse ok
```

I also ran the broader `:ecc_base` TestItemRunner slice. The new DEM item passed in that batch, but the batch still ends with an existing unrelated `test/test_cecc.jl` / Hecke error at `test/test_cecc_base.jl:45`:

```text
1045 passed, 0 failed, 1 errored
QuantumClifford.jl/test/test_ecc_detector_error_model.jl | 17 passed
```

## Checklist

- [x] The code is properly formatted and commented where useful.
- [x] Substantial new functionality is documented.
- [x] New functionality is tested.
- [ ] The full automated test suite passes on GitHub Actions.
- [x] Formatting checks pass for the changed diff.
- [x] I have read and followed the bug bounty AI usage and disclosure policy.

## AI usage disclosure

Travis Crew (CrewRiz) used AI-assisted coding support while preparing this PR. I reviewed the implementation, tests, documentation, and local verification output, and I take responsibility for the final contribution.


test/test_ecc_detector_error_model.jl
```diff
@@ -0,0 +1,85 @@
...
+                `$python -c "import importlib.util; print(importlib.util.find_spec('stim') is not None)"`,
+            )
+            if has_stim == "True"
+                @test success(
+                    `$python -c "import pathlib, stim, sys; stim.DetectorErrorModel(pathlib.Path(sys.argv[1]).read_text())" $path`,
+                )
+            else
+                @info "Skipping Stim parse smoke test because Python package stim is not installed."
+            end
+        end
```
Krastanov: This is a really awkward test that will probably never actually run. Moreover, it adds noise to the testrunner -- if we start putting info messages like this, it will quickly become unmanageable, and this particular test does not seem meaningful enough to warrant an exception.

test/test_ecc_detector_error_model.jl
```diff
@@ -0,0 +1,85 @@
...
+    end
+
+    @testset "probabilities are validated and zero probabilities are omitted" begin
+        @test isempty(detector_error_model(Steane7(); px=0, py=0, pz=0).errors)
+        @test_throws ArgumentError detector_error_model(Steane7(); px=-0.1)
+        @test_throws ArgumentError detector_error_model(Steane7(); py=1.1)
```
Krastanov: This type of `@test_throws` is quite dangerous. You do not know what ArgumentError is being detected here, so this test might actually HIDE a real problem. `@test_throws` can test for specific substring in the error message -- that would be much safer.

src/ecc/detector_error_model.jl
```diff
@@ -0,0 +1,137 @@
...
+[`faults_matrix`](@ref), declared as `L` targets.
+"""
+struct DetectorErrorModel
+    num_detectors::Int
+    num_observables::Int
+    errors::Vector{DetectorErrorModelError}
+end
+
+function _dem_probability(p, name::Symbol)
+    p isa Real || throw(ArgumentError("`$name` must be a real probability."))
```
Krastanov: This is not a very idiomatic check in Julia. A type might have a well-defined real-like API without subtyping Real. Moreover, this type of stuff should generally be left to be a MethodError as it semantically means "that type is not supported" -- doing this type of overwrite makes things just more complicated.

Overall this particular function is pretty overengineered and it does not seem necessary. Just a quick check that all three values are between 0 and 1 would have been enough. As a rule, avoid doing fancy abstractions when the obvious code is simpler and shorter.

src/ecc/detector_error_model.jl
```diff
@@ -0,0 +1,137 @@
...
+
+`num_detectors` is the number of stabilizer checks declared as `D` targets.
+`num_observables` is the number of logical-observable rows from
+[`faults_matrix`](@ref), declared as `L` targets.
+"""
+struct DetectorErrorModel
+    num_detectors::Int
+    num_observables::Int
+    errors::Vector{DetectorErrorModelError}
+end
```
Krastanov: Why are these abstractions needed? Why not just directly write a the string that will be contained in the DEM file? What value is provided by having this abstraction / indirection?

Krastanov (changes requested):
Thank you for starting this! There are a few pieces that seem a bit overengineered when a simpler piece of code would work just as well -- please check out the comments below.

DennisWayo:
Hi @CrewRiz, Since my importer PR #747 would benefit from your landing, I’m glad to help push this over the line.

Here; I could not push directly to this branch, so I pushed a small support patch to my fork here: https://github.com/DennisWayo/QuantumClifford.jl/tree/assist-pr-751-dem-export

It applies on top of this PR and follows the review direction: direct DEM text writing, simpler probability checks, no optional Stim smoke test, message-specific throws checks, and a small fix to use `comm(tab(checks), ...)`. 

Cherry-pick commit `04916757` if useful.

test/test_ecc_detector_error_model.jl
CrewRiz: Yeah, that one snuck through, thats my bad, i'll remove it with my next push.

PR #787, opened 2026-08-20
Remove closure boxes from core and extensions

Krastanov-agent:
## Summary

- isolate reassigned locals with `let` bindings in the sum-type generator, Hecke circulant conversions, and Oscar homological-product conversion
- move the `Test.detect_closure_boxes` regression checks into `test/test_closure_boxes.jl`
- run those checks on Julia 1.14 final and later with `@static if VERSION >= v"1.14"`
- keep the closure checks out of the Aqua-tagged test group while preserving `:oscar_required` on the Oscar extension check

## Validation

- Julia 1.14.0-DEV.3007: direct scan of QuantumClifford and the Hecke and Oscar extensions reports no boxes
- Julia 1.14.0-DEV.3007: standalone closure-box file discovers both items and executes 0 assertions because the final-release gate is false
- Julia 1.12.6: standalone closure-box file discovers both items and executes 0 gated assertions
- Julia 1.12.6: Aqua passed 11/11
- Julia 1.12.6: representative sum-type, Hecke, and Oscar behavior checks passed
- Julia 1.12.6: full package test entry point passed with 416068 passes, 9 expected broken tests, and 416077 total results

No changelog entry is included because this changes compiler lowering and allocations without changing public behavior.

### DIFF
--- a/ext/QuantumCliffordHeckeExt/generalized_hypergraph_product.jl
+++ b/ext/QuantumCliffordHeckeExt/generalized_hypergraph_product.jl
@@ -116,7 +116,9 @@ function _polynomial_matrix_to_circulant_matrix(H_poly, l)
         H_block = _circulant_matrix(poly, l)
         H[(i-1)*l+1:i*l, (j-1)*l+1:j*l] = H_block
     end
-    H = [Int(lift(ZZ, H[i,j])) for i in 1:nrows(H), j in 1:ncols(H)]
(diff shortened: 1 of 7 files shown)


Krastanov-agent:
Maintainer action: please apply the Skip-Changelog label. This PR changes compiler lowering and allocations without changing public behavior, and the fork account cannot add labels on the upstream repository.

PR #796, opened 2026-08-22
Add correct and CI-friendly Tile2D codes

Krastanov-agent:
Supersedes #708 while preserving its original commits and authorship. This branch starts from the full #708 head and merges the current upstream `master` before applying the review fixes.

## Correctness

- retain only qubits supported by both the X- and Z-check families, retruncate every check, and then remove empty checks
- validate every tile coordinate against the zero-based `B × B` box
- support valid tiles with different horizontal and vertical support cardinalities
- return sparse Boolean parity matrices, matching the `AbstractCSSCode` contract and decoder expectations
- make layout and qubit ordering deterministic

The pruning regression uses a case that previously produced 21×46 matrices with one-sided columns. It now produces commuting 15×34 X/Z matrices with `k = 4` and no unsupported columns. Tests also cover an asymmetric tile and an exact `Lx != Ly` layout.

## API and consistency

- export `Tile2D` from `QuantumClifford.ECC`, with a public-import regression test
- document coordinate/edge conventions, automatic complementary Z-tile construction, and the supported open rectangular layout
- use scoped helper names and idiomatic local/test naming
- bump QECCore to 0.1.4 and use the `Tile2D` name consistently in both changelogs

## Tests and CI cost

- replace six 288–512-qubit generic fixtures with one 8-qubit non-square representative
- remove the duplicated Appendix B fixture
- remove the 288-qubit exact-distance MIP from required tests and doctests while retaining paper-table n/k, rank, commutation, and check-weight coverage

That MIP previously ran in the root doctest, QECCore test, and QECCore doctest paths, adding roughly 26–28 minutes each. The MIP implementation remains covered by the library's dedicated distance tests; Tile2D is tested through its standard CSS parity-matrix contract.

Local focused results on Julia 1.12:

- QECCore `Tile2D` test item: 107/107 passed
- `QuantumClifford.ECC` public API test: 2/2 passed
- QECCore doctests: passed
- `git diff --check upstream/master`: passed

## Release sequencing

As requested in the review on #708, this first-stage PR bumps `lib/QECCore/Project.toml` to 0.1.4 but leaves QuantumClifford's current lower bound at 0.1.3. After QECCore 0.1.4 is registered, a follow-up must raise the root compatibility floor to 0.1.4 before any QuantumClifford release.


Krastanov-agent:
Release coordination: #799 also prepares QECCore 0.1.4 and must register that version first so the BCH namespace move can be released before QuantumClifford raises its QECCore compatibility floor. Please rebase this PR after #799 merges and prepare Tile2D as QECCore 0.1.5; registering two different QECCore trees as 0.1.4 would make one change unreleasable.

PR #800, opened 2026-08-22
Normalize Buildkite job labels

Krastanov-agent:
## Summary

- apply one target-first naming scheme to every Buildkite matrix: `<tested package> - <suite or dependency mode> (local dependency)`
- normalize `base tests` to `General` and the three GPU shard names
- place distinguishing target and shard text before Buildkite truncation

Examples:

- `QuantumClifford - ECC Base (local QECCore)`
- `QECCore - JET (local QuantumClifford)`
- `QuantumSavory - Release (local QuantumClifford)`
- `QuantumSavory - Development (local QuantumClifford)`

## Verification

- parsed `.buildkite/pipeline.yml` with PyYAML
- expanded all 19 matrix jobs
- verified all 19 full labels and truncated GitHub context slugs are unique
- `git diff --check`

Only labels changed; commands, selectors, queues, plugins, and package lists are unchanged.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -5,7 +5,7 @@ secrets:
   - CODECOV_TOKEN
 
 steps:
-  - label: "QuantumClifford Tests (with local QECCore) - {{matrix.LABEL}}"
+  - label: "QuantumClifford (local QECCore) - {{matrix.LABEL}}"
     plugins:
       - JuliaCI/julia#v1:
           version: "1"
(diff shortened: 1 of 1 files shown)


Krastanov-agent:
Closing as requested. The context collision is real, but this naming-only change is not required for the restored test fixes.

Krastanov:
it seems that the long buildkite labels were leading to some part of the pipeline to just truncate the slug, making it difficult to distinguish which CI group was being reported

PR #801, opened 2026-08-22
Fix restored ECC Base shard

Krastanov-agent:
## Summary

- repair the stale `LiftedCode` fixture exposed by the restored ECC Base shard
- preserve default and custom representations for concrete group-algebra matrices
- return machine-integer lifted-code dimensions and remove the obsolete typed Gallager fixture

## Verification

- ECC Base shard: 1,603/1,603 assertions passed
- direct default, custom, shift, and GF(3) representation probes passed
- `git diff --check`

This fixes the ECC Base failure reported by Buildkite build 3614.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## unreleased
 
+- **(fix)** Correct `LiftedCode` construction and dimensions for concrete GF(2) group-algebra matrices.
 - Allow Nemo 0.56 in QuantumClifford and QECCore, and update Oscar quotient-ring examples for Nemo's new finite-field representation.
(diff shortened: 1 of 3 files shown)


Krastanov-agent:
Updated as requested: the inner `GroupAlgebraElemMatrix` constructor now owns the default `representation_matrix` and GF(2) guard, and the overlapping outer constructor (and `invoke`) is gone. Two independent dispatch reviews agreed on this shape. Default/custom GF(2), custom/default GF(3), LP-code, parity-matrix, and method-ambiguity probes pass; the full ECC Base shard passes 1,603/1,603.

PR #807, opened 2026-08-23
[unvetted AI slop] Remove table-decoder lookup races

Krastanov-agent:
## Summary\n\n- use a per-call syndrome lookup key instead of each table decoder's shared scratch key\n- add concurrent-reuse regressions for quantum and classical table decoders\n\n## Testing\n\nNot run locally; relying entirely on repository CI as requested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## unreleased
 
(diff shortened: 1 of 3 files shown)


Krastanov:
Nope, this is going to add a ton of allocations. This should use thread-local or task-local storage instead.

PR #808, opened 2026-08-23
[unvetted AI slop] Isolate dense Clifford scratch state

Krastanov-agent:
## Summary\n\n- allocate dense Clifford scratch storage per application instead of mutating shared gate state\n- add concurrent full and indexed application regressions\n\n## Testing\n\nNot run locally; relying entirely on repository CI as requested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## unreleased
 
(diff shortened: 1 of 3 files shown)


Krastanov:
Nope, this is going to add a ton of allocations. This should use thread-local or task-local storage instead.

PR #809, opened 2026-08-23
Honor RNGs in all-to-all circuit sampling

Krastanov-agent:
Summary
- route every all-to-all destination draw through the caller-supplied RNG
- add a regression that varies global RNG state while keeping the explicit RNG fixed


Krastanov:
oh, good catch, bot

PR #813, opened 2026-08-23
[unvetted AI slop] Embed supplied LiftedCode group elements

Krastanov-agent:
## Summary\n\n- embed each supplied group element instead of indexing from the uninitialized output matrix\n- add a focused constructor regression\n\n## Testing\n\nNot run locally; relying entirely on repository CI as requested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## unreleased
 
+- **(fix)** Embed the supplied group elements when constructing a `LiftedCode`.
 - **(fix)** Correct `LiftedCode` construction and dimensions for concrete GF(2) group-algebra matrices.
 - Avoid runtime dispatch when resetting qubits in `MixedDestabilizer` states.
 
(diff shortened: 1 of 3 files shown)


Krastanov-agent:
### Why this fix is necessary

This is a direct input-loss bug in the constructor for raw Hecke group elements:

```julia
LiftedCode(group_elem_array::Matrix{<:GroupOrAdditiveGroupElem}; ...)
```

Before this PR, the method allocated `A` as a matrix of `zero(GA)` and then assigned:

```julia
A[i, j] = GA[A[i, j]]
```

The right-hand side reads the newly allocated output matrix, not `group_elem_array`. Consequently, the supplied group element is never read. A nonidentity input can therefore be replaced by the initial zero value, or the conversion can fail through the wrong Hecke indexing path. The intended operation is to embed the actual input group element as a basis element of the group algebra, which is exactly what `GA(group_elem_array[i, j])` does.

This overload is also used by `LPCode(::Matrix{<:GroupOrAdditiveGroupElem}, ...)`, so the defect affects both direct `LiftedCode` construction and lifted-product construction from raw group-element matrices. It can change the resulting parity matrix and therefore the code itself.

The existing lifted-code test covered two different overloads:

1. `B` was already a matrix of group-algebra elements, so it dispatched directly to `LiftedCode(::GroupAlgebraElemMatrix)`.
2. `base_matrix` was a matrix of integers, so it dispatched to the cyclic-shift constructor `LiftedCode(::Matrix{Int}, l)`.

The larger LP fixtures likewise construct matrices whose entries are already elements of `GA`. None of those cases dispatches to the broken `Matrix{<:GroupOrAdditiveGroupElem}` method. The existing assertions also checked that downstream operations completed, but did not check that a raw input element was preserved during embedding.

The new regression supplies an actual group generator from `gens(G)` and verifies that the stored entry is `GA(group_generator)`. That directly exercises the previously uncovered overload and checks its essential semantic contract.

PR #814, opened 2026-08-23
[unvetted AI slop] Use the Goppa polynomial base field

Krastanov-agent:
## Summary\n\n- build the default support from the Goppa polynomial's actual base field\n- add a regression using a noncanonical finite-field representation\n\n## Testing\n\nNot run locally; relying entirely on repository CI as requested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## unreleased
 
+- **(fix)** Construct default Goppa support sets in the polynomial's base field.
 - **(fix)** Correct `LiftedCode` construction and dimensions for concrete GF(2) group-algebra matrices.
 - Avoid runtime dispatch when resetting qubits in `MixedDestabilizer` states.
 
--- a/lib/QECCore/CHANGELOG.md
+++ b/lib/QECCore/CHANGELOG.md
@@ -1,5 +1,9 @@
 # News
 
+## Unreleased
+
+- Construct default Goppa support sets in the polynomial's existing base field.
+
 ## v0.1.4 - 2026-08-23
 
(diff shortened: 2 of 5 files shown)


Krastanov-agent:
### Why this fix is necessary

For a binary Goppa code, the polynomial and support must belong to the same concrete field:

```text
g ∈ F[x]
L ⊆ F
```

The previous three-argument constructor discarded the field attached to `g` and independently created `GF(2, m, :α)`. Although every field of size `2^m` is abstractly isomorphic, Nemo field elements carry a specific parent and a specific polynomial representation. The integer `m` alone does not identify that parent or provide an isomorphism from a newly created field to the field of `g`.

As a result, the old constructor could create `L` in a different parent from the coefficients of `g`. That violates the constructor's mathematical contract and later reaches operations such as `evaluate(g, o_i)` in `parity_matrix`. Depending on the concrete field presentations, this can cause incompatible-parent errors or make the support use a representation unrelated to the polynomial's field.

The authoritative field is already available as `base_ring(parent(g))`. Enumerating that field guarantees that every support element can be evaluated by `g` without an implicit or guessed field conversion.

Most Goppa coverage bypassed this code path:

- Random-code generation constructs `g` and `L` together and calls the four-argument constructor.
- The reference examples with an explicit `L` also build the support from the same `F` as `g`, then call the four-argument constructor.

The two existing examples that did use the default-support constructor created their polynomial field with the same `(2, m, :α)` presentation that the implementation hard-coded. That special case masks the unsupported assumption. Those tests checked code length, rank, and parity-matrix properties, but never asserted the parent invariant

```julia
parent(a) === base_ring(parent(g))
```

for the generated support.

The regression in this PR checks that invariant directly. This matters even when two field presentations happen to produce the same numerical code parameters: the returned `Goppa` object must remain internally coherent and valid for subsequent Nemo operations.

PR #816, opened 2026-08-23
[unvetted AI slop] Select MIP summaries by distance

Krastanov-agent:
## Summary

- compare minXZ optimization summaries by their distance value
- select the per-logical-qubit summary with minimum weight rather than relying on dictionary order
- add regression coverage for minXZ summary selection

## Testing

Not run locally; relying entirely on repository CI as requested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## unreleased
 
+- **(fix)** Select minimum-distance MIP summaries by their distance value.
 - **(fix)** Correct `LiftedCode` construction and dimensions for concrete GF(2) group-algebra matrices.
 - Avoid runtime dispatch when resetting qubits in `MixedDestabilizer` states.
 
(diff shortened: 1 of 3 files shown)


Krastanov-agent:
### Why this fix is necessary

`opt_summary=true` changes each MIP result from an integer distance to a pair:

```julia
(distance, solution_summary)
```

The old `:minXZ` path still used `min(dX, dZ)`. That asks Julia to order the complete tuples. Tuple ordering is lexicographic, so when the X and Z distances are equal it proceeds to the second elements and attempts to order the solver summary objects. A solver summary is metadata, not part of the distance ordering and not a valid semantic tie-breaker. Equal X and Z distances are common for CSS codes, so this is a normal documented configuration, not a contrived input.

The corrected code compares only `first(dX)` and `first(dZ)`, then returns the complete result associated with the smaller distance. It deliberately selects the X result on a tie.

There is a second determinism issue when a code has more than one logical qubit. The per-logical-qubit results are stored in a `Dict`. The previous `argmin` selected the first equal-weight entry encountered through the dictionary's value iterator. Multiple logical operators can have the same minimum weight, and dictionary iteration is not the API contract that should determine which solver summary is returned. This PR uses the logical-qubit index as an explicit tie-breaker, so the requested summary is stable while the numerical minimum remains unchanged.

Before this PR, every test invocation of `DistanceMIPAlgorithm` used the default `opt_summary=false`. The existing `:minXZ` tests therefore reduced two integers with `min(::Int, ::Int)`, which works and never enters either summary-selection branch. Other distance tests also asserted only the integer distance; none requested or inspected a solver summary.

Thus the suite had extensive coverage of the MIP formulation but no coverage of the return-shape change caused by the documented `opt_summary` option. The new regression combines `:minXZ` with `opt_summary=true` on the symmetric [[48, 6, 8]] code, where the tied X/Z case exercises the exact boundary that the old tuple comparison handled incorrectly.

